PR #74, opened 2014-02-11
Link collection filtering docs to URL formatting guide

kadamwhite:
As noted in the compatibility document, URLs do not provide any standard convention for passing query parameters as an array. Linking the collection filtering guide to the article explaining the URL array syntax used in this API will reduce confusion for new users.

This tripped me up for several hours this morning; I expect other users would encounter the same difficulties.


Twintails:
This link is returning a 404

PR #256, opened 2014-05-31
Update comparison with missing XML-RPC features

maxcutler:


danielbachhuber:
We can revisit these clarifications in the future if we need to.

PR #310, opened 2014-06-11
Composer

technosailor:
Adds a composer.json file for composer-based WordPress installs.


composer.json
```diff
@@ -0,0 +1,16 @@
...
+  "authors":[
+	{
+	  "name":"Ryan McCue",
+	  "homepage":"http://ryanmccue.info/"
+	}
+  ],
+  "type": "wordpress-plugin",
+  "require": {
+    "composer/installers": "v1.0.6",
+    "php": ">=5.3.2"
```
QWp6t: The version of PHP used for composer can differ from the version used for WP-API.

For example a user might have 5.2.4 installed on her server, but access a newer version of PHP from the command line for composer, e.g., `php54 composer.phar install`

I think the PHP requirement in composer.json should match the version of PHP that's actually required by WP-API.

rmccue: > I think the PHP requirement in composer.json should match the version of PHP that's actually required by WP-API.

I agree; however, where is the requirement actually checked? If it's checked at install-time, then it doesn't matter that it's >5.2.4, and IIRC it's not checked at run-time.

(Would still merge a PR to change it anyway. :) )

chrisvanpatten:
I'd be happy to submit a new PR for this but I don't want to step on any toes. @technosailor - would you mind if I worked up a new version that accounts for the suggestions/issues raised here?

retlehs:
:eyes:

danielbachhuber:
Closing in favor of #1157

PR #327, opened 2014-06-23
Add single revision endpoint

### LINKED ISSUE #314: No endpoint for retrieving single revision
Although not incredibly useful, the situation will arise where someone wants to get a single revision without iterating through an entire collection. This is also valuable in terms of endpoint consistency.

tlovett1:
Fixes #314


danielbachhuber:
We aren't taking enhancements against `master` at this point.

PR #328, opened 2014-06-23
Added 202 Response code

nickkos101:
Returned the 202 response code.


danielbachhuber:
We're not taking any enhancements against `master` at this point.

PR #351, opened 2014-07-14
Adds ability to set custom taxonomy terms on create and edit post

pdewouters:
I couldn't get the `tax_input` paramenter of `wp_insert_post` to work reliably for some reason. As we already have the post ID at this point, `wp_set_object_terms` works fine.

`wp_set_object_terms` can also use an array of IDs or slugs, or just an ID or a slug, which should probably be handled also.

I went with the assumption that consumer will always pass an array of term names or Ids. Even for a single term.
- [x] Detect if tern names or IDs were passed, for sanitizing

Consumer could pass

```
$post['tax_input'] => array(
  'my_custom_tax' => array(
    'term_names' => array( 'My Term', 'My other Term' )
  )
)
```

or 

```
$post['tax_input'] => array(
  'my_custom_tax' => array(
    'term_ids' => array( 23,44)
  )
)
```

### DIFF
--- a/tests/test-json-posts.php
+++ b/tests/test-json-posts.php
@@ -385,7 +385,6 @@ class WP_Test_JSON_Posts extends WP_Test_JSON_TestCase {
 
 		// Custom tax
 		$custom_tax_term = get_term_by( 'name', 'My WP API Term 1', 'my-wp-api-tax' );
-		$this->assertNotFalse( $custom_tax_term );
 		$this->assertTrue( has_term( $custom_tax_term, 'my-wp-api-tax', $new_post ) );
 	}
 


samurai926:
Daniel,
"We'll be going a different route on this"

Could you please elaborate about what the different route is?  I can't find any documentation in v2 regarding how to put a post into a category.

Thanks,
Jeffrey

danielbachhuber:
> Could you please elaborate about what the different route is?

We haven't decided yet #561

> I can't find any documentation in v2 regarding how to put a post into a category.

You'll want to make a separate request to set terms on a post. I don't believe the documentation is up yet.

PR #354, opened 2014-07-15
Reverse the endpoint registration examples

BronsonQuick:
While I was following through the docs from the top down the simplest example of registering your own endpoints was below the first example.

To me it would make more sense to reverse the order and have the simplest example first then follow it with the more robust example that uses custom routes.


danielbachhuber:
We're actively maintaining the v2 docs repo at this time https://github.com/WP-API/docs-v2

Issue #224, opened 2014-05-27
/media doesn't return the alternative text


BronsonQuick:
The media endpoint should return the alternative text for attachments

danielbachhuber:
This was addressed in #904

PR #395, opened 2014-08-01
#224: Added alternative text to media API.

mauteri:
This is applied to all media, though WP only supports images at this point I believe.


danielbachhuber:
@mauteri A delayed thanks! Handling this in 38d76d2

mauteri:
You bet, thanks! :-D

PR #399, opened 2014-08-01
Add a json_prepare_type-{post_type} filter on the /posts endpoint

kadamwhite:
This is intended mainly as the stimulus for a discussion. We currently have the following filters on posts:

- `json_prepare_post`: Runs on anything requested via /posts
- `json_prepare_{post_type}`: Runs on anything requested via and endpoint extending `WP_JSON_CustomPostType`

Right now, if you don't make a custom endpoint and request a CPT, you have to filter on `json_prepare_post` and conditionally check for the target post type.

Having the filters exist per-endpoint is fine (although I feel like I've seen a situation where another endpoint hit json_prepare_post, via a call to `WP_JSON_Posts`'s `prepare_post`, I cannot find it atm). However, I wanted to raise the question of whether we should be able to filter against CPT objects retrieved via /posts?type= separately from `json_prepare_posts`, and, if so, whether that filter should also be invoked/invokable from a custom endpoint.

This implementation is set up to add a new filter method:

- `json_prepare_type-{post_type}`

This method processes any request passed through the `/posts` endpoint.

My question:

Should this be applied to objects retrieved through custom endpoints? *i.e.*, should `json_prepare_type-{post_type}` be called from within `WP_JSON_CustomPostType`'s `prepare_post`? If so, this would mean that

- `json_prepare_post` would be called for anything on `/posts`
- `json_prepare_{post_type}` would be called for anything on a custom endpoint
- `json_prepare_type-{post_type}` would be called for any custom type returned from the system, regardless of through which endpoint it was requested.

I personally like this approach, because it lets you pick your level of specificity with a filter: whether it should handle all resources of a type, or whether it should handle all resources from a custom endpoint (or the default one). It makes things more consistent with the filter `json_prepare_attachment`, as well, since that's then effectively a `json_prepare_{post_type}` call, just as is `json_prepare_post` is for that endpoint. It de-couples filters for the type of object from filters for how it was retrieved. (Names may need to change; json_prepare_posts and json_prepare_media, *etc*, though that may be irrelevant if we go with an exclusively root-discoverable API).

Thoughts?

### DIFF
--- a/lib/class-wp-json-customposttype.php
+++ b/lib/class-wp-json-customposttype.php
@@ -210,7 +210,11 @@ abstract class WP_JSON_CustomPostType extends WP_JSON_Posts {
 			$_post['meta']['links']['up'] = json_url( $this->base . '/' . $post['ID'] );
 		}
 
-		return apply_filters( "json_prepare_{$this->type}", $_post, $post, $context );
+		// json_prepare_{post_type} filters all objects requested via this endpoint
+		$_post = apply_filters( "json_prepare_{$this->type}", $_post, $post, $context );
+
+		// Run post type-specific filters
+		return apply_filters( "json_prepare_type-{$this->type}", $_post, $post, $context );
 	}
 
 	/**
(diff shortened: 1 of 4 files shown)


danielbachhuber:
We're revisiting where and when we add filters in #875

danielbachhuber:
We're not taking any enhancements against `master` at this point. We have an equivalent filter on `develop`: https://github.com/WP-API/WP-API/blob/develop/lib/endpoints/class-wp-rest-posts-controller.php#L1052

Issue #346, opened 2014-07-08
Fetch WordPress settings through wp-api


Buooy:
Is there any api reference as to how we can pull setting details from the WordPress Instance.

I would like to grab the UTC and date format that users enter in the settings.

danielbachhuber:
We'll handle in #816

PR #432, opened 2014-08-07
Options Endpoint

EdHurtig:
Creates an endpoint for options.  All requests must be authenticated with a `manage_options` user.  

` GET /options => All Options `

` GET /options/<name> => Get Option `

` POST /options + data => Add Option `

` PUT /options/<name> + data => Update Option `

` DELETE /options/<name> => Delete Option `

Still needs some work, but I thought I should PR and let people weigh in.  I put a wish list on #346 for things that should be be added. Transient routes are commented, but I would like to add that functionality before merging 

PR Includes tests and docs. 

See #346


EdHurtig:
@danielbachhuber sorry, I neglected this for the longest time.  Is there an options endpoint planned, built, or otherwise in thought?  I know it is pretty late in the dev cycle for development on a whole new plugin.  But If you guys want an options endpoint and think it would be able to make it in time, I will gladly dive back into this.

danielbachhuber:
No worries.

We're creating a more abstracted endpoint for options. You can follow along in the conversation on #816

PR #434, opened 2014-08-08
Added 'offset' to $json_valid array

nrutkowski:
Added 'offset' to $json_valid array to define the custom offset filter for get_posts function. 
This fixed the issue with offset not working in the GET posts request.


danielbachhuber:
We're no longer taking enhancements against `master`. `develop` query params will be visited in #924

PR #436, opened 2014-08-09
Hide terms that belong to a non public taxonomy.

markoheijnen:
Currently when making request to /posts or /posts/<id> it also shows terms of hidden non-public taxonomies. I guess we should hide those since we do hide the taxonomies from /taxonomies.


danielbachhuber:
This is addressed in `develop`: https://github.com/WP-API/WP-API/blob/develop/lib/endpoints/class-wp-rest-terms-controller.php#L252

PR #474, opened 2014-09-08
Fix BC Break with Post Meta Handling in Post Routes

rachelbaker:
During the split of the meta handling into it's own classes (#358) we broke documented functionality, including:

1. Allowing post_meta creation when creating a new post.
2. Allowing post_meta actions (add/edit/delete) when editing an existing post.
3. Returning post_meta values within the response for a post.

Additional unit tests needed here to make sure all edge cases are working as expected.

Originally identified by @tlovett1.


rachelbaker:
merged #474

PR #480, opened 2014-09-19
Pagination headers missing on /users or terms collections

rachelbaker:
The pagination headers are only getting set for `WP_JSON_Response` objects within the `get_posts` method. This means that there is no way to request a subset of users, or of terms. Users is the critical one, for me at least: the default wp-json/users endpoint returns a default collection of 10 posts (filterable via `number`), and the `page` query param works, but without the headers it is impossible to easily traverse the collection.


danielbachhuber:
This will be addressed by #537

Issue #370, opened 2014-07-27
Add ability to get status as different header (support HTTP envelope)


rmccue:
Browsers always follow 3xx redirects, and it's not possible to intercept requests between sending the first and any redirects. This causes an issue with in-browser OAuth, as requests must be signed and the signature includes the path.

We should add the ability to get the header back as e.g. `X-WP-Status` instead, and always return 200.

PR #493, opened 2014-09-23
WP-API/WP-API#370 - Remove 302 http redirection ...

elebescond:
... for avoid problems with CORS restriction

"The request was redirected to 'http://xxxxx.com/wp-json/users/1', which is disallowed for cross-origin requests that require preflight."

### DIFF
--- a/README.md
+++ b/README.md
@@ -48,10 +48,10 @@ Want to test out WP-API and work on it? Here's how you can set up your own
 testing environment in a few easy steps:
 
 1. Install [Vagrant](http://vagrantup.com/) and [VirtualBox](https://www.virtualbox.org/).
-2. Clone [Chassis](https://github.com/sennza/Chassis):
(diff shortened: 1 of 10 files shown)


hyyan:
@elebescond , @kosso ,@rmccue , Hey guys , this PR causing me an redirect error , I mean it is causing the problem it must solve , I don't know why ? 

I keep getting the following error : 
```
XMLHttpRequest cannot load http://testo.localhost.com/wp-json/users/me. The request was redirected to 'http://testo.localhost.com/wp-json/users/1', which is disallowed for cross-origin requests that require preflight.

```

Removing the following lines : 
```
	$data = $response->get_data();
-
-		$response->header( 'Location', $data['meta']['links']['self'] );
-		$response->set_status( 302 );

```

Solve the problem , any idea why ? thanks in advance

PR #513, opened 2014-10-05
Allow setting featured image when inserting post. 

### LINKED ISSUE #322: Featured image with new post.
Currently, a featured image cannot be created with a new post.  In order for a featured image to be set, the media has to be created.  Then the post has to be created.  Then the post has to be updated 

`public function attach_thumbnail( $post, $data, $update ) {if ( ! $update ) {return;} ... `

This, in my opinion, is inefficient and requires a remote device to make three requests to the server instead of just on or two.  The - is $update - conditional should be removed?

On this same note, in order for a remote device to update meta data, the device has to call the post/post-type meta get request, then parse the available meta to obtain meta ID's and then send that data back in the form of an automatically indexed array.  

One, the meta ID is not readily available in wordpress.  So, a devloper cannot readily provide the device developer with appropriate meta key values to use for creation/updating.  I understand that accessing the meta via get request allows for all use cases, but using the id as the determining get/set value is obscure as far as typical wordpress goes.  

Two, because the array is expected to be automatically indexed, each index consisting of an associative array with keys ID, Key, and Value, if a developer needs to hook into the create/edit post to deal with meta in a specific way (such as arrays, objects or serialized data), they then have to loop through all meta and "look" for a key (or an ID, but no one will be looking for meta id, they will be using keys) manually.  This is time consuming.

The array, in my opinion, should look more like
 `$data['post_meta'] = array("key1"=>"val1", "key2" => "val2", ...) `
instead of 
`$data['post_meta'] = array(
     array("ID"=>"id1","Key"=>"key1", "Value"=>"val1), 
     array("ID"=>"id2","Key"=>"key2", "Value"=>"val2), 
);`

It makes it easier to deal with just the value you want to deal with, instead of having to loop through the entire stack (or even up to found point with break).  At the very least structuring it like this would be preferable:
`$data['post_meta'] = array(
     "ID1"=array("Key"=>"key1", "Value"=>"val1), 
    "ID2"=array("Key"=>"key2", "Value"=>"val2), 
);`

Shelob9:
Allows post_thumbnail to be set (as an attachment ID)
Fixes #322

### DIFF
--- a/lib/class-wp-json-posts.php
+++ b/lib/class-wp-json-posts.php
@@ -1094,6 +1094,11 @@ class WP_JSON_Posts {
 			}
 		}
 
+		//post thumbnail
+		if ( ! empty( $data['post_thumbnail'] ) && intval( $data['post_thumbnail'] ) > 0 && post_type_supports( $post['post_type'], 'thumbnail' ) ) {
+			set_post_thumbnail( $post_ID, $data['post_thumbnail'] );
+		}
+
 		do_action( 'json_insert_post', $post, $data, $update );
 
 		return $post_ID;
--- a/plugin.php
+++ b/plugin.php
@@ -759,3 +759,66 @@ function json_last_error_msg() {
 	}
 }
 endif;
+
+if ( ! function_exists( 'json_insert_post_from_json' ) ):
+/**
+ * Insert a post form a remote site on current site
+ *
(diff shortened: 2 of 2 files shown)


danielbachhuber:
Assigning `featured_image` will be addressed in #832.

Creating an attachment and assigning it as the featured image will be sorted out with #561

Issue #492, opened 2014-09-23
API incompatible with setup_postdata


sebastiaandegeus:
Hello,

Using this API and setting up the results with `setup_postdata()` again in another WordPress site creates incompatibility issues on the following properties:

`post_author` versus `author`
`post_date` versus `date`
`post_content` versus `content`
`post_title` versus `title`

This means that `the_title()`, `the_content()`, etc will not work anymore.

I can understand the desire to change these ancient property names to something that makes sense but it breaks compatibility with very standard WordPress functions which most people expect to work. 

We fixed our problem by filtering `json_prepare_post` and assigning the properties.

Related closed issue. 
https://github.com/WP-API/WP-API/issues/41

danielbachhuber:
The object schema for the API will be different than the internal WordPress database model. The API project gives us the opportunity to revisit 12 years of decisions, and make changes we think will be better for the next 12 years. Some object attributes will be similar, others will be different.

#605 is a good reference point for this on-going conversation.

PR #514, opened 2014-10-05
#492

Shelob9:
This PR adds a new function that takes a response from a GET request for an individual post, prepares it for `wp_insert_post()` and then creates the post using `wp_insert_post()`.

The one thing it doesn't do is handle taxonomies. Instead it places that part of the original response on a hook and unsets that part of the response. I didn't think it was safe to assume the term IDs will match on the current site to what is on the site that the incoming JSON data is from.

Possible solutions to this that I see:
1) Make term insert optional. That way if the IDs are known to match, than the end user can choose to allow use them.
2) Provide a second function for setting the terms that would allow for ID switching.

### DIFF
--- a/plugin.php
+++ b/plugin.php
@@ -816,7 +816,7 @@ if ( ! function_exists( 'json_insert_post_from_json' ) ):
 		 * @param int $id The ID of the post that was created.
 		 * @param int $original_id ID of the post being copied from
 		 */
-		add_action( 'json_insert_post_from_json_terms', $terms, $id, $post[ 'import_id' ]  );
+		do_action( 'json_insert_post_from_json_terms', $terms, $id, $post[ 'import_id' ]  );
 
 		return $id;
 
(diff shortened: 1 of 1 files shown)


danielbachhuber:
We're not accepting any enhancements against `master` at this point.

Shelob9:
@danielbachhuber I totally forgot about this PR. I think this is an important feature. Is it too late to get it into 2.0 if I make a new PR against develop?

danielbachhuber:
> Is it too late to get it into 2.0 if I make a new PR against develop?

I'm not sure it's suitable for the plugin.

rmccue:
> I'm not sure it's suitable for the plugin.

Agreed. If you want to do this in any case, you should be using the API endpoints (`WP_REST_Posts_Controller::create_item`) or lower-level handlers to get the correct error handling. This is also important for sanitisation and validation, as well as permission checking.

PR #448, opened 2014-10-06
Possible BC break with timezone handling

rachelbaker:
Looks like with 6687602b73064d2e8ce6708b7018b2dd02bf2beb / 6d0ad766ca525c7a3aee921b66554ddf21a5c023 / 46498224624f419d915ce1a14afb399dc4d95850, we did two things which break BC: removed `date_tz` and `modified_tz`, and began stripping timezones from date strings.

While I agree these should be changed, they're a BC break in 1.x, so we need to revert at least the `*_tz` part.

I *think* we can get away with stripping timezone information, but we should check that.

### DIFF
--- a/lib/class-wp-json-comments.php
+++ b/lib/class-wp-json-comments.php
@@ -187,7 +187,11 @@ class WP_JSON_Comments {
 		}
 
 		// Date
+		$timezone     = json_get_timezone();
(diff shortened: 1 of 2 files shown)


rachelbaker:
@rmccue Since this is a regression in master, should we go ahead and fix this?

rachelbaker:
Merged #448

PR #524, opened 2014-10-09
WP_JSON_Posts::get_post_types() should return array, not hash/object

wkw:
`WP_JSON_Posts::get_post_types()` may have had some old code around. it was returning an object whose keys were the Posttype names, rather than an array of objects. Discovered this while exploring the client-js repo.

### DIFF
--- a/lib/class-wp-json-posts.php
+++ b/lib/class-wp-json-posts.php
@@ -546,7 +546,7 @@ class WP_JSON_Posts {
 				continue;
 			}
 
-			$types[ $name ] = $type;
+			$types[] = $type;
 		}
 
 		return $types;


danielbachhuber:
This was fixed for `develop` in #914

rmccue:
Two things here:
1) This was correct behaviour; it's intended to be a hash, as the post type names are unique IDs. It makes it much easier to use by ensuring you can key by name where possible (due to technical limitations, we cannot do the same for anything with numeric IDs).
2) This is a backwards compatibility break, so cannot go into 1.2. The one in #914 for v2 is a separate issue we need to tackle.

Reverting in #933.

PR #531, opened 2014-10-21
add handler for terms and term_names in insert_post method like it was in xml-rpc

aslikeyou:
Here I tried to make the same logic as it was in wordpress xml-rpc plugin. There you can sent only term_names like { category: ['cat1', 'cat2', 'cat3'], post_tag: ['tag1', 'tag2', 'tag3'] } without any ids and api can add unexist terms automatically.@see http://codex.wordpress.org/XML-RPC_WordPress_API/Posts#wp.newPost


rachelbaker:
@aslikeyou,

How we are going to approach requests like this is still being discussed in #561.

PR #602, opened 2014-11-12
Specify Capability in Route

joehoyle:
Support for specifying capabilities in the endpoints. Also we send the "Allow" header taking into account the capabilities and the current user.

Registering a route one can now do:

```PHP
        register_json_route( 'wp', '/taxonomies/(?P<taxonomy>[\w-]+)', array(
                'methods'         => 'GET',
                'callback'        => array( $controller, 'get_item' ),
                'capability'      => 'manage_terms',
        ) );
```
@rmccue I added `send_header` in the `dispatch` methods, as this is really the only place of having context of the matched route, and also the rest of the other methods for the same route. Is this breaking a pattern by having `dispatch` output headers?


joehoyle:
@WP-API/amigos #reviewmerge

danielbachhuber:
Looks fine to me. Defer to @rmccue for the final merge

joehoyle:
Ping @rmccue

joehoyle:
Ping @rmccue ... again

danielbachhuber:
Yolo. We can iterate from here as needed.

PR #622, opened 2014-11-19
Change param name for json_get_avatar_url() to $id_or_email 

### LINKED ISSUE #585: Use ID instead of email address to get user's avatar
Can we change the `json_get_avatar_url()` function to accept `$email_or_id` the same way `get_avatar()` works and then use the user's ID to get the avatar in `prepare_user()`?

I have a use case (I'm assuming others would as well) where some users don't have an email address because they were created from a twitter account. Because we're passing the email address instead of the ID, I have no way to detect what user is requesting an avatar in the `get_avatar` filter.

https://github.com/WP-API/WP-API/blob/develop/lib/class-wp-json-users.php#L282

Shelob9:
Also use $user ID in new user controller when calling this function. Resolves #585 

Replaces PR #599

### DIFF
--- a/lib/class-wp-json-users-controller.php
+++ b/lib/class-wp-json-users-controller.php
@@ -214,7 +214,7 @@ class WP_JSON_Users_Controller extends WP_JSON_Controller {
 			'nickname'    => $user->nickname,
 			'slug'        => $user->user_nicename,
 			'url'         => $user->user_url,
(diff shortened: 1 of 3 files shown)


rachelbaker:
We are going to use the new core function `get_avatar_url()` that does allow the passing of id or email.  https://core.trac.wordpress.org/changeset/31107

PR #625, opened 2014-11-21
Bugfix: getting error when updating a post with featured_image

kennyki:
PHP fatal error:  Cannot use object of type WP_JSON_Response as array in /var/www/wp-content/plugins/json-rest-api/lib/class-wp-json-media.php


danielbachhuber:
This will be fixed with #832

PR #627, opened 2014-11-21
Return post data from post delete endpoint

### LINKED ISSUE #618: Delete requests shouldn't return response with message body
In the user's controller, we return this response when the delete request is successful:

```
return array( 'message' => __( 'Deleted user' ) );
```

As a matter of practice, I don't think we should *ever* return a "message" in a 2xx response. "Success" is implied by the response code, and it's inconsistent generally.

@WP-API/amigos opinions?

rmccue:
Fixes #618.

### DIFF
--- a/lib/class-wp-json-posts.php
+++ b/lib/class-wp-json-posts.php
@@ -369,18 +369,33 @@ class WP_JSON_Posts {
 			return new WP_Error( 'json_user_cannot_delete_post', __( 'Sorry, you are not allowed to delete this post.' ), array( 'status' => 401 ) );
 		}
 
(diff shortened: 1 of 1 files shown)


rmccue:
This one's on me, sorry. Will fix up ASAP.

rmccue:
(Also needs unification with #789)

nacin:
What needs discussion here?

danielbachhuber:
> What needs discussion here?

#789

rachelbaker:
Punting to Beta 2.

danielbachhuber:
Handling in #1224

PR #653, opened 2014-11-27
Allow passing just a boolean as the arg options

rmccue:
Right now, we have a huge amount of repeated code that looks like:
```php
'x' => array(
    'required' => true,
),
'y' => array(
    'required' => false,
),
```

We can remove this boilerplate while still allowing the extensibility of different options for those who want to use it. This PR changes the code to allow passing just a boolean if you're doing the simple stuff.
```php
'x' => true,
'y' => false,
```

### DIFF
--- a/lib/class-wp-json-server.php
+++ b/lib/class-wp-json-server.php
@@ -472,6 +472,13 @@ class WP_JSON_Server {
 		}
 
 		foreach ( $attributes['args'] as $key => $arg ) {
+			// Short-syntax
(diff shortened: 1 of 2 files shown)


rachelbaker:
On December 29, 2014 @rmccue expressed interest in keeping this PR open.  

Link to discussion on Slack #core-rest-api channel: https://wordpress.slack.com/archives/core-restapi/p1419895946001647

PR #668, opened 2014-12-02
Fixes a fatal error when calling parent::get_post() in WP_JSON_Media::get_post()

goblindegook:
Fixes a fatal error encountered when calling the (non-existent) `get_post()` in `WP_JSON_Media::get_post()` when fetching a post with attachments.


danielbachhuber:
@rmccue @rachelbaker Any reason we haven't merged this yet?

rmccue:
Looks like this isn't needed any more with the new attachments controller.

Thanks @goblindegook for the patch anyway, sorry for not merging sooner!

Issue #605, opened 2014-11-12
Verify all model fields exposed in various contexts


danielbachhuber:
We should produce a spreadsheet that identifies the fields associated with each model, and the view context required to expose them.

For instance, in #603, we're still not quite sure which user attributes should be exposed in which context. It would be best to have a master list to work from (with agreed-upon model attribute names).

Also, as a part of this issue, we should audit the query arguments for collections. As much as possible, they should be consistent.

danielbachhuber:
@rachelbaker is documenting all of our responses here https://hackpad.com/ep/group/ngDYn5lciXQ

danielbachhuber:
@rachelbaker now that schemas have dropped and we have clear articulation of (most) of the existing resource attributes, how would you like to proceed?

danielbachhuber:
We could write the schema first for attachments, and build from there. Attachments is a tricky wicket because audio _should_ have different attributes than video and file, etc.

danielbachhuber:
This is in pretty good shape now, but we should do a final pass as a part of b1

danielbachhuber:
These are pretty well documented in the schemas. We can open issues in the future as we have specific discussion points.

PR #693, opened 2014-12-08
Comments Endpoints

joehoyle:
Still a Work In Progress, but opening to show progress and encourage any feedback as this progresses.

I took the (controversial) decision to use /comments as the canonical place for comments. I was then thinking `/posts/1/comments` would 301 to `/comments?post_id=1` or somehting along those lines. I think it's nicer to keep the comments at their own top level so they are independant of posts apart from the embedding we are doing elsewhere already.


joehoyle:
@rachelbaker apologies, haven't had any time to spend on this over the holidays, you are more than welcome to carry on where I left off. If you do, just assign this PR to you so I don't continue!

rachelbaker:
@joehoyle Thank you.  I will pick this up.

rachelbaker:
This is still missing the update_comment endpoint, I will wrap that up this week.

rachelbaker:
In the future, I would like to use `wp_new_comment()` instead of `wp_insert_comment()`.  Related core ticket: https://core.trac.wordpress.org/ticket/14601 

Also `wp_update_comment()` doesn't update the `comment_post_ID`.  Related core ticket: https://core.trac.wordpress.org/ticket/30946

rachelbaker:
@WP-API/amigos #reviewmerge  

Test coverage (per the PHPUnit Coverage report) is currently at the following:
get_items: 94.12%
get_item:  92.86%
create_item: 80.95%
update_item: 85.19%
delete_item: 69.23%

rachelbaker:
- Added support for more GET /comments endpoint query parameters
- Filtered the GET /comments endpoint query parameters based on permissions.
- Added additional test coverage for the delete_item method, so that is now: 84.62%
- Audited DocBlocks.

I am happy with where this is at.

lib/class-wp-json-comments-controller.php
```diff
@@ -0,0 +1,470 @@
...
+	 *
+	 * Correctly handles posts with the inherit status.
+	 *
+	 * @param  WP_Post $post Post Object.
+	 * @return boolean Can we read it?
+	 */
+	protected function check_read_post_permission( $post ) {
+		$posts_controller = new WP_JSON_Posts_Controller;
+
+		return $posts_controller->check_read_permission( $post );
```
danielbachhuber: Seems too tightly-coupled. Can we do it in a more de-coupled manner?

Do comments always need to have a post? If so, why are we exposing them at `/wp/comments` ?

lib/class-wp-json-comments-controller.php
```diff
@@ -0,0 +1,470 @@
...
+	 * @param  WP_JSON_Request $request Full details about the request.
+	 * @return WP_Error|WP_JSON_Response|mixed
+	 */
+	public function get_items( $request ) {
+		$prepared_args = $this->prepare_items_query( $request );
+
+		$query = new WP_Comment_Query;
+		$comments = $query->query( $prepared_args );
+
+		foreach ( $comments as &$comment ) {
```
danielbachhuber: Looping by reference is a problem when your loop checks read permission, but doesn't drop the variable on failure :cry:

Now that I think of it, we have the same problem in the Posts controller :cry: :cry: #788

lib/class-wp-json-comments-controller.php
```diff
@@ -0,0 +1,470 @@
...
+			$post = get_post( $comment->comment_post_ID );
+			if ( ! $this->check_read_post_permission( $post ) || ! $this->check_read_permission( $comment ) ) {
```
danielbachhuber: Rather than scattering a bunch of duplicate checks in the controller, can we have `check_read_permission()` also check `check_read_post_permission()` if the comment is assigned to a post?

lib/class-wp-json-comments-controller.php
```diff
@@ -0,0 +1,470 @@
...
+			return new WP_Error( 'json_comment_exists', __( 'Cannot create existing comment.' ), array( 'status' => 400 ) );
+		}
+
+		$post = get_post( (int) $request['post_id'] );
+		if ( empty( $post ) ) {
+			return new WP_Error( 'json_post_invalid_id', __( 'Invalid post ID.' ), array( 'status' => 404 ) );
+		}
+
+		if ( ! $this->check_read_post_permission( $post ) ) {
+			return new WP_Error( 'json_user_cannot_read', __( 'Sorry, you cannot read this post.' ), array( 'status' => 401 ) );
```
danielbachhuber: Seems like `check_read_post_permission()` should return this `WP_Error`, so we don't have it duplicated everwhere.

lib/class-wp-json-comments-controller.php
```diff
@@ -0,0 +1,470 @@
...
+	/**
+	 * Edit a comment
+	 *
+	 * @param  WP_JSON_Request $request Full details about the request.
+	 * @return WP_Error|WP_JSON_Response|mixed
+	 */
+	public function update_item( $request ) {
+		$id = (int) $request['id'];
+
+		$comment = get_comment( $id );
```
danielbachhuber: Let's reuse `self::get_item()` here.

lib/class-wp-json-comments-controller.php
```diff
@@ -0,0 +1,470 @@
...
+
+	/**
+	 * Delete a comment.
+	 *
+	 * @param  WP_JSON_Request $request Full details about the request.
+	 * @return WP_Error|array|mixed
+	 */
+	public function delete_item( $request ) {
+		$id = (int) $request['id'];
+		$force = isset( $request['force'] ) ? (bool) $request['force'] : false;
```
danielbachhuber: I still think deleting an item should always delete an item. If you want to trash the item, it should be `PUT status=trash`

danielbachhuber: Filed an issue for it #789

lib/class-wp-json-comments-controller.php
```diff
@@ -0,0 +1,470 @@
...
+	 * @param  WP_JSON_Request $request Request object.
+	 * @return array $fields
+	 */
+	public function prepare_item_for_response( $comment, $request ) {
+		$fields = array(
+			'id'           => (int) $comment->comment_ID,
+			'post_id'      => (int) $comment->comment_post_ID,
+			'parent_id'    => (int) $comment->comment_parent,
+			'user_id'      => (int) $comment->user_id,
+			'author'       => $comment->comment_author,
```
danielbachhuber: Should this be `author_name`, to make it less ambiguous?

lib/class-wp-json-comments-controller.php
```diff
@@ -0,0 +1,470 @@
...
+			'author_email' => $comment->comment_author_email,
+			'author_url'   => $comment->comment_author_url,
+			'date'         => json_mysql_to_rfc3339( $comment->comment_date ),
```
danielbachhuber: Should this be `publish_date` for consistency?

lib/class-wp-json-comments-controller.php
```diff
@@ -0,0 +1,470 @@
...
+		}
+
+		$links = array();
+
+		// Author
+		if ( 0 !== (int) $comment->user_id ) {
+			$links['author'] = array(
+				'href' => json_url( '/wp/users/' . $comment->user_id )
+			);
+		}
```
danielbachhuber: Should we include a link to the post too?

plugin.php
```diff
@@ -320,6 +321,109 @@ function create_initial_json_routes() {
...
+				'type'         => array(
+					'default'      => 'comment',
+				),
+				'parent_id'    => array(),
+				'search'       => array(),
+				'order'        => array(
+					'default'      => 'DESC',
+				),
+				'orderby'      => array(
+					'default'      => 'comment_date_gmt',
```
danielbachhuber: `comment_date_gmt` should instead be whatever name we're actually exposing it by.

plugin.php
```diff
@@ -320,6 +321,109 @@ function create_initial_json_routes() {
...
+		array(
+			'methods'  => WP_JSON_Server::CREATABLE,
+			'callback' => array( $controller, 'create_item' ),
+			'args'     => array(
+				'post_id'      => array(
+					'required'     => true,
+				),
+				'type'         => array(
+					'required'     => true,
+					'default'      => 'comment',
```
danielbachhuber: If there's a default value, does it need to be required?

danielbachhuber:
Thanks @rachelbaker , looking pretty good. Just needs a little bit of cleanup.

lib/class-wp-json-comments-controller.php
rachelbaker: Related Trac Ticket: https://core.trac.wordpress.org/ticket/23931  
When checking capabilities some checks return wp_die or fail if there isn't a post_id associated with a post.

lib/class-wp-json-comments-controller.php
rachelbaker: @danielbachhuber This approach was the result of a discussion I had in our Slack channel with @joehoyle yesterday.  We cannot reuse the get_item method, if we want to allow unauthenticated users (user_id=0) the ability to create a comment.  
If an unauthenticated user creates a comment, the comment (depending on the comment moderation settings) will have a status of 'hold'.  The authenticated user cannot view the comment via the GET endpoint unless they have the `moderate_comments` permission.

lib/class-wp-json-comments-controller.php
rachelbaker: @danielbachhuber comments don't have modified_date like posts.  We could make it `comment_date`?  `publish_date` seems confusing since you can modify the date.

plugin.php
rachelbaker: This is strictly a `WP_Comment_Query` parameter.  If we convert parameter names here, we would have to do that for all of the endpoints.  I would rather remove it.

lib/class-wp-json-comments-controller.php
danielbachhuber: > publish_date seems confusing since you can modify the date.

Can't you modify the date of posts?

plugin.php
danielbachhuber: > If we convert parameter names here, we would have to do that for all of the endpoints.

Yep, we should. A developer building off the API shouldn't be expected to know WordPress internal params / database column names.

> I would rather remove it.

We shouldn't cut corners on this project :smile:

lib/class-wp-json-comments-controller.php
```diff
@@ -0,0 +1,476 @@
...
+			'search'  => $request['search'] ? sanitize_text_field( $request['search'] ) : '',
+			'orderby' => sanitize_key( $request['orderby'] ),
+			'order'   => sanitize_key( $request['order'] ),
+			'status'  => 'approve',
+			'type'    => 'comment',
+		);
+
+		$prepared_args['offset'] = $prepared_args['number'] * ( absint( $request['page'] ) - 1 );
+
+		if ( current_user_can( 'edit_posts' ) ) {
```
rachelbaker: @rmccue @justinshreve @joehoyle  I expanded the available /comments query parameters BUT I limited the parameters available to unauthenticated users.  I tried to separate what is "public" vs "private" using some logic.  Any concerns here?

justinshreve: The private vs public separation here makes sense to me.

plugin.php
rachelbaker: Added issue #791 since we have to do this for all the endpoints.

rachelbaker: Added for comments in  dc59663 & 7c47f1a

lib/class-wp-json-comments-controller.php
joehoyle: Hmm in the context of `check_read_post_permission` returning a `WP_Error` seems to indicate there was an issue checking the permissions, not that the permission check is 'not allowed'. Pretty easy either way, convention seems to be that it would return `bool`

rachelbaker: Convention is to return a `boolean` value in our permission checks, and then return the appropriate error message.  This seems like bikeshedding to me, but feel free to open a PR post merge.

lib/class-wp-json-comments-controller.php
rachelbaker: Related #605

lib/class-wp-json-comments-controller.php
rachelbaker: Related #605

fahmi182:
and how to enable this?

like this?
`add_filter('rest_allow_anonymous_comments',false);`

kingstakh:
+1 how enable guest comments now?

fahmi182:
@kingstakh

`add_filter( 'rest_allow_anonymous_comments', '__return_true' );`

Issue #708, opened 2014-12-10
json_ensure_response() should also handle WP_Error


danielbachhuber:
It's 2014 and it's annoying to write tests when it doesn't.

Or, we can have `WP_JSON_Server::dispatch()` always return `WP_JSON_Response`.

@WP-API/amigos Thoughts? Both?

danielbachhuber:
This is in a good shape now.

PR #714, opened 2014-12-10
WP_Json_Server::dispatch() should always return a WP_JSON_Response

joehoyle:
Discussed in #708

### DIFF
--- a/lib/class-wp-json-response.php
+++ b/lib/class-wp-json-response.php
@@ -9,6 +9,20 @@ class WP_JSON_Response extends WP_HTTP_Response {
 	protected $links = array();
 
 	/**
+	 * The route that was to create the response
+	 * 
+	 * @var string
+	 */
+	protected $matched_route = '';
+
+	/**
+	 * The handler that was used to create the response
+	 * 
+	 * @var null|array
+	 */
+	protected $matched_handler = null;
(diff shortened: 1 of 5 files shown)


rmccue:
> From your example, couldn't we just patch is_wp_error() to also handle an errored Response object?

We could (and that's my preferred option), but that'd need to wait for core to change. My hope is that `->is_error()` gets us the same for now, and we can easily switch in the future.

Above thoughts were just to see if we can solve the specific problem without changing the `WP_Error` responses (i.e. minimum viable patch).

joehoyle:
@rmccue the reason for this patch is to return an object that can contain the matched route, so that route can be used upstream when generating extra headers and things like that. To do that cleanly, we need `dispatch()` to return a consistent object, in this case `WP_JSON_Response`. `dispatch()` is _always_ going to cause _some_ response, is that response is an error, well... it's still an response, and a "meta error" about the route / caller, not so much an error with "dispatching a response". 

I'm still in favor of this patch - though we have been back and forth quite a bit now around this issue, and I don't want to code up yet another approach without knowing where we are headed if not this direction. 

+1 for merging this.

joehoyle:
@WP-API/amigos reviewmerge plz!

PR #727, opened 2014-12-17
Update class-wp-json-media.php

mathieuaubert:
$thumbnail is a WP_JSON_Posts
ID is a dimension to data.

For your interest :
The post property "post_thumbnail" mentionned into the documentation (http://wp-api.org/#entities_post_post-thumbnail) is replaced by "featured_image" into the code.


danielbachhuber:
Thanks for the pull request, @mathieuaubert. We're converting `/media` to a new pattern in #904

PR #746, opened 2014-12-27
Let the title, content and excerpt be empty

### LINKED ISSUE #745: Can't update the post with empty content, title or excerpt

iseulde:
Fixes #745. Brings up another issue with the excerpt...

### DIFF
--- a/lib/class-wp-json-base-posts-controller.php
+++ b/lib/class-wp-json-base-posts-controller.php
@@ -70,6 +70,8 @@ abstract class WP_JSON_Base_Posts_Controller extends WP_JSON_Controller {
 			return new WP_Error( 'json_post_invalid_id', __( 'Invalid post ID.' ), array( 'status' => 404 ) );
 		}
 
+		setup_postdata( $post );
+
 		if ( 'edit' === $request['context'] && ! $this->check_update_permission( $post ) ) {
 			return new WP_Error( 'json_post_cannot_edit', __( 'Sorry, you are not allowed to edit this post.' ), array( 'status' => 403 ) );
 		} elseif ( ! $this->check_read_permission( $post ) ) {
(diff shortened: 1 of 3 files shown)


lib/class-wp-json-base-posts-controller.php
```diff
@@ -68,6 +70,8 @@ public function get_item( $request ) {
 			return new WP_Error( 'json_post_invalid_id', __( 'Invalid post ID.' ), array( 'status' => 404 ) );
 		}
 
+		setup_postdata( $post );
```
rmccue: If we set this up here, we need to ensure we restore it afterwards. We're already doing that inside `prepare_item_for_response`, do we need it here too?

lib/class-wp-json-base-posts-controller.php
```diff
@@ -375,31 +379,28 @@ protected function prepare_item_for_database( $request ) {
 		}
 
 		// Post title
-		if ( ! empty( $request['title'] ) ) {
+		if ( isset( $request['title'] ) ) {
 			if ( is_string( $request['title'] ) ) {
 				$prepared_post->post_title = wp_kses_post( $request['title'] );
-			}
-			elseif ( ! empty( $request['title']['raw'] ) ) {
+			} else if ( isset( $request['title']['raw'] ) ) {
```
rmccue: Should be `elseif` per coding standards.

tests/test-json-posts-controller.php
```diff
@@ -786,7 +828,11 @@ protected function check_get_posts_response( $response, $context = 'view' ) {
 
 		$all_data = $response->get_data();
 		$data = $all_data[0];
+
 		$post = get_post( $data['id'] );
+
+		setup_postdata( $post );
```
rmccue: Is there a reason to do this inside the tests?

rmccue:
Will need a merge resolution too. :)

danielbachhuber:
@iseulde I think this is addressed by #953 and #952. Can you confirm?

danielbachhuber:
This has been addressed a while back.

PR #753, opened 2014-12-30
Resolve regressions in Posts Controller

### LINKED ISSUE #738: Regressions in post controller
#684 introduced some regressions in the post controller from v1.

* `setup_postdata` is not called before preparing the post, which breaks filtering (such as `the_content`)
* Passworded posts are not handled correctly
* `post_date` is checked for `0000-00-00 00:00:00`, but `post_date_gmt` should be checked instead

rachelbaker:
Fixes #738

### DIFF
--- a/tests/test-json-posts-controller.php
+++ b/tests/test-json-posts-controller.php
@@ -310,6 +310,11 @@ class WP_Test_JSON_Posts_Controller extends WP_Test_JSON_Controller_Testcase {
 		$new_post = get_post( $data['id'] );
 		$this->assertEquals( 'draft', $data['status'] );
 		$this->assertEquals( 'draft', $new_post->post_status );
+		// Confirm dates are null
+		$this->assertNull( $data['date_gmt'] );
+		$this->assertNull( $data['modified_gmt'] );
+		$this->assertNull( $data['date'] );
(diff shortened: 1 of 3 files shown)


rachelbaker:
@rmccue #reviewmerge

danielbachhuber:
@rachelbaker  should we add tests for the other bugs?

tests/test-json-posts-controller.php
```diff
@@ -148,6 +148,16 @@ public function test_get_post_context_without_permisson() {
...
+	public function test_get_post_with_password_without_permisson() {
+		$post_id = $this->factory->post->create( array(
+			'post_password' => 'always$inthebananastand',
+		) );
+		$request = new WP_JSON_Request( 'GET', sprintf( '/wp/posts/%d', $post_id ) );
+		$response = $this->server->dispatch( $request );
+
+		$this->assertErrorResponse( 'json_user_cannot_read', $response, 401 );
+	}
+
```
danielbachhuber: We should have a test for with permission too

rachelbaker: @danielbachhuber We actually already had this covered in the `test_create_post_with_password()` method.

rachelbaker:
We haz unit tests

PR #758, opened 2015-01-03
Default `args` to an empty array

### LINKED ISSUE #740: PHP Warning and Notice on /wp-json/ in develop branch
Reported by Janneke (@avryl) in Slack.

```
Notice: Undefined index: args in ###/wp/plugins/WP-API/lib/class-wp-json-server.php on line 588
Warning: Invalid argument supplied for foreach() in ###/wp/plugins/WP-API/lib/class-wp-json-server.php on line 588
```

I couldn't reproduce locally (even with the proper error levels).

danielbachhuber:
It's always referred to in `dispatch()`, so it needs to be set.

Fixes #740

### DIFF
--- a/lib/class-wp-json-server.php
+++ b/lib/class-wp-json-server.php
@@ -464,6 +464,7 @@ class WP_JSON_Server {
 			'accept_json'   => false,
 			'accept_raw'    => false,
 			'show_in_index' => true,
+			'args'          => array(),
 		);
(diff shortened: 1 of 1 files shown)


danielbachhuber:
@WP-API/amigos #reviewmerge

danielbachhuber:
Merging because it's breaking things.

Issue #718, opened 2014-12-10
Grand unified theory of content types


danielbachhuber:
Rather than debate this philosophically, we produce and debate a proposal for:

* Named content types.
* Appearing in named routes.
* With defined schemas.
* Said schema discoverable through the API.

Related #577, #605

@joehoyle and I will be drafting this, hopefully with some input from @maxcutler .

danielbachhuber:
Grand Unified Theory is well-underway.

PR #759, opened 2015-01-03
Use JSON Schema to define schemas for our Controllers

danielbachhuber:
Publishing schemas for each item (e.g. Post, Term, Comment, User) represented by a controller will allow clients to better consume the API. Because our schema will be standardized, based on the existing [JSON Schema](http://json-schema.org/), WordPress developers can make modification to the controllers without breaking clients. In fact, properly written clients will know how to interpret the modification, provide UI for it, etc.

Each controller's schema will be used to:

* Report expected data format for each endpoint.
* Programmatically generate API docs.
* <del>Validate request input.</del>
* <del>Ensure response data confirms to a schema.</del>

Schemas will be optional for custom endpoints, but enforced for core endpoints. In order for a WordPress developer to add or remove a field, they'll first need to modify the schema. 

**Update 1/31**: For those just coming into the conversation, the entire thread has a good amount of history. At this time the large open question can be summarized as: should WordPress support (and expose) dynamic schemas, or should the core schema be fixed, and any modifications to the schema sandboxed?

See #718


rmccue:
FWIW, we used to have a schema for a bunch of this: https://github.com/WP-API/WP-API/blob/04ce8e36c5180f3a51d4559cd203876f4e91143a/docs/schema.json

Problem is, it never got updated. If we work that into the code, it'll help with that, but it's also a potential problem for developers as a barrier to entry, as it is quite complicated.

plugin.php
```diff
@@ -84,94 +83,98 @@ function create_initial_json_routes() {
...
-		array(
-			'methods'         => WP_JSON_Server::READABLE,
-			'callback'        => array( $controller, 'get_items' ),
-			'args'            => array(
-				'context'          => array(
-					'default'      => 'view',
+	foreach( get_post_types( array( 'public' => true ) ) as $post_type ) {
+		$controller = new WP_JSON_Posts_Controller;
+		$controller->post_type = $post_type;
+		$base = $post_type . 's';
```
rachelbaker: @danielbachhuber Appending an 's' will not always result in the plural form, or even make sense, for many other languages.

danielbachhuber: Yeah, should've marked that as `@todo` — it was a hack I used just to get it working, and should be a point of discussion today.

rmccue:
We should make sure that we allow adding fields without needing to register them, as clients need to handle this case for forward-compatibility already. Adding fields is harmless to the rest of the data, it's only changing or removing fields that should need to fail validation.

That said, should be simple for developers to register their fields if they want to opt-in to the validation on their own as well.

rachelbaker:
@justinshreve Any thoughts or feedback here?

plugin.php
rydurham: Perhaps the Doctrine/Inflector library would be helpful for pluralization?

joehoyle: Hmm, not such a fan of doing _anything_ automatic here - if you register a post type through `register_post_type`, and you never explicitly declare the json route url, it doesn't really follow that the path to your CPT is something _you never entered_. Auto pluralisation is just a bit too magic for my liking.

danielbachhuber: I like having some default behavior if you've set `access_in_json => true`, with the secondary ability to specify the slug if you want.

tests/test-json-posts-controller.php
```diff
@@ -482,10 +482,10 @@ public function test_create_page_with_parent() {
 		) );
 		wp_set_current_user( $this->editor_id );
 
-		$request = new WP_JSON_Request( 'POST', '/wp/posts' );
+		$request = new WP_JSON_Request( 'POST', '/wp/pages' );
```
danielbachhuber: @WP-API/amigos hawtness!

danielbachhuber:
I'm feeling pretty good about this now. Either before or after merge, these are some things we still need to look into:

* [ ] Determine how the schema gets exposed: do we adopt `swagger.json`, or create our own format. Currently, the code only implements JSON Schema.
* [ ] Determine how docs can be automatically generated from the schema.
* [ ] See whether we can use the schema to auto-validate request somehow. Started in #768
* [ ] Define query parameters via a class method too, and automatically expose those.

lib/class-wp-json-posts-controller.php
```diff
@@ -106,4 +869,196 @@ public function prepare_item_for_response( $post, $request ) {
...
+						'description'     => 'The format for the post',
+						'type'            => 'string',
+						'enum'            => get_post_format_slugs(),
+						);
+					break;
+
+			}
+
+		}
+
```
rachelbaker: @danielbachhuber Why for Posts are you not providing a `required` array, like you do for Users and Terms?

danielbachhuber: Oh, it's kind of pseudo-code right now. Need to decide whether arguments are only required for Create requests, or whether we should support a use case where arguments are required for updating (beyond just the ID)

There's a PR open to this effect that branches off this branch. I'm on my phone and don't have easy access to the number.

danielbachhuber: #768

rachelbaker:
I like the idea of having `.../schema` routes available, regardless of if we decide to use Swagger or another third-party API documentation service/standard.

danielbachhuber:
> I like the idea of having .../schema routes available

Yep — seems like a quick and useful way of doing it.

rachelbaker:
@WP-API/amigos Any ideas/suggestions around how the schema properties would reflect different user permission levels?

My original thought was to use `"additionalProperties"` to include properties that may be included in a `context=edit` response if the user's permission level allow.

lib/class-wp-json-posts-controller.php
```diff
@@ -106,4 +869,196 @@ public function prepare_item_for_response( $post, $request ) {
...
+			);
+		foreach( $post_type_attributes as $attribute ) {
+			if ( ! post_type_supports( $this->post_type, $attribute ) ) {
+				continue;
+			}
+
+			switch( $attribute ) {
+
+				case 'title':
+					$schema['properties']['title'] = array(
```
rachelbaker: @danielbachhuber Shouldn't 'title', 'excerpt', 'content', and 'guid' be objects, with 'rendered' and 'raw' properties?

lib/class-wp-json-posts-controller.php
```diff
@@ -106,4 +869,196 @@ public function prepare_item_for_response( $post, $request ) {
...
+						'description'     => 'The title for the Post.',
+						'type'            => 'string',
```
danielbachhuber: @rachelbaker not sure where your comment went, but yes, this schema is actually incorrect :)

danielbachhuber: (meaning, title should actually be an object instead of a string)

lib/class-wp-json-posts-controller.php
```diff
@@ -1,9 +1,741 @@
...
+	 *
+	 * @param object|string $post_type
+	 * @return bool Is post type allowed?
+	 */
+	protected function check_is_post_type_allowed( $post_type ) {
+		if ( ! is_object( $post_type ) ) {
+			$post_type = get_post_type_object( $post_type );
+		}
+
+		if ( ! empty( $post_type ) && $post_type->show_in_json ) {
```
rachelbaker: @danielbachhuber For #769, you will want to change this conditional to $post_type->access_in_json.

rachelbaker:
@WP-API/amigos I compiled my feedback here for review/comments.  

I agree that a schema is **necessary** to provide structure to the API.  In fact, I agree with all four of @danielbachhuber's goals in this PR:

> Report expected data format for each endpoint.
Programmatically generate API docs.
Validate request input.
Ensure response data confirms to a schema.

I spent some time digging into this PR last week.  @danielbachhuber has a good start going here, but it is just a start.  I think Daniel has done a great job of putting this proof of concept together.  

I do not (*yet*) see the benefit over where I thought we were going with defining request parameters in each endpoint registration function.   Maybe that is because I don't see the full vision in what has been put together in this PR so far.  

Some questions:

* Is `get_item_schema()` only intended to get used on CREATE/UPDATE requests?  
* How will we validate the request parameters for GET (single and multiple) requests?
* Can we rename this method `get_request_schema()` or something related to distinguish it from a response schema?

I see the next steps here as:

* The schemas for Posts, Users, and Terms should be completed (or at least remove the schema validation until they are correct).
* Review feedback from at least one or two others (@joehoyle, @rmccue, @dd32, @justinshreve)

rachelbaker:
> I do not (yet) see the benefit over where I thought we were going with defining request parameters in each endpoint registration function.

Clarifying what I mean here:

I thought we would add parameter type & description to endpoint args arrays.  Example

```
	register_json_route( 'wp', '/terms/(?P<taxonomy>[\w-]+)', array(
		array(
			'methods'     => WP_JSON_Server::CREATABLE,
			'callback'    => array( $controller, 'create_item' ),
			'args'        => array(
				'name'        => array(
					'required'     => true,
                                        'type'           => 'string',
                                        'description' => 'The name of the Taxonomy.',
				),
				'description' => array(
                                       'type'           => 'string',
                                       'description' => 'The description of the Taxonomy.',
                                ),
				'slug'        => array( 
                                        'type'           => 'string',
                                        'description' => 'The slug of the Taxonomy.',
                                 ),
				'parent_id'   => array(
                                        'type'           => 'integer',
                                        'description' => 'The parent ID of the Taxonomy.',
                                ),
			),
		),
	));
```

danielbachhuber:
> I do not (yet) see the benefit over where I thought we were going with defining request parameters in each endpoint registration function.
> How will we validate the request parameters for GET (single and multiple) requests?

Right — communicating *request* parameters is a different boat. We could have a `get_query_parameters()` that returns a similarly structured array, however it's really only practical for collection requests, not individual items.

It probably makes sense to keep registering request parameters a part of `register_api_route()`.

> Can we rename this method get_request_schema() or something related to distinguish it from a response schema?

It's the schema for the *Resource*, which remains the same on `GET` and UPDATE`. It's conceptually independent of the full format of the response or request. For instance, in the RAML docs the JSON Schema-based resource schema is embedded within the response schema:

image

danielbachhuber:
> I thought we would add parameter type & description to endpoint args arrays.

Seems like a reasonable way to get to the same end goal of having more structured data about parameters.

lib/class-wp-json-posts-controller.php
danielbachhuber: Hm, seems like tests should've failed for #769 then. Suspicious.

lib/class-wp-json-terms-controller.php
```diff
@@ -212,4 +212,43 @@ protected function check_valid_taxonomy( $taxonomy ) {
...
 
+	/**
+	 * Get the schema for a term
+	 *
+	 * @return array
+	 */
+	public function get_item_schema() {
+		return array(
+			'title'                => 'Term',
+			'type'                 => 'object',
```
joehoyle: @danielbachhuber could you give an example when `type` is not `object`? I guess collection endpoints?

lib/class-wp-json-posts-controller.php
```diff
@@ -106,4 +869,196 @@ public function prepare_item_for_response( $post, $request ) {
...
+			/*
+			 * Base properties for every Post
+			 */
+			'properties'           => array(
+				'id'               => array(
+					'description'  => 'The unique identifier for a Post.',
+					'type'         => 'integer',
+				),
+				'type'             => array(
+					'description'  => 'An alphanumeric identifier for a Post unique to its type.',
```
joehoyle: nitpicking but types and slugs are not alphanumeric.

joehoyle:
Okay, after thinking on this for a while I think I should put down my thoughts about this general approach. I think it's a pretty big decision and has a lot of ramifications in terms of forwards compatibility, development effort to create endpoints and of course consumer libraries and how they typically interpret the data.

The below are some general thoughts as I don't see them outlined in the conversation so far, and also come technical opinions on this implementation. 

Typically *most* APIs (at least in my experience) don't provide this level of abstraction for defining data, it seems 90% of the time the API is fixed, the WordPress.com API is like this for the most part - posts are at `/posts` etc. I think this is how most developers / api integrators think about APIs. Find a new API - where is the URL for "products" at (from documentation), code for that. Compare that to say the Amazon Web Services which is very declarative with another level of abstraction. 

This means an API client typically won't have anything hardcoded for routes, fields etc etc. Fewer things can be _presumed_ about the API so the client has to built an abstract system for interacting with the server. This leads to the "system" not inherently knowing what a "post" is, or a "page", rather a collection of defined schemas that conform to the system in place.

Depending on how you like to code I think the above can be seen as a good thing, or a bad thing. It's more versatile, but more complex. For example, if I am working with integrating an API I groan when I see more abstraction - because there is a system / schema defining available routes, what ones contain what object types; it gives the impression that I need to be handling the case of endpoints moving, fields changing - nothing seems "fixed". This means I can't presume `/posts` gives me Post objects, I can't rely on intuition or experience of other APIs, '/posts/<id>' seems like it _should_ return a single post; but having this level of abstraction doesn't give me confidence that it will be.

I'm not saying that this means we shouldn't take this approach - rather to shown there is some tradeoff. For example, does this mean telling everyone they should build clients that draw from the abstract schema rather than making assumptions make the API seem better or worse? If we don't do that, what is the value in all this abstraction anyway?

I think if there was a project that needed this level of abstraction, the WordPress one would be it - as any site built on WordPress can be **so** different from another. One might not have Posts, Pages or taxonomies with them. They might be using CTPs in a totally different way and provide API routes not resembling the default ones. However, it's worth noting our abstraction only provides so much flexibility - to work with API clients, any "custom" stuff I build would still need to fit into the field types and data types that have been pre-defined. Registering my own custom types doesn't do anything for existing WP-API clients that don't inherently understand those field types. That's one problem I see with this approach - we are trying to come up with a declarative specification for the purpose of supporting modifications / customizing of responses without truly knowing that those modifications could be (or ever will know).

On to the technicalities of supporting a schema driven approach - my main concern is with the added legwork needed to support a schema. Thought I appreciate this is *opt in*. It might just be me as a lazy developer! It seems now you would _start_ by defining your schema for an object type, then implement say `get_item` but your implementation would need to interpret the schema on the fly to see what it is outputting as the schema can be modified; i.e. another level of abstraction. I presumably can't make an assumptions of fields on `update_item` as again the schema could have been modified.

The added overhead means you essentially can't do any "passthrough" controllers where you just return 'get_post( $id )` for `get_item` as you can to take into account schema changes. Though that's not really a negative on this implementation, more a fact about what we have to go through to say an object is a "post" without just inherently treating it as one.

Personally, I hate too much abstraction and prefer to be procedural and be able to iterate quickly with modifications. I feel like this system would slow me down in that - so I don't think I would use this in projects where I am developing bespoke endpoints. Again - not a case for Core not using it, just the perspective from Developer doing custom work.

So after all this rambling, I'm still not decided on whether this in the long run is going to be a good decision of a bad one. I feel like it's defining a precedent for the perception of the API, turning it into an abstract definition of objects with field types from a predictable "fixed" more restrictive simple API. From my experience, I for one haven't cracked the abstract approach in much that I do and more often than not back myself into a declaratively small boxed corner where I run into unforeseen use cases I didn't initially consider.

I think my undecided-ness is stopping me from getting excited about this and to start help coding it. I see what still needs to be done and have a pretty good handle on that - so I think the bigger question for me is the approach rather than the implementation, the implementation is looking very good IMO.

danielbachhuber:
To mention here what I emailed Friday night, adopting schemas for core primitives also opens up the possibility of dropping meta endpoints entirely. Instead of exposing meta endpoints, the expected pattern is that a developer would register given metadata field to the schema, which would then support CRUD on the metadata.

Given the problems we already have with meta endpoints, this seems like a slam dunk to me. No more issues with public versus private, PHP-serialized data, etc.

Remember, in building this API, we're creating a clean abstraction for interfacing with WordPress. We want to establish the best pattern to follow for each scope we're working in (modeling, authentication, etc), while still allowing some degree of flexibility. We're not just adding CRUD to the WordPress database schema.

justinshreve:
I mentioned in last week's chat that I would try to get a better grasp on this issue. Joe's write up helped me grasp everything a bit better. I think what I originally thought http://json-schema.org/ was for and what it actually is slightly different. I hope this post doesn't get too rambly :).

I'll start off by saying that I definitely support these goals:

* Documenting things about endpoints (reporting expected data, programmatically generating docs)
* Validating request input
* Validating response output

All of these are important. The WordPress.com API currently does all these things with an approach similar to the example here: https://github.com/WP-API/WP-API/pull/759#issuecomment-69692197.

As Joe mentioned the WordPress.com REST API is pretty easy to grasp. If a client wants posts they know to hit /posts. If a client wants media they know to hit /media. A client calling us knows what the structure of "/posts" (because of documentation) is. They knows it shouldn't change on another site because of a plugin. Sure, they can discover it through the abstraction proposed here, but I think it makes things much harder to deal with just because something _could_ change.

In general, I think WP-API might be worrying too much about supporting plugins and an infinite amount of possibilities of how data or endpoints might change. I think the core API (what I consider posts, pages, media, categories, tags etc...) should be the main focus for the API. We can make it easy allow plugins to add fields or endpoints -- but we shouldn't make it easy to change other things. As someone who has dealt with partners using our API I think most clients (say a mobile blogging app) just want that core functionality.

dd32:
I've been thinking through this for a while now since it was mentioned to me, and I think I've got a pretty good grasp on what is being suggested here.

However, I'm not sure I agree with all of the original aims
**Report expected data format for each endpoint.** & **Programmatically generate API docs.**
Documenting the API is obviously required, but using a schema that's used for multiple reasons doesn't seem an appropriately simple solution.
The API doesn't need to self-document itself, however it definately should be available through https://developer.wordpress.org/ for consumption by developers - In order to make it available there, we need to have the data in standard format, but it doesn't have to be held within primarily-machine-readable-code.
The simplest option here that fits all the requirements is PHPDoc prefixing `register_json_route()` call. As an added bonus, WordPress already uses this and has a standardised Parameters/return format available for it. Of course, it's not accessible to the API code itself, but realistically I don't see a need for it to be.

**Validate request input.** This is something that other API's (such as WordPress.com's) does, however, I'm not fully convinced that WP-API needs to strictly validate input data.
Type checking for example (especially as PHP is not strongly typed, and Javascript is not much more strict).
Validating that the data is in the correct format feels like something that should be handled by the individual callback, as the callback has to sanitize it, validate it's correct, validate it's within the correct range, etc.
One exception that seems acceptable here, is that the list of required fields should be able to be set, `'required' => ['name', 'slug']` for example which would return a generic "missing argument" error.
Another function that could be added for added developer convenience could be a `get_parameter_as( 'int', 'query_var' );` method on the `WP_JSON_Request` object that's passed.

**Ensure response data confirms to a schema.** This is something that I feel the developer is responsible for. If a field is supposed to be an int, and a endpoint returns a boolean, should that response fail? Should it be casted to an int? What if a String or Object was returned?  Why should the API handle either of those cases? The client can detect invalid results (or interpret them as appropriate) and the API endpoint itself can self-validate against a schema if it cares to double check it's return data such as `return array( 'id' => (int) $id, 'title' => (string) $title, 'enabled' => (bool) "whatever" )`

So I'm mostly against implementing any kind of schema, as I don't see the need for it yet. Keeping it simple..

rmccue:
I've been following the conversation here, so it's probably time for me to weigh in. :)

Looking back at the four goals:

> * Report expected data format for each endpoint.
> * Programmatically generate API docs.

These two are nice-to-have, but don't necessarily warrant additional complexity. The expected data format would usually be inferred from the URL in a static API (non-distributed), which is a reasonably small piece of data.

Documentation is fantastic, but as @dd32 noted, it doesn't have to be machine-readable. phpDoc would work fine for us here. I'd love to have it available from the API itself (yay, developer discovery tools), but it does add a non-trivial amount of code to do so.

> * Validate request input.

This would be cool, but means we need to be declarative, and we'd likely end up inventing our own Turing-complete language to do so. For example, for inserting posts: you need either a title, content, or an excerpt, but only if the post type is not `attachment` (also: more complicated, because filters). Expressing this complexity is not something we should do declaratively, and I don't think it makes sense to use a schema on the input side.

> * Ensure response data confirms to a schema.

This might help clients, but it's not necessarily a holy grail solution. My biggest issues are the following:

* Robustness
* Complexity

**Robustness**: Everyone knows Postel's Law (also known as the Robustness Principle):

> Be conservative in what you do, be liberal in what you accept from others.

The issue I have here is that we're being a bit too conservative, or rather that we're only considering one side of this. Let's say we're sending off a post response to a client, and they only really care about the permalink and the title. If a plugin removes the comment count field (e.g.), we shouldn't send a blocking error, because it doesn't actually matter to the client. Failing would remove the ability to have more fine-grained failure modes, instead forcing the client to show essentially just "sorry, some error occurred".

This is one of the places where validating this schema would hurt the client. We can't presume to know how every client is going to behave, so we shouldn't assume that every deviation from the schema instantly requires us to fail the request. (It's true that we don't necessarily need schema mismatches to fail a request, but see below under Complexity for more on that.)

I think what we're trying to do here is be conservative with sending API responses, which is great for helping clients with interoperability. The problem is that we're not being liberal on the other side of that by allowing plugins to make changes to the API. We need to assume that plugin developers for the most part are smart enough to know what they're doing, although they may need guiding towards the best practice. It's our responsibility there to provide good UX for the internal API design to facilitate that.

Rather than the hard-handed approach of stopping developers from doing things, it makes more sense to take a soft approach and guide them (strongly) towards doing it the right way. Make it easy to do it as nicely as possible, but possible to do whatever they want.

**Complexity**: One of the benefits of adding schema validation to the API is that we can indicate to clients what a response type is, saving them from having to check all the fields before they use them. However, this doesn't necessarily reduce client complexity.

Not having a schema has the potential to increase error-handling complexity on the client side by making them run the checks, but realistically, they'll be doing this anyway. What happens if they access a WordPress-like API (say, a Drupal compatibility API) that does things slightly differently? What if an API screws up and asserts a schema that's not accurate for the endpoint? Clients need to be built robustly anyway, so this realistically only adds complexity rather than removing it.

Ideally, I think we'd indicate with each response the "type" that the response matches, based on a schema (probably indicated via a header). This wouldn't error out if the schema didn't match, we just wouldn't assert that it had matched. However, well-designed clients would still check that the type did match anyway, in order to be robust, so I'm not sure this gives any gains.

@danielbachhuber noted in the original description:

> Schemas will be optional for custom endpoints, but enforced for core endpoints. In order for a WordPress developer to add or remove a field, they'll first need to modify the schema.

Realistically though, what do we gain by forcing developers to add/remove fields from the schema? If the schema itself is malleable, how does that help clients? It can certainly communicate to the client if a field is a string or an integer, but that's part of the JSON anyway, so we don't gain anything there. (Enums suffer from not being extensible, which is why the original spec.md indicated things like `"post" / "page" / token`, and gave instructions to clients on how to handle values they didn't understand.)

----

I think my main comment here is that I think this gives minimal benefit for what is essentially a huge addition to the API. I'm not sure that right now it gives benefit to API clients nor plugin developers (or at least, whether any benefit it has is worth the detriment).

One thing that I think does make it easier for client developers is helping them to understand "invalid" responses. For example, the original specification specifies what post statuses are available, but notes that others may exist depending on the site, and recommends best practices for handling that. Likewise, we can handle things like missing fields through the same mechanism. Clients can then have nuanced failure modes, where some may ignore a missing field or use the best practice for handling it missing, while others may display an error to the user.

For a way forward: there's some good pieces in here that we definitely want regardless. Checking `post_type_supports` before outputting the fields is a great idea that we should have incorporated earlier. Automatically registering post type routes is pretty decent, but not a fan of the `$type . 's'` construct, so let's see if we can use that and improve it. We should also check the field renaming and see if we want that independently. Anything else non-schema-related that we want to separate out and consider?

danielbachhuber:
I just wrote a really long, persuasive comment on my phone, and lost it when the browser crashed.

In the interest of saving everyone time, let me just say this: I think it would be irresponsible to ship WP-API without schemas. To me, it's a question of the implementation, not whether or not. At the very least we get self-documenting endpoints. Which, given that ours have been hand-maintained and always out of date, is not trivial. I'm not really sure how to respond to the suggestion that PHPdoc is a viable alternative.

Yes this pull request is incomplete. Yes there are outstanding implementation details, which I've asked others to dive into so they can better understand the nature of schemas. Yes we don't have to solve all of the stated goals in the first implementation. Again, having schemas at the core of our endpoints gives us fundamentally more robust code.

We are building a 10 year foundation for WordPress, not a project where it's acceptable to cut corners. WP-API will mostly be used for custom post types, which usually deviate from the core Post data model. Without schemas, we have no pattern of discoverability for these endpoints.

If you, Ryan, decide "no" on schemas, it won't have my blessing, and I'll need to withdraw my responsibility for this project.

maxcutler:
I'm observing two distinct philosophies in this thread:

**Ph1**: WP-API should be a framework for quickly building bespoke JSON APIs. WordPress core should make it easy for plugin authors or agencies to create custom APIs and consume them from their themes' JS code or other custom integration points.

**Ph2**: WP-API should be a globally distributed JSON API network. Each site might make additive customizations, but each installation exposes enough metadata for generic clients to robustly talk to _any_ WordPress site.

It is difficult to optimize the design to serve both philosophies without significant compromises, leading to a repeat of this debate every few months. I think it's time that the WP-API amigos and Core leadership make a decision on which of these two is the desired goal so that progress can be made.

As someone who has worked on mobile clients and written general-purpose client libraries for the past four years, I lean strongly towards Ph2 and I have advocated as such for the past 9+ months on my blog and at several WordCamps. But I recognize that most of the aforementioned stakeholders work at agencies or on Core and will lean towards Ph1.

The project README is wonderfully vague about its purpose, which lets everyone make their own assumptions. It's long past time to put a stake in the ground so that everyone is on the same page and can rally towards a shared vision.

dave1010:
> Clients can then have nuanced failure modes, where some may ignore a missing field or use the best practice for handling it missing, while others may display an error to the user.

Being forced to do this instead of having a spec is horrible from a client's perspective! Null checks and manual type checks all over the code. If an API provides a spec and a client wanted to do this then they can still ignore the spec if they want to.

Having a spec means that plugins and clients know of they're not going to break anything. Without a spec developers will return a response and have to hope that it works.

This API won't just be consumed by dynamically typed languages like PHP and JS.

> I think it would be irresponsible to ship WP-API without schemas.

Agreed 100%.

bkirby989:
It doesn't seem like the issue is schema or no-schema, but a fixed schema vs a dynamic schema. 

The choice is between 
1. An API that primarily does CRUD on existing WP objects, limiting customizations to (probably) additive meta data, or
2. An API that is completely agnostic to existing WP conventions/object types, where the client has to check that those conventions/objects actually exist. But api customization is limitless. 

But is option 2 a schema the way most people think about it at all? It seems like option 2 is really just a schema for making schemas. It both allows, but more importantly, requires the client to build a lower level of abstraction to interact with the database.

Instead of GET post always returning exactly what I think it will, I have to GET the post schema and actually confirm that it does. And I have to do this every time I'm interacting with any object in the API. 

I just don't see the benefit. But I could totally be misunderstanding the whole thing / missing something big.

austinpray:
>The choice is between 
1. An API that primarily does CRUD on existing WP objects, limiting customizations to (probably) additive meta data, or
2. An API that is completely agnostic to existing WP conventions/object types, where the client has to check that those conventions/objects actually exist. But api customization is limitless.

Option 1 seems far more in scope for a WordPress application. There are already excellent frameworks out there for creating bespoke APIs. I would love to hear why anyone would want to use WordPress to serve a custom API because I am not getting it. If you wanted custom data types and fields that have nothing to do with WordPress, I would go as far as to create a separate sister API in a modern web application framework. Separate your concerns and let WordPress do what it is best at.

I would expect API customization to be all about augmentation. That way people can be create client tooling and libraries knowing that the base API is always going to behave as expected. Otherwise when creating clients people will have to reinvent the wheel every single time.

danielbachhuber:
> It doesn't seem like the issue is schema or no-schema, but a fixed schema vs a dynamic schema.

Almost. I'd suggest it's _unknown_ schema vs. dynamic schema because, for the former, it will still possible for a WordPress developer to add or remove fields from a response. A client developer will need to accommodate this behavior without being about to expose the schema in advance.

> It seems like option 2 is really just a schema for making schemas. It both allows, but more importantly, requires the client to build a lower level of abstraction to interact with the database.

I think this is a fair statement. It's worth noting, and re-emphasizing, using this schema is opt-in for WordPress developers. The API will also support simply supplying a PHP closure for a callback on a type of request, and think nothing of it.

> I just don't see the benefit. But I could totally be misunderstanding the whole thing / missing something big.

For me, it boils down to this: WordPress is mostly custom content types, not Posts and Pages. For instance, [Fusion](http://fusion.net/), where I'm the Director of Engineering, has Stories, Videos, Galleries, Comics, Interactives, and other content types. I'd like editorial to be able to use the WordPress for iOS app, or any other API client for that matter, to publish to these content types.

Some WordPress content types are also, currently, poorly represented through the API (e.g Menus and Attachments).

I also maintain WP-CLI. As a WP-CLI user, I'd like to be able to list, get, update, and delete the primitive data types for a site, and have the interface for each resource exposed to me without me having to bash my head about discovering it. Think `wp event --help` and `wp gallery --help`.

bkirby989:
> For instance, Fusion, where I'm the Director of Engineering, has Stories, Videos, Galleries, Comics, Interactives, and other content types. I'd like editorial to be able to use the WordPress for iOS app, or any other API client for that matter, to publish to these content types.

Could you explain how the proposed schema system would accomplish that? 

You would still have to build a plugin that extended the API to include separate routes for those content types (using CPTs?) right? - A plugin that itself had a fixed schema? Are you just proposing this as a programmatic way to retrieve custom schema parameters?

joshkoenig:
_Chiming in from the far far peanut gallery. I'm a [relative WP n00b)(https://www.getpantheon.com/blog/notes-drupal-ninja-wordpress-n00b), but I've built quite a few APIs, and consumed an order of magnitude more, so take this for what it's worth._

I don't know how to weigh the WP internals of this debate, but I can say a few things about APIs in general from experience:

If you plan on people customizing WP-API as even a 20% use-case, making those customizations programmatically discoverable and self-documenting is probably more important than anything else you could ever consider if you want widespread adoption. 

This can be done with minimal overhead to developers who are implementing. It's easy to read this architectural suggestion as over-reaching, both in terms the amount of abstraction and in terms of the scope (e.g. needing a turning-complete language for validation). I'm sensitive to @joehoyle's points. 

However, I believe the intention is to make a number of good things standard / easy / automagic for developers, not put a bunch of repetitive scutwork in the way of building something.

1. Self-documenting endpoints: as per above, can't say enough about that. They will increase adoption  by an order of magnitude by making it a common process to interact with the API.
2. Programatic validation: not from a security standpoint, but from a schema-match standpoint. It makes testing oodles easier if you have a common machine-readable way to expose the spec for your endpoints.
3. Future-proofing: the whole point of having an API is to integrate with other tools and services. By allowing other things to understand the API without having to know WordPress (by using internet-wide conventions) you increase your addressable audience significantly. 

I believe the spirit of this proposal is necessary if you intent this effort to make WP integrate with more things "not built here", and/or if the intent is to further the vision for "WP as an App".

That's my outsider's $0.02. A suggestion for moving forward would be to list out the common use-cases you have in mind as developers, and do the thought-experiment of "how would we satisfy this requirement?" Talking through that can really help clarify things. 

Good luck, and thanks to everyone for the hard work. I'll be watching excitedly. :)

danielbachhuber:
> Could you explain how the proposed schema system would accomplish that?

Core would register endpoints corresponding to these post types and expose core fields defined by my use of `post_type_supports()`. As a WordPress developer, I would then register the additional fields I'd like on the schema for the resource (e.g. `embed_html` for videos, or `image_ids` for galleries). In doing the field registry, I as a developer would provide callbacks for returning and updating the data.

> You would still have to build a plugin that extended the API to include separate routes for those content types (using CPTs?) right? - A plugin that itself had a fixed schema?

Yes, and yes.

> Are you just proposing this as a programmatic way to retrieve a custom schema definition?

Not sure I follow this question.

brianfeister:
As a very low level WP developer, but a full time JavaScript developer who consumes and writes APIs all day long for custom web applications, here are my 2 cents:

Not having a self documenting API would be a huge mistake. My feeling is that this project actually turns WP (a platform that doesn't really meet most of my needs these days) in a whole new direction. For writing a web application I would of course continue to work with tech like Angular and Node. But if I had some mid-budget project I could quickly and easily build out a CRUD utility that compartmentalizes something like "public user adds an event to site and submits for admin approval."  Right now that is incredibly painful because I need to do all of the work to customize the admin interface based on this unique role. And even then my branding capability remains limited without tons of work. Also I could create a simple JS poll that is public and tie it to a data visualization. 

Sorry, too many examples. Back to the point. The API spec itself needs to act as a way of normalizing the relationship between core and client. Without that, the appeal to folks like me who got tired of WP core is greatly diminished. I'm not sure the target audience, but I suppose (perhaps like everyone else) that it feels aimed at front end devs (I am one) who want to write their own interfaces and leave WP to manage content and do work for higher privilege admin functionalities.

brianfeister:
^^^ in support of schemas, for what it's worth.

rmccue:
I'll start some quick inline comments, with a more meaty discussion below. :)

@dave1010:
> Being forced to do this instead of having a spec is horrible from a client's perspective! Null checks and manual type checks all over the code. If an API provides a spec and a client wanted to do this then they can still ignore the spec if they want to.
>
> Having a spec means that plugins and clients know of they're not going to break anything. Without a spec developers will return a response and have to hope that it works.

Properly written clients need to deal with this possibility anyway. Blindly trusting received data could cause the client to crash if a proxy interferes (or if you're interacting with a different WP version).

> This API won't just be consumed by dynamically typed languages like PHP and JS.

I've programmed in statically typed languages (and am working on projects right now using them), so I understand the concern here. However, typing itself is built-in to JSON itself, so the schema isn't adding anything here. The schema declares fields as strings or numbers, but JSON natively encodes those. Enums are one of the "special" types that provide more data, but WP is extensible enough that these are practically useless (custom post types, custom post statuses).

There are other types available here, such as datetimes, but these can be handled by pretty much any parser anyway.

@maxcutler:
> I'm observing two distinct philosophies in this thread:
>
> Ph1: WP-API should be a framework for quickly building bespoke JSON APIs. WordPress core should make it easy for plugin authors or agencies to create custom APIs and consume them from their themes' JS code or other custom integration points.
>
> Ph2: WP-API should be a globally distributed JSON API network. Each site might make additive customizations, but each installation exposes enough metadata for generic clients to robustly talk to any WordPress site.

Our central use case is the site's users, and everything follows from this. Both of these use cases (custom APIs and distributed APIs) need to be considered, but the obvious one to prioritise is distributed APIs. The long tail of sites contains the majority of users, so it's where we need to focus attention.

My concern is more that having fluid schemas (as proposed here) is actually worse for this. We increase client complexity for (as far as I can tell) no real benefit.

----

I think I may have miscommunicated with my comment, but @bkirby989 summarised it effectively:

> It doesn't seem like the issue is schema or no-schema, but a fixed schema vs a dynamic schema.

I absolutely agree with having a fixed schema if we can, because it means you can do minimal validation on the client side (note: you still need at least some validation, but with a fixed schema, this can easily be provided by a library or other tool).

My main issue is with having a schema that can be changed; that is, a schema generated by the server itself, rather than a "standard" schema that it conforms to. If the server is generating the schema, it doesn't help clients, because clients need to first parse the schema to understand what fields are available.

Let's say I add a field, and the corresponding field is then added to the schema. How does this help the client? Yes, they can now check the schema to see "oh, this field is now there", but they could just check the response for that anyway. They don't gain meaningful information, apart from a human-readable string about the field. (No one dislikes documentation, mind you, the argument is that we can explore non-schema-related ways to handle this.) On the flip side, let's say a field is removed from the schema and hence from the response. Couldn't you just check whether that field exists on the response instead?

The way I'd prefer us to handle the schema is similar to the HTML specification. It defines what you should expect to get back, as well as indicating what the parts inside it actually mean. However, it also provides a robust error-handling model to follow, which is a lesson learned the hard way (everyone loves browser compatibility). We would return requests saying "the response should be handled as per this schema", indicating that we believe it should conform. Clients still need to take responsibility for ensuring it does, but the way to handle failures is clearly defined.

----

To those commenting on this issue: thanks for your feedback! Do keep in mind though that this pull request relates specifically to **adding generated schemas returned by the API**, which can be modified by plugins similarly to the response itself.

danielbachhuber:
Argument for WP-API schemas

To make sure my points are appropriately communicated for tomorrow's conversation, here's a summary of what I'm proposing:

* Every WordPress primitive (Post, Page, User, Term, Attachment, etc) has a defined schema. A schema is essentially a name associated with one or more fields, each of which has basic type defined and a human-readable description.
* We adopt JSON Schema for our API, which has existing widespread adoption.
* Internally, these schemas are used to some degree for the mechanics of the  WP_JSON_Controller class.
* Externally, the schemas would be exposed at wp/posts/schema, wp/users/schema and similar.
* The schemas would be used to programmatically generate human-readable documentation.
* In order to add or remove fields from a primitive's response, average Joe developer would need to register those changes to the schema. We'd remove the existing controller class filters.
* Endpoints for custom post types would be registered when access_in_json is set for the post type. The default schema would reflect use of post_type_supports
* Average Joe developer can still register an arbitrary callback to an endpoint. If he did so, no schema would be exposed.
* Average Joe can extend a primitive's controller class to use it on his custom endpoint.

Why I'm proposing these points:
* We should optimize WP-API for those building clients intended to be used across many WordPresses (think apps, themes) because they'll have the least amount of flexibility, tolerance for inconsistencies, etc.
* Custom post types, and implied, custom data models, are equally the future of WordPress as Posts and Pages. We should give them emphasis during the development of WP-API. Not doing so in the past has given us the mess of post meta. This is actually an opportunity to create a new interface to custom post types. 
* Schemas give us a robust internal and external framework for dealing with non-Post and Page data models.
* Average Joe developer can still register one-off endpoints and callbacks for his bespoke project, but schemas give him a strong pattern to follow when he wants to ship a plugin to 23% of the web.

Note that querying collections of resources isn't covered in this proposal, but I suspect we can implement it as some derivative of the schema implementation.

I'm not proposing these at this time, but they could be cool later:
* Expose all of our endpoints using swagger.json
* Magically validate request data using a domain-specific language (beyond those attributes supported by JSON Schema)

It's important to note that no one has distributed an API to 23% of the web before. We're in uncharted territory to some degree, and we should keep that in mind as we discuss what's _ideal_.

dd32:
I do agree with what @danielbachhuber stated here:
> For me, it boils down to this: WordPress is mostly custom content types, not Posts and Pages. For instance, Fusion, where I'm the Director of Engineering, has Stories, Videos, Galleries, Comics, Interactives, and other content types. I'd like editorial to be able to use the WordPress for iOS app, or any other API client for that matter, to publish to these content types.

There does need to be a way for custom content types to be discoverable by clients, and create them, I don't think anyone will disagree with that statement, even those who have voiced disagreement with a schema here.

What I do disagree with, was what the original PR presented.
I do not think that
- All outputs from endpoints should be validated against a schema, even if it's a schema defined on-the-fly
- All outputs from a standard endpoint should be determined by a schema
- All post-type endpoints should be handled by the same class even if they have significantly different outputs (although Custom Content Types could re-use that same code, so using `post_type_supports()` as appropriate is expected)
- That the API itself needs to provide the documentation for how to access it
- That the API should provide machine-readable documentation which is then used to generate human-readable documentation
- That clients should not be expected to handle extra fields in a response

I do think that
- WordPress should have a range of published schemas that describe the default output of an endpoint
- All public endpoints should match their defined `type`, if the developer specifies one, and that this should be exposed on the endpoint listing, or, in a specific endpoint that describes the types of content that the site supports. If a developer fails to provide the type, and the client does not expect that route, it should be ignored.
- All endpoints should strive to output data according to their pre-defined type, but that it's not a fatal error (either by clients, or the endpoint itself) if it fails to match them. A client may choose to error on such an endpoint, or it may decide to handle it if it can.
- All public endpoints should optionally provide a link to documentation on how to utilise it (So for example, the endpoint type could be a URI, or there could be a specific documentation field)
- Documenting the API should be done in a published format, but not necessarily machine-readable, although could optionally be presented in both. Examples of usage, full range of inputs, full range of output and error conditions, all logic that the endpoint lives by, etc needs to be documented and placing that all within a schema does not make logical sense for a publishing platform. The HTML DTD, and RFC's do a good job of stating what to expect from something, but very few clients will ever hold a response strictly to said documentation and instead interpret it as appropriate to their application.

maxcutler:
>> Let's say I add a field, and the corresponding field is then added to the schema. How does this help the client?

The primary scenario that I've brought up in previous conversations: mobile clients (esp. the offical Automattic-developed apps) and a future WP-API-powered WP admin need a way to dynamically generate custom field input controls. Having a machine-readable schema makes this orders of magnitude easier and user-friendly.

I spent four months back in 2012 looking at adding custom field support to the official Android WP app using the XML-RPC API. There's basically no way to do post meta editing in any reasonable manner without a schema; the Poster app made the best attempt to-date by allowing the end-user to manually configure their custom fields, but this doesn't scale across a team very well and is a hassle to users.

You dismiss the value of enums and other simple enhancements over JSON's data types, but even small amounts of additional schema metadata allow a client to produce a more accurate and user-friendly interface.

I don't understand the appeal of manually-written docs over a self-documenting API, but that's really a secondary benefit of schemas IMO. I wrote all of the XML-RPC API documentation on the Codex (including retroactively documenting the previous 8+ years of its existance) and I would never wish for a person to have to do the same for a new API. Sarah has done a great job, but it's not a scalable process and I can't imagine developers are ever going to do an adequate job publicly documenting their API customizations.

rmccue:
> I spent four months back in 2012 looking at adding custom field support to the official Android WP app using the XML-RPC API. There's basically no way to do post meta editing in any reasonable manner without a schema; the Poster app made the best attempt to-date by allowing the end-user to manually configure their custom fields, but this doesn't scale across a team very well and is a hassle to users.

One of the reasons for this is that core WP doesn't support field types. If/When WP core supports different types here, it would make sense to then expose this via the API, but that's an issue the metadata team was looking at. We don't want to bite off more than we can chew here.

> You dismiss the value of enums and other simple enhancements over JSON's data types, but even small amounts of additional schema metadata allow a client to produce a more accurate and user-friendly interface.

My issue with enums specifically is that they can lie to the client. For example, saying that `status` is an enum discounts the fact that plugins can add and remove these at will. Better than that is that plugins can change what they do on each request.

I'd love to give more information to clients, but it's important to never give information to clients that's straight up wrong.

joshkoenig:
Just wanted to call out what I see as a difference of opinion in how this would be used/useful:

> Let's say I add a field, and the corresponding field is then added to the schema. How does this help the client? Yes, they can now check the schema to see "oh, this field is now there", but they could just check the response for that anyway.

From my perspective, the virtue of this proposal is not making automagically-adapting clients; though schemas would be a pre-req, the singularity is still a ways off. The main virtue isn't for a client to automatically "notice change", or for developers to write clients that query a schema before querying a response. 

The main imminently tangible virtue is that as a client developer, that there's a common, reliable way to discover a schema that communicates itself in web-standard terms. "Checking the response" to see what an API does is effectively reverse-engineering, which is not the ideal way to develop an API client. 

Further, there is one highly viable client type that actually could make programatic use of a schema immediately: a test. Having schemas makes mocking responses from an API a quick bit of work; ditto generating basic client test cases. Trying to backfill these things without the benefit of automation is no bueno. 

All that said, I'm certainly sensitive to what @rmccue is saying in terms of "this information could be wrong". The question there boils down to design tolerance. It needn't be 100% correct to provide a lot of value, just close enough, especially if the ways that it can be incorrect are understood and documented. 

I don't know how frequently a schema might be rendered inaccurate due to internals, but maybe this could become something of a standard practice? "Don't hack core" <---> "Don't break the schema"?

jeremyfelt:
I think providing a schema is a good idea and does more good than harm.

I've only consumed the current WP-API (1.1.1) with one off clients written specifically to handle one off scenarios on individual site. This is the kind of work that doesn't need a schema, though I don't think it would have gotten in the way.

I imagine there are many other complex scenarios waiting to happen. To me, the real power of a new API for WordPress is in handling these scenarios, not necessarily in providing for my use case. To write a client that can be used or reused on many different sites, having a schema would mean less uncertainty overall.

This example might be a stretch, but it's the one I've been using when thinking this through. :smile: 

If I have a client app that shows all items from all public post types that either have a post thumbnail or could be assigned one.

1. Schema-less, as I understand it, would mean—go through a discovery process to find all post types, retrieve a post from each, determine if a field for post thumbnail is included, query for more posts of each of these post types, display the view, and cache "supported" types at the app level for future requests.
2. Schema would mean—Use the schema to see the post types supporting post thumbnails. Query for posts in these post types. Display the view.

So, +1 for JSON schema from me. Auto generated API docs are a nice side benefit. As far as validation, etc... I'm still thinking through my thoughts on that.

rachelbaker:
Update with where we left off in our discussion yesterday.  

Things we all (seemed to) agree upon:

1. Through documentation for endpoints that includes types and descriptions for query (GET), request (PUT/POST), and response parameters.
2. It should be easy to register additional endpoints that are silo'd from the "core" endpoints.  These silo'd endpoints can be used by custom post types or plugins.  
3.  It should be difficult to remove fields from the responses of the "core" endpoints.  
4. It should be easy to add fields to the responses of the "core" endpoints.

Where there was a differing of opinion:  

1. Should responses from the "core" endpoints be strictly held to a schema?  Is it reasonable to require that any additional fields and their data types be registered with the schema for the core endpoints?   
2. Should request input from endpoints be strictly held to a schema?  Example: if a site configured the page post type to not support `editor` but `content` is passed in a POST request should we discard it or return a fatal response?
3. Is a dynamic schema the only answer for allowing clients to discover endpoints and query/request/response parameters? 

Where we ended up:

After some respectful discussion, we decided to hit pause (for now) on the concept of a dynamic strict schema.

danielbachhuber:
Here's a new pull request for the non-controversial parts #820

Once that's merged, I'll put together a fully-functional PR for JSON Schema. I think part of the problem is that we're still debating a lot of hypotheticals — it will be easier to discuss when everyone can see the code.

danielbachhuber:
In my mind, the remaining open question is: Should WordPress support (and expose) dynamic schemas, or should the core schema be fixed, and any modifications to the schema sandboxed?

I've updated the description to this effect.

tollmanz:
I'm late to the party, but since it seems that this conversation will continue in the future, I'll jot down some of my thoughts. I'm mostly responding to themes in this thread since I have the advantage of reading a month's worth of discussion before responding.

**Usability/User Experience**

I'm approaching this issue from the perspective of someone who has worked with many external JSON APIs, but surprisingly, has not worked with JSON schema. Thus, I've had to do research into the concept in order to inform my opinion. I think this is an important perspective as we will have a lot of devs approaching the JSON schema issue in our API from this same perspective. 

JSON schema is easy to grok from an implementation standpoint. With about 5-10 minutes of reading, I could easily understand it and felt confident in being able to implement it. With a solid API UX for updating our schema, I think that developers will not have a problem updating JSON schema for their bespoke needs. As we've seen in the WP community, an authoritative article on the subject with good placement in the community can easily train devs. *I do not think that the increased complexity from a UX perspective is a good reason not to implement this schema*.

**Documentation**

I really like the idea of using this for documenting the API for an individual site. The docs are accessible for developers reading code *and* via the API. Additionally, I think this is easier to standardize than PHPDoc. Individual developer use of DocBlocks is all over the place. PHP doesn't care how you implement them so it is hard to enforce them. If we have an API that is more forceful about the implementation, we can *almost* guarantee doc quality. I think it's a win/win from a docs standpoint.

**Dynamic vs. Static Schema**

I really like @maxcutler's comments about the two different use cases for the API (bespoke vs. network of JSON APIs). The API should be designed both both use cases, because, damn, that's powerful. My question is then, *can we have both static and dynamic schema?*. Having both helps for the two uses cases.

Let's say I'm developing a hot new Blackberry WP app and I want to be able to connect to any WordPress site to read data. I want to make this easy, so I decide to only handle core content (e.g., posts, pages) via the API. Couldn't we have a canonical source for the *core* schema provided by WordPress.org? The app could rely on that source for verifying the JSON structure returned by individual blogs. This would help verify that a site has readable data in the format expected and my app could read from that blog.

Now let's say @danielbachhuber is getting his Fusion on and designing an iPad app to read comics and only comics from Fusion's API. We need a way to extend the JSON schema to support this use case (although, in bespoke apps, the JSON schema may be of less utility) for custom content. Dynamic schema would help with this. Imagine if they then open sourced their comic server and client, having the schema would be extremely useful in developing an app that discovers the comic  superpowers of a WordPress site that allows for aggregation in a single app.

I think there are good reasons to have static and dynamic schemas and we could probably support both.

**Data Validation**

As I am just learning about JSON schema, I am really worried about numerous tutorials and references to using JSON schema to validate API data. When digesting data from an API, I usually go into tinfoil hat mode. An angry employee, a compromised network, a partial response, or a simple human error can all result in API data that can cause strange or even malicious behavior in a client. If the JSON schema is used as the source of validation for this data, I am not sure how your app could be secure given that it is susceptible to these same data integrity and authenticity issues. Even digesting these APIs over secure connections are susceptible to MitM attacks and all of the issues mentioned above. I just cannot envision a world where I could trust an API to explain the API to me in such a way that I would programmatically use it for validation. I would prefer to validate for my specific use case.

Note that I'm not advocating that we don't use schema for discoverability or for suggestions about the conformity of data. I just do not think it should be used for the type of validation and escaping that one might normal do when constructing HTML documents. My main concern is about messaging around the JSON schema implementation. I hope that we are careful about how we talk about its security implications.

**Conclusion**

I am in favor of JSON schema for the WP API in general and think that it is worth the increased effort for building the API. I do not see how it hurts the API other than keeping it from being integrated into WP core because of the extra work needed to build the API.

wycks:
+ 1 for JSON Schema and at least a basic ability for developers to extend/create query based data discovery. The more accessible and easy the API is, the more traction it will generate.

I urge everyone to grab a big coffee and watch this video: http://www.infoq.com/presentations/web-api-html.

ps. I have worked with two large API's recently (closed source commerce applications),  both of which had poor documentation and were rigid.  The biggest problem was they had non-existent data validation customization, this hurt the revenue stream and user experience a great deal.

danielbachhuber:
Thanks for the video, @wycks. We'll check it out.

debrouwere:
@tollmanz Not sure I follow re: data validation. In cases where JSON schemas are used to determine whether an API reponse contains what it says it should contain, it's all simple type checking (is this field present and is it an integer?) with no opportunities for arbitrary code execution or anything of the sort. You're not handing over the keys to your computer to the schema :-)

Of course it's smart to layer your own validation on top of that, using the schema-based validation as just a preliminary sanity check.

API response validation is really just one use-case. Another that I find more interesting is e.g. using a JSON schema on the client side for instant feedback when filling out a form in a custom web application built on top of this (or indeed in WordPress itself). In this case, too, the client side validation serves as user feedback but no more, the real validation should happen on the server where it can't be tampered with. Heck, not just validation, you can even use it for form generation: https://github.com/joshfire/jsonform

debrouwere:
I'm not a part of the WordPress community, but I've done a lot of API work in the past, so Daniel asked me if I could share some thoughts.

Many people have shared examples of the power of using schemas: automatic documentation generation and even automatic API explorers (see e.g. some of the projects at http://raml.org/projects.html), shared client- and server-side input validation, a foundation to speed up the development of API clients and so on. It's DRY nirvana in that sense. And not a dream either, this is exactly how http://www.django-rest-framework.org/ and other REST frameworks do things today.

On the other hand, don't underestimate the engineering effort. Not just in setting up the schemas and in providing a good way for other developers to extend those schemas or even provide their own ones (which, for non-content API endpoints, makes sense), but in making sure that all of those wonderful advantages that exist in theory (from validation to form generation to practically-for-free API clients) will also exist in practice. There might be more work on the tooling than you think. And if none of the supporting tools are there, you're really just asking people to do extra work for no return.

So the big question, to me, is not a technical one and it's also not a question of which is the superior approach, but rather: to what extent is the broader WordPress community willing to support the better-but-harder approach?

Just my two cents.

danielbachhuber:
Resource schemas dropped in #844

They're generally exposed for documentation purposes at `/wp/<type>/schema`.

We're still discussing how new fields should be registered, and whether data could be added to a response without registering appropriate schema #875

Schemas will also likely be used to perform basic request validation #871

mbijon:
Thanks for looping me in @danielbachhuber. Your points about CPTs being at the core use-case for API driven sites is key to me. That doesn't mean the majority of all WP sites, but it's bound to be "most" sites with enough dev effort put in to use the API.

Then @maxcutler's point about two philosophies is notable. However, coming from a React & golang microservice world lately, I don't think those ends are distinct at all. A huge speed up on my recent project has come from the tools available in golang & React for generating and consuming Swagger docs.

We can generate swagger.json dynamically from our build process and both philosophies are sped-up: 

Ph1 is sped through generation of both human-readable docs for reference (or copy-paste if devs prefer that approach).

Ph2 is sped by having swagger.json on a consistently available path and pointing code generators at it. We could have maybe 60-80% of the models & handlers needed for an API consumer to work off Daniel's Fusion site in seconds. Plus, this level of discoverability opens up the possibility of Google or (ideally) new players to offer Internet-wide functionality searches, instead of just content searches.
(Sounds a bit like some RSS thing I hear helped democratize publishing)

Validation seems like something best done in custom ways. It should be possible, but I can't think of too many instances where a site can trust JS validation end-to-end.

Another benefit to doc generation not noted above that's valuable to maintaining a stable microservice platform -- Versioning. Versioned endpoints have been massively important, in my experience, to maintain the kind of flexibility Max's Ph1 aims for. This also address the easy vs. hard changes @rachelbaker summarized above.

Without versioned endpoints an addition of one param to a verb that's not idempotent forces us to to update + test 2-3 UX projects and multiple dependent services. With them the old endpoint keeps working and new ones add filters and params as needed. No forced-refactors of the dependent a services or apps (barring changes to the data requirements like required fields that force old contract rules to expire). 

Plus, machine-readable doc formats might be discovered in a CI build using something like Dredd. Then you can test if new endpoint versions will regress other apps.

Maybe the age of this ticket means more tooling needs to be built: A good PHP Swagger/API Blueprint/Slate parser and static doc sample would help get some of this discussion out of text and into code. But we could skip the larger amounts of implementation code until there's some boilerplate to reference.

PR #761, opened 2015-01-03
Do not require type parameter to be set when updating a Post

### LINKED ISSUE #760: Setting `type` should not be required when updating a Post
There is a bug in the logic of WP_JSON_Base_Posts_Controller::update_item where if the 'type' parameter is not set the following error is returned:

```
return new WP_Error( 'json_cannot_change_post_type', __( 'The post type may not be changed.' ), array( 'status' => 400 ) );
```

rachelbaker:
Fixes #760

Only check if a post_type is being changed if the type parameter is set.
Fallback to the post_type of the post being updated when doing permission checks.
Includes unit test.


rachelbaker:
@WP-API/amigos #reviewmerge

PR #768, opened 2015-01-07
Set endpoint properties from the schema

danielbachhuber:
See #759, #718

### DIFF
--- a/plugin.php
+++ b/plugin.php
@@ -87,6 +87,8 @@ function create_initial_json_routes() {
 		$controller = new WP_JSON_Posts_Controller;
 		$controller->post_type = $post_type;
 		$base = $post_type . 's';
+		$schema = $controller->get_item_schema();
+
 		register_json_route( 'wp', '/' . $base, array(
 			array(
 				'methods'         => WP_JSON_Server::READABLE,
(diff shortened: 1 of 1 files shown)


danielbachhuber:
Open question: should arguments only be required when creating, not updating?

rachelbaker:
> Open question: should arguments only be required when creating, not updating?

Great question.  I believe we are currently inconsistent in this regard. 

We cannot rely on the `PATCH` method being available on this distributed API of ours.  I believe (and correct me if I am wrong here @rmccue) the decision made around this for v1 to treat `PUT` requests like `PATCH` requests, and as though they are partial resource updates.  However, we still (specifically when updating Posts) required certain fields. This has resulted in pain points for folks trying to use the REST API in projects like a front-end editor or a new WP-Admin concept.  

How can we differentiate between the two very different resource update intentions?

* Full resource changes - required fields must be included, and optional fields not included should be removed.
* Partial resource changes - no required fields, only fields included will be modified.

rachelbaker:
Asked @justinshreve how the .com API handles this and his response was that they:

- don't require any fields on update.
- on create they only require title OR content.

We should probably match them.

rmccue:
> Open question: should arguments only be required when creating, not updating?

Arguments should only be required on creation; post data on update should be treated as extending the existing data (i.e. merging the new data into the old).


> We cannot rely on the `PATCH` method being available on this distributed API of ours. I believe (and correct me if I am wrong here @rmccue) the decision made around this for v1 to treat `PUT` requests like `PATCH` requests, and as though they are partial resource updates.

Actually, our use of `PATCH` is kind of an abuse of the verb; realistically, it should be closer to a [JSON Patch](http://jsonpatch.com/) format here. We support it with the same semantics as PUT because we don't need the complexity of that, but it wouldn't be regarded as a "real" `PATCH` request by many API/REST aficionados.

`PUT` is fine to allow partial updates; our data models require that fields exist, and partial updates are a practical implementation of reducing redundancy in request data.

danielbachhuber:
Added support for this in #844

PR #769, opened 2015-01-08
Register post types for API based on `access_in_json` argument

danielbachhuber:
`json_base` specifies the base slug used for the post type. Like core's
existing behavior, it will use `name` as the default.

`json_controller_class` probably needs a better name. It specifies which
controller class to use, because more complex post types will break
outside the standard `WP_JSON_Posts_Controller` class.

See #759, #718

### DIFF
--- a/plugin.php
+++ b/plugin.php
@@ -76,16 +76,40 @@ function register_json_route( $namespace, $route, $args = array(), $override = f
 }
 
 /**
+ * Add the extra Post Type registration arguments we need
+ * These attributes will eventually be committed to core.
+ */
+function _add_extra_api_post_type_arguments() {
+	global $wp_post_types;
+
+	$wp_post_types['post']->access_in_json = true;
+	$wp_post_types['post']->json_base = 'posts';
+	$wp_post_types['post']->json_controller_class = 'WP_JSON_Posts_Controller';
+
+	$wp_post_types['page']->access_in_json = true;
+	$wp_post_types['page']->json_base = 'pages';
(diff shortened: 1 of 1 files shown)


danielbachhuber:
@WP-API/amigos What do you think of this registration pattern?

danielbachhuber:
We should also automatically register revision and other endpoints based whether the post type supports those endpoints.

plugin.php
```diff
@@ -76,16 +76,40 @@ function register_json_route( $namespace, $route, $args = array(), $override = f
...
+	$wp_post_types['page']->access_in_json = true;
+	$wp_post_types['page']->json_base = 'pages';
+	$wp_post_types['page']->json_controller_class = 'WP_JSON_Posts_Controller';
+
+	// $wp_post_types['attachment']->access_in_json = true;
+	// $wp_post_types['attachment']->json_base = 'media';
+	// $wp_post_types['attachment']->json_controller_class = 'WP_JSON_Attachments_Controller';
+
+}
+add_action( 'init', '_add_extra_api_post_type_arguments', 11 );
```
rachelbaker: @danielbachhuber can the be hooked into `wp_loaded` instead?

danielbachhuber: Could be, why?

I added them just to get things working, with the assumption they'd eventually be committed to `create_initial_post_types()`

rachelbaker: Gotcha.  Makes sense.

danielbachhuber:
Continued in #820

PR #776, opened 2015-01-13
Update class-wp-json-media.php  Missing argument 2 error

tqeto:
Warning: Missing argument 2 for WP_JSON_Media::upload_attachment() in E:\xampp\www\wordpress\wp-content\plugins\WP-API\lib\class-wp-json-media.php on line 216

and 
print_r($_files);
here
$_files is a WP_JSON_Request Object 

so line 247 upload_from_file function
line 392 $_files['file'] got nothing!
https://github.com/WP-API/WP-API/blob/develop/lib/class-wp-json-media.php#L392

### DIFF
--- a/lib/class-wp-json-media.php
+++ b/lib/class-wp-json-media.php
@@ -213,7 +213,7 @@ class WP_JSON_Media extends WP_JSON_Posts {
 	 * @param array $_headers HTTP headers from the request
 	 * @return array|WP_Error Attachment data or error
 	 */
-	public function upload_attachment( $_files, $_headers, $data = null, $post_id = 0 ) {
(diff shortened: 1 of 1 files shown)


danielbachhuber:
@tqeto Can you branch and PR against `master` branch too?

tqeto:
@danielbachhuber 
here 
https://github.com/WP-API/WP-API/blob/develop/lib/class-wp-json-server.php#L613

added print_r(func_get_args()); function upload_attachment shows only one arg pass to upload_attachment

rmccue:
This is because the media endpoints haven't yet been updated for v2. We'll get to this one eventually, but right now, you'll need to use v1 (the `master` branch) if you want media.

Thanks for reporting this!

PR #786, opened 2015-01-18
Make json_pre_dispatch and json_post_dispatch consistent

joehoyle:
Currently `json_post_dispatch` passes the response data and the request
object, whereas the `json_pre_dispatch` passes the response data (null), the
server and the request object.

These filters should be consistent.


joehoyle:
@WP-API/amigos #reviewmerge

rachelbaker:
Merged #786 

:snowman:

Issue #542, opened 2014-10-29
CORS missing Access-Control-Expose-Headers for pagination


matogertel:
Need to add this in the CORS section, otherwise pagination doesn't work:
`header( 'Access-Control-Expose-Headers: X-WP-Total, X-WP-TotalPages');`

shprink:
+1
It actually took me quite some time to discover that the AngularJs Client was not the problem ;) https://www.npmjs.com/package/wp-api-angularjs

romuloctba:
+1 CORS and mosquitos are the worst pain I face daily

2015-01-23 17:39 GMT-02:00 Julien Renaux <notifications@github.com>:

shprink:
Pull request pending: https://github.com/WP-API/WP-API/pull/802

asimpson:
Any reason why this isn't merged? I just spent hours trying to figure out why I wasn't getting the proper headers in the browser versus on the server.

Chaser324:
+1. Merge this.

aaronsmulktis:
+1 need this

danielbachhuber:
These headers are included in core infrastructure — no longer a bug.

PR #802, opened 2015-01-23
Adding Access-Control-Expose-Headers header to allow pagination through ...

shprink:
...CORS

https://github.com/WP-API/WP-API/issues/542

I have tested, it works great


joehoyle:
@rmccue is probably the best one to provide feedback here

shprink:
If you merge do not forget to backport to v2 :)

joehoyle:
@shprink this is against the `develop` branch, which is v2

shprink:
Did not know develop was v2 :) thanks @joehoyle

joehoyle:
From Slack, it's thought if the spec just allows access to all headers, that would be sane, as we don't have anything we want to hide.

rmccue:
Closing, as it's now essentially a duplicate of #1529

PR #805, opened 2015-01-24
Correct /posts/ endpoint read post permission logic

### LINKED ISSUE #788: Flawed logic when checking post read permission
In this loop, we're not unsetting `$post` if the user doesn't have read permission:

```
foreach ( $posts as &$post ) {
	if ( ! $this->check_read_permission( $post ) ) {
		continue;
	}

	$post = $this->prepare_item_for_response( $post, $request );
}
```

rachelbaker:
Fixes #788 

No longer passing $post by reference in `get_items()` method of Posts Controller, instead using an array to gather posts once they pass the permission check.

Added unit test to confirm logic is no longer flawed.  

This was not an issue in any of the other endpoint Controllers.


rachelbaker:
@WP-API/amigos #reviewmerge

PR #820, opened 2015-01-30
Posts controller abstraction

danielbachhuber:
This pull request pulls the non-controversial changes out of #759 so we can continue moving forward.

Specifically, I have:

* Collapsed `WP_JSON_Base_Posts_Controller` into a `WP_JSON_Posts_Controller` that can be used with multiple post types.
* Automatically registered endpoints for post types registered with the `show_in_json` argument. The `WP_JSON_Posts_Controller` is used by default for these post types, but they would be able to register their own controller class.
* Made resource fields reflect `post_type_supports()` and other post type attributes (e.g. a post type has to support `title` for it to be present in the response). Extra data included in `POST` and `PUT` requests is silently discarded.

This pull request does not:

* Change the name of any fields.
* Introduce a JSON Schema for each resource.

Once this is merged, I'll put together pull requests for each of the two aforementioned points.

See #718


danielbachhuber:
@WP-API/amigos #reviewmerge

joehoyle:
@danielbachhuber if I had a _very_ custom post type, what would be the best way to subclass `WP_JSON_Posts_Controller` and use that subclass in the routes? Would I need to roll all my own `register_json_route` calls? Doh, just seen `json_controller_class` I guess that's for exactly this purpose. I was actually going to suggest this!

`json_base` arg name, would `json_route` be a little more descriptive? Don't want to bike-shed so not too bothered though!

+1 for merging

danielbachhuber:
> Doh, just seen json_controller_class I guess that's for exactly this purpose. I was actually going to suggest this!

Yep. Probably should also make sure that the supplied class name is a subclass of `WP_JSON_Controller`.

> json_base arg name, would json_route be a little more descriptive? Don't want to bike-shed so not too bothered though!

Open to suggestions, or we can file another issue and come back to it.

Issue #572, opened 2014-11-03
Add field include/exclude parameters to change which fields are returned


rmccue:
Commonly requested feature.

We're going to do this in a plugin, as suggested by @danielbachhuber.

dsebastien:
Can you explain a bit the idea/rationale for implementing that as part of a plugin? Would that mean defining a separate API endpoint or extending the current one with more 'filtering' capabilities?

joehoyle:
@dsebastien I think basically we didn't think this fit the 80/20 split. The thought was a simple add-on plugin could make the necessary hooks in to modify the existing endpoints to remove any fields that were not requested for any request.

elyobo:
I've added a simple plugin which allows very flexible field selection (among other things) at https://github.com/elyobo/wp-rest-jmespath; it's very simple, just listening in before the response is sent and using [JMESPath](http://jmespath.org) to do all the heavy lifting.  JMESPath allows doing lot more than just field selection and seems like a pretty good fit for this problem to me.

kadamwhite:
See also https://wordpress.org/plugins/wp-rest-api-filter-fields/ (from comment in #446); may be stale, hasn't been updated in a year.

kadamwhite:
I'm moving this discussion to Trac, since this is a concern that would need to be tackled on the core infrastructure side. I believe this is important functionality to include, and agree that it should be tried out in plugins before we consider merging it into core.

Trac ticket: https://core.trac.wordpress.org/ticket/38131#ticket

elyobo:
See also the simple jmespath plugin I added for this; a steeper learning curve (you need a little JMESPath knowledge - http://jmespath.org), but more flexible in what it allows and will work with any type of
content, including completely custom APIs.

https://wordpress.org/plugins/wp-rest-jmespath/

PR #835, opened 2015-02-04
Allow Pretty Printing of JSON Responses on PHP 5.4+

chrisguitarguy:
This changes two big things

1. Moves all calls to `json_encode` into a wrapper method so flags and such can be set in one place
2. Looks for the `pretty` query string param so you can get human-readable JSON with PHP 5.4+

Even if the second one isn't okay, I do think the first item is worth a look. I stole this idea from Elasticsearch, which sends non-pretty JSON by default and has all endpoints supporting a `pretty` query string argument.

Normal:

screen shot 2015-02-04 at 12 50 55 pm

Pretty Printed:

screen shot 2015-02-04 at 12 51 07 pm

### DIFF
--- a/lib/class-wp-json-server.php
+++ b/lib/class-wp-json-server.php
@@ -151,7 +151,7 @@ class WP_JSON_Server {
 		}
 		$error = compact( 'code', 'message' );
 
-		return json_encode( array( $error ) );
+		return $this->json_encode( array( $error ) );
 	}
 
 	/**
@@ -282,13 +282,13 @@ class WP_JSON_Server {
 			// Embed links inside the request
 			$result = $this->response_to_data( $result, isset( $_GET['_embed'] ) );
 
-			$result = json_encode( $result );
+			$result = $this->json_encode( $this->prepare_response( $result ) );
 
 			$json_error_message = $this->get_json_last_error();
 			if ( $json_error_message ) {
 				$json_error_obj = new WP_Error( 'json_encode_error', $json_error_message, array( 'status' => 500 ) );
 				$result = $this->error_to_response( $json_error_obj );
-				$result = json_encode( $result->data[0] );
+				$result = $this->json_encode( $result->data[0] );
 			}
 
 			if ( isset( $_GET['_jsonp'] ) ) {
@@ -839,5 +839,16 @@ class WP_JSON_Server {
 		}
 
 		return $headers;
+	}
+
+	/**
+	 * JSON encode a value. This is a wrapper around `json_encode` so argument
+	 * flags and such can be set in a single place.
+	 *
+	 * @param array|object $data The value to encode
+	 * @return string The json encoded value
+	 */
+	protected function json_encode( $data ) {
+		return json_encode( $data );
 	}
 }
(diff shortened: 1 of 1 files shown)


JustinSainton:
:+1:, though won't `JSON_PRETTY_PRINT` [always be defined](https://github.com/WordPress/WordPress/commit/45afcde6408e795ee1efa28e6b9597014126058f) as of 4.1?

chrisguitarguy:
Good call. Not sure if the `defined` check should be removed or not -- depends on what WP versions `develop` is targeting.

danielbachhuber:
> depends on what WP versions develop is targeting.

We still need to support PHP 5.2, unfortunately. I don't think this makes sense for core, but it would be a nice supplementary plugin, similar to #572

If you want to put it in a `wp-api-pretty-print` repo, we'd be happy to transfer it into the WP-API organization for visibility.

chrisguitarguy:
This does support PHP 5.2, it just does nothing with 5.2. If the pretty print thing isn't okay daee5cd83c92b18568fdf32c8f3f8be83d977d70 and a2bd1bda9e0d74ba67ab43da1e264db6a65370cc could still be used to open the door for extensions to how JSON is rendered.

rmccue:
> This does support PHP 5.2, it just does nothing with 5.2.

The problem with this is that it's inconsistent across sites, which means we can't mention it in documentation/etc. The typical use for this is debugging, which is typically on a singular site, however there are other ways to improve the visual representation of JSON.

Thanks for the pull request, but I don't think we can merge it right now. Adding features that are conditional on versions of PHP sucks for end-users. If we can shim this into PHP 5.2 (and for the JSON compat lib included with WP), I'd be happy to reconsider, but I get the feeling it might end up with a significant amount of code. (I'd love to be proved wrong!)

If you want better looking JSON in a browser, I recommend [JSONView](http://jsonview.com/) for Firefox and the [unofficial port](https://chrome.google.com/webstore/detail/jsonview/chklaanhfefbnpoihckbnefhakgolnmc) for Chrome. For command-line requests, I recommend [jq](http://stedolan.github.io/jq/) (as a standalone tool) or [httpie](https://github.com/jakubroztocil/httpie) (as a replacement for cURL, also a fantastic way to send HTTP requests).

johnbillion:
Also any half decent HTTP client or HTTP proxy will allow you to view formatted JSON. There's no reason to be interacting with a JSON API via a browser for development purposes.

Some examples:

 * [Paw](https://luckymarmot.com/paw) or [RESTClient](https://github.com/wiztools/rest-client) for OS X
 * [Postman](https://chrome.google.com/webstore/detail/postman-rest-client/fdmmgilgnpjigdojojpjoooidkmcomcm) or [DHC](https://chrome.google.com/webstore/detail/dhc-resthttp-api-client/aejoelaoggembcahagimdiliamlcdmfm) extensions for Chrome
 * [Charles Proxy](https://github.com/wiztools/rest-client) for OS X / Windows / Linux

AstonBraham:
Please Is there a way to remove the "rendered : " in Content field ?
I need to have the content as a String only. Help please!

PR #836, opened 2015-02-04
Posts Controller Headers and Links Fixes

rachelbaker:


rachelbaker:
Fixes #738

rachelbaker:
@WP-API/amigos #reviewmerge

lib/class-wp-json-posts-controller.php
```diff
@@ -748,6 +747,22 @@ protected function check_delete_permission( $post ) {
...
 	/**
+	 * Get the base path for a post type's endpoints.
+	 *
+	 * @param object|string $post_type
+	 * @return string       $base
+	 */
+	protected function get_post_type_base( $post_type ) {
+		if ( ! is_object( $post_type ) ) {
+			$post_type = get_post_type_object( $post_type );
+		}
```
danielbachhuber: Do we really need this?

lib/class-wp-json-posts-controller.php
```diff
@@ -748,6 +747,22 @@ protected function check_delete_permission( $post ) {
...
+
+		$base = ! empty( $post_type->json_base ) ? $post_type->json_base : $post_type->name;
```
danielbachhuber: Crafty! We should probably have `register_post_type()` fill in `$post_type->json_base` so we don't need this conditional.

lib/class-wp-json-posts-controller.php
danielbachhuber: Eh, I'm fine with it.

PR #841, opened 2015-02-05
Don't noop `future` status. It's confusing

### LINKED ISSUE #723: "future" status is converted to "publish" when getting posts
When using the Basic Auth plugin and making a request as an authenticated user in the `edit` context, posts with `future` status are always being converted to `publish` status. An example cURL request looks like:

```bash
curl --user un:pw mysite.com/wp-json/posts/12345/?context=edit
```

Looking through the source, I found [some code](https://github.com/WP-API/WP-API/blob/87b0f7b08c54a0b8c730099511b3e43a9697266d/lib/class-wp-json-posts.php#L757-L759) that explicitly changes `future` to `publish`, with the comment:

> Consider future posts as published

I assume that there was a reason that this was done, but it's a bummer that we cannot actually get posts with their correct `future` status. Can anyone clarify this? I'd like to remove those lines of code, but I do not know the implications of the change.

danielbachhuber:
Fixes #723

### DIFF
--- a/lib/class-wp-json-posts-controller.php
+++ b/lib/class-wp-json-posts-controller.php
@@ -785,10 +785,6 @@ class WP_JSON_Posts_Controller extends WP_JSON_Controller {
 
 			$data = array_merge_recursive( $data, $data_raw );
 
-			// Consider future posts as published
-			if ( 'future' == $data['status'] ) {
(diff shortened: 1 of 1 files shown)


danielbachhuber:
@WP-API/amigos #reviewmerge

PR #844, opened 2015-02-06
JSON Schemas for our Controllers, second attempt

danielbachhuber:
Work in-progress

See #718, conversation on #759


danielbachhuber:
@WP-API/amigos #reviewmerge

danielbachhuber:
@joehoyle @rachelbaker Can I get your review and thoughts? This needs to go in sooner rather than later so we can work from it.

lib/class-wp-json-posts-controller.php
```diff
@@ -933,4 +940,206 @@ protected function prepare_links( $post ) {
...
+					'description'  => 'Type of Post for the object.',
+					'type'         => 'string',
+				),
+				'slug'             => array(
+					'description'  => 'An alphanumeric identifier for the object unique to its type.',
+					'type'         => 'string',
+				),
+				'link'             => array(
+					'description'  => 'URL to the object.',
+					'type'         => 'string',
```
rachelbaker: @danielbachhuber Should the string format be included?  'format' => 'uri'

danielbachhuber: Yes, we can be more specific if we'd like, or we can do another pass later.

lib/class-wp-json-posts-controller.php
```diff
@@ -933,4 +940,206 @@ protected function prepare_links( $post ) {
...
+					break;
+
+			}
+
+		}
+
+		if ( 'post' === $this->post_type ) {
+			$schema['properties']['sticky'] = array(
+				'description'      => 'Whether or not the object should be treated as sticky.',
+				'type'             => 'bool',
```
rachelbaker: `bool` should be expanded to full type name of `boolean`

lib/class-wp-json-posts-controller.php
```diff
@@ -933,4 +940,206 @@ protected function prepare_links( $post ) {
...
+								'description'     => 'Excerpt for the object, transformed for display.',
+								'type'            => 'string',
+								),
+							),
+						);
+					break;
+
+				case 'thumbnail':
+					$schema['properties']['featured_image'] = array(
+						'description'     => 'A featured image for the object.',
```
rachelbaker: This description is not very helpful.  Should be something like "ID of the featured image for the object."

lib/class-wp-json-users-controller.php
```diff
@@ -331,4 +331,35 @@ protected function prepare_item_for_database( $request ) {
...
+			'properties'           => array(
+				'id'               => array(
+					'description'  => 'Unique identifier for the object.',
+					'type'         => 'integer',
+					),
+				'name'             => array(
+					'description'  => 'Display name for the object.',
+					'type'         => 'string',
+					),
+				'email'            => array(
```
rachelbaker: How would we account for returning `false` for `email` in the `view` context?

lib/class-wp-json-users-controller.php
```diff
@@ -331,4 +331,35 @@ protected function prepare_item_for_database( $request ) {
 
 		return apply_filters( 'json_pre_insert_user', $prepared_user, $request );
 	}
+
+	/**
+	 * Get the Post's schema, conforming to JSON Schema
```
rachelbaker: @danielbachhuber Should be changed to "Get the User's schema..."

lib/class-wp-json-users-controller.php
danielbachhuber: Variable data based on context is something we still need to discuss.

rachelbaker:
@danielbachhuber @joehoyle @rmccue I am still not sold on the idea of `fixed` schemas, but willing to acquiesce to a majority opinion.  I see the rigidness as a drawback, and am concerned we wouldn't be solving a real problem.  Just my opinion.

Looking through this PR brought the following questions to mind: 

1. How do we handle expressing the additional fields returned in non-default contexts?  
2. Are there only three User object schema properties defined because that schema is incomplete?

danielbachhuber:
>  I am still not sold on the idea of fixed schemas, but willing to acquiesce to a majority opinion.

I was under the assumption that we had all agreed on fixed schemas?

> Are there only three User object schema properties defined because that schema is incomplete?

Oh, I didn't finish. I can finish User and Comment schemas before merging if you'd like

danielbachhuber:
I'm going to finish up schemas for all Resources, and then this will be ready for a final reviewmerge

lib/class-wp-json-posts-controller.php
danielbachhuber: 159d939

lib/class-wp-json-posts-controller.php
danielbachhuber: 9a6741f

lib/class-wp-json-posts-controller.php
danielbachhuber: 034f332

lib/class-wp-json-users-controller.php
danielbachhuber: 2def040

lib/class-wp-json-users-controller.php
danielbachhuber: #894

PR #845, opened 2015-02-06
Make `get_post_type_base()` public so we can DRY

danielbachhuber:


danielbachhuber:
@rachelbaker how do you feel about this? should we make `get_json_base()` a method of the base controller class? Also, we probably don't need to pass the argument...

rachelbaker:
@danielbachhuber Looks great to me.

danielbachhuber:
What do you think about having this for other controllers too?

PR #853, opened 2015-02-07
Remove Duplicate Logic for Post Type Attributes

rachelbaker:
Removes legacy `json_register_post_type()` that was modifying post_type attributes in conflict with #820.  The `json_register_post_type()` function was causing additional revision routes to be registered that are not supported.  

This PR also includes some minor code clean up.


rachelbaker:
@WP-API/amigos #reviewmerge

joehoyle:
Makes sense

PR #854, opened 2015-02-08
Separate permissions logic for comments

joehoyle:
As we can now specify the permissions logic separatly to the callbacks themselves, I tried this for comments. This has the benefit of more modular code, but also the client can now send HEAD requests to the resource to find out what method they are allowed to perform on the resource.

I removed the specific errors for each permission, but I can put them back in if we require. I also updated the tests to expect a 403 rather than a 401, as that's what we decided for failed cap
 checks I believe (feel free to correct me here, I was going off what I interpretted from http://stackoverflow.com/questions/3297048/403-forbidden-vs-401-unauthorized-http-responses)


lib/class-wp-json-comments-controller.php
```diff
@@ -191,6 +167,115 @@ public function delete_item( $request ) {
...
+
+	/**
+	 * Check if a given request has access to read comments
+	 * 
+	 * @param  WP_JSON_Request $request Full details about the request.
+	 * @return bool
+	 */
+	public function get_items_permissions_check( $request ) {
+
+		// If the post id isn't specified, presume we can create
```
rachelbaker: We should presume we can view or get, not create.

joehoyle: @rachelbaker I'm not sure the best way to handle this, as `post_id` is a required param, this will never return `true` however, this is also used for the `OPTIONS` pre flight check, so if someone does `OPTIONS /wp/comments` do we say they have permission to `POST` there or not? (as opposed to `OPTIONS /wp/comments?post_id=123`

rachelbaker: @joehoyle I think I understand where you are coming from.  `$post_id` is a required parameter for creating a comment, but not for returning a collection of comments.  
The issue is what to do in the following use-case:

Jane wants to sent a POST request to create a comment.
Jane first sends an OPTIONS request to `wp/comments` 
Jane expects to receive a response to her request that includes if she CAN create a comment.

@joehoyle Is that correct?

joehoyle: @rachelbaker doh you are totally right, I misread where you had put this comment. Yes, the use case you mentioned is correct, though the user should probably send a OPTIONS to `wp/comments/?p=123` to check for that specific comment. In which case we should probably return `false` for `POST wp/comments` as I presume someone can't POST there.

rachelbaker:
Merged #854

Issue #864, opened 2015-02-09
Further abstraction of controller test structure


danielbachhuber:
https://github.com/WP-API/WP-API/blob/develop/tests/test-json-posts-controller.php is getting pretty meaty — ~1k lines. We should abstract it further so it's easier to read.

How about breaking the controller test pattern into separate classes for each type of operation (CRUD)?

rmccue:
> How about breaking the controller test pattern into separate classes for each type of operation (CRUD)?

My only concern with this is making sure we avoid complexity to the tests. I intentionally try and keep tests as free of complexity (including avoiding branches) since it can make bugs super hard to find.

If we can just do a split, :+1: :)

danielbachhuber:
Abstraction is going pretty well now. There are a couple good follow-up issues.

PR #870, opened 2015-02-11
Break Pages tests into a separate class

danielbachhuber:
As we continue expanding our tests, abstraction will make our tests
easier to review.

See #864


danielbachhuber:
Erps, @WP-API/amigos #reviewmerge

rachelbaker:
merged :ok:

Issue #717, opened 2014-12-10
Split core infrastructure into separate directory


rmccue:
From #571. This should make it easier to review the infrastructure, and is our first step towards actually getting this stuff reviewed.

rmccue:
Basically, we want to take the API and split it into two sections: the infrastructure, and the endpoints. The infrastructure basically encompasses the server/router, request/response objects, and the main `plugin.php` file. We want to take those and move them into something like `lib/infrastructure` and the rest into `lib/endpoints`.

Essentially, it's everything you need to create an API, but not the user facing API itself.

danielbachhuber:
@rmccue I've broken out `/lib/infrastructure` and `/lib/endpoints`. Not clear what we want to do with `plugin.php` or `lib/wp-json.php` right now — can you specify?

PR #872, opened 2015-02-11
Move infrastructure classes to `lib/infrastructure`, part one

danielbachhuber:
See #717


danielbachhuber:
@WP-API/amigos #reviewmerge

PR #874, opened 2015-02-11
Add documentation on why meta is unavailable

### LINKED ISSUE #857: Post Meta / Authentication - Update Docs / Enable certain meta fields to be read all?
Is there a way to define which post meta should / not be secured behind authentication?

Or is there additional doc / examples other than http://wp-api.org/guides/authentication.html regarding the use of cookie authentication?

Using v0.9 of the API I had been able to retrieve meta (intended as viewable by all) through the following in a template:
```` 
$('.capDisplay').on('click', function(e){		
	e.preventDefault();
        var capName = $(this).text();
        var capID = $(this).attr('data-id');
        var l = Ladda.create(this);
	l.start();
			
	restURL = '/wp-json/posts/' + capID;		
	$.getJSON(restURL, function(response) {

		var dataTypes = ['description', 'future_direction', 'applications', 'contacts'];		
		$.each( dataTypes, function( key, dataType ) {
			//Parse post_meta and update UI accordingly
		});
	}).always(function() { l.stop(); });

});
````

The http://wp-api.org/guides/authentication.html documentation suggests that for plugins / themes using cookie authentication is suggested format, but is a little sparse on details to be able to create equivalent as above for /wp-json/posts/ [ID] /meta

In this scenario it is not a valid expectation that a site visitor would be both registered, signed-in and have associated edit capabilities tied to their role. I'm hoping that just as get_post_meta enables presentation via theme for meta that there is a way to specific which meta fields should be accessible via the API in non-authenticated scenarios?

Thanks!

rmccue:
We get constant questions about why meta isn't accessible. Let's improve the documentation here a bit so it's easier to understand, and we can avoid having issues constantly filed.

Fixes #857

### DIFF
--- a/docs/routes/routes.md
+++ b/docs/routes/routes.md
@@ -5,7 +5,7 @@ Create a Post
 -------------
 
 	POST /posts
-  
+
   Requires [authentication](http://wp-api.org/guides/authentication.html)
 
 ### Input
@@ -24,7 +24,7 @@ data[title]=Hello%20World!&data[content_raw]=Content&data[excerpt_raw]=Excerpt
 
 ```
 Content-Type: application/json
-	
+
 {"title":"Hello World!","content_raw":"Content","excerpt_raw":"Excerpt"}
  ```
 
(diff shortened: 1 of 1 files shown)


rmccue:
@pollyplummer Can you give this a review/copy pass please?

PR #883, opened 2015-02-14
Passing a value for the slug parameter should update the post_name.

### LINKED ISSUE #856: Updating `post_name` expects `name`, not `slug`
See https://github.com/WP-API/WP-API/blob/develop/lib/class-wp-json-posts-controller.php#L455

We should make sure this has tests too, so it doesn't happen again.

rachelbaker:
Fixes #856.

Correct request parameter conditional logic to look for the 'slug' parameter to update a Posts post_name property.
Includes unit test.


rachelbaker:
@WP-API/amigos #reviewmerge

danielbachhuber:
:+1: top notch

PR #885, opened 2015-02-15
Empty checks in Posts Controller make setting values to Falsy impossible

joehoyle:
Currently we are relying on a lot of `! empty( $var )` checks which fail on setting `0` or `""` for valid values. I fixed up `menu_order`, `contnet`, `password` and `excerpt`


joehoyle:
@wp-api/amigos #reviewmerge

PR #888, opened 2015-02-16
Return 200 and an empty array for valid queries with 0 results.

### LINKED ISSUE #862: Post query returns json_invalid_query for valid query
If you do a query that happens to have no results (e.g. `cat=2` where the category has no posts), you get back a 404 and a `json_invalid_query` error. This should just be an empty JSON list, and not an error.

rachelbaker:
Fixes #862. 

A post_type collection endpoint will no longer return a 404 status and a `json_invalid_query` WP_Error for valid queries that return 0 results.  

Now, valid queries that return 0 results will return a 200 status and an empty array.


rachelbaker:
@WP-API/amigos #reviewmerge

danielbachhuber:
:apple:

PR #890, opened 2015-02-18
Don't bail early if there are no found posts

danielbachhuber:
If we're at the end of a range, we'd still want our query navigation
headers

### DIFF
--- a/lib/class-wp-json-posts-controller.php
+++ b/lib/class-wp-json-posts-controller.php
@@ -26,9 +26,6 @@ class WP_JSON_Posts_Controller extends WP_JSON_Controller {
 
 		$posts_query = new WP_Query();
 		$query_result = $posts_query->query( $query_args );
-		if ( 0 === $posts_query->found_posts ) {
-			return json_ensure_response( array() );
-		}
 
 		$posts = array();
 		foreach ( $query_result as $post ) {


danielbachhuber:
@WP-API/amigos #reviewmerge

rachelbaker:
Related #888 and #862 

@rmccue Thoughts?

rmccue:
Hmm, looks OK to me. With this change, we should still see `[]` back from an endpoint, but now get the query nav headers too, which is nice. What happens if you're way over the page limit though? (`page=999` with only 2 posts in the collection)

danielbachhuber:
> What happens if you're way over the page limit though?

You'd still get query nav headers. I don't think that's unexpected

rmccue:
> You'd still get query nav headers. I don't think that's unexpected

Right, but they'd link to `?page=998`, which isn't where we should link IMO.

Here's what I'm thinking. We keep prev/next the same when you're on a page that "exists" (that is, the page is between 1 and `max_num_pages`). On all pages, link to `rel=first` and `rel=last`. This would ensure we never link to pages that don't exist, while still allowing navigation.

danielbachhuber:
Seems reasonable. I'll update when I have a moment.

danielbachhuber:
Will be handled in #1209

PR #892, opened 2015-02-18
Mark some schema fields as `readonly`

danielbachhuber:
Fully implementing this will require us extending JSON Schema and
publishing our additions

Depends on #844


danielbachhuber:
Going to re-open this later when the idea is more fleshed out.

PR #904, opened 2015-02-22
Convert /media to new controller pattern

danielbachhuber:
Related #605


lib/endpoints/class-wp-json-attachments-controller.php
```diff
@@ -0,0 +1,338 @@
...
+				$title = $image_meta['title'];
+			}
+
+			if ( empty( $request['caption'] ) && trim( $image_meta['caption'] ) ) {
+				$caption = $image_meta['caption'];
+			}
+		}
+
+		$attachment = $this->prepare_item_for_database( $request );
+		$attachment->post_type = $this->post_type; // Currently defaults to 'post'
```
danielbachhuber: This hack is because of #913

danielbachhuber:
@WP-API/amigos #reviewmerge

danielbachhuber:
Oops, I've fixed the build.

lib/endpoints/class-wp-json-attachments-controller.php
```diff
@@ -0,0 +1,337 @@
...
+	 * Create a single attachment
+	 *
+	 * @param WP_JSON_Request $request Full details about the request
+	 * @return WP_Error|WP_HTTP_ResponseInterface
+	 */
+	public function create_item( $request ) {
+
+		// Permissions check - Note: "upload_files" cap is returned for an attachment by $post_type_obj->cap->create_posts
+		$post_type_obj = get_post_type_object( $this->post_type );
+		if ( ! current_user_can( $post_type_obj->cap->create_posts ) || ! current_user_can( $post_type_obj->cap->edit_posts ) ) {
```
joehoyle: I know we didn't merge #854 yet, but the `permissions_callback` already exists. Maybe we just do that on a second pass though?

lib/endpoints/class-wp-json-attachments-controller.php
```diff
@@ -0,0 +1,337 @@
...
+			if ( empty( $request['caption'] ) && trim( $image_meta['caption'] ) ) {
+				$caption = $image_meta['caption'];
+			}
+		}
+
+		$attachment = $this->prepare_item_for_database( $request );
+		$attachment->file = $file;
+		$attachment->post_mime_type = $type;
+		$attachment->guid = $url;
+		$id = wp_insert_post( $attachment, true );
```
joehoyle: To be explicit, would wp_insert_attachment be better? I doesn't do much right now, but seems more correct, especially if hooks are added in the future.

danielbachhuber: I used `wp_insert_post()` because there's no functional reason to use `wp_insert_attachment()` and `wp_insert_post()` supports returning `WP_Error` when there's a failure (and `wp_insert_attachment()` does not).

lib/endpoints/class-wp-json-attachments-controller.php
danielbachhuber: #922

joehoyle:
Looks good to me r+

PR #913, opened 2015-02-22
Always defer to controller for post type

### LINKED ISSUE #884: Why do we allow changing the post type of a post
Given that we are potentially exposing different schema for different post types, it has some weird behaviour if you switch a post between post types. It seems we should not allow this, also we should throw a 404 for `/wp/posts/123` if 123 is a page, which we currently don't I think.

danielbachhuber:
If `type` is present in the request, safely ignore it

Fixes #884


danielbachhuber:
@WP-API/amigos #reviewmerge

rachelbaker:
merged #913

Issue #704, opened 2014-12-09
Remove old-style WP_JSON_Posts and related tests


rachelbaker:
Waiting on #684

PR #914, opened 2015-02-22
Expose post type data at `/types`

danielbachhuber:
See #704

### DIFF
--- a/lib/endpoints/class-wp-json-post-types-controller.php
+++ b/lib/endpoints/class-wp-json-post-types-controller.php
@@ -47,13 +47,48 @@ class WP_JSON_Post_Types_Controller extends WP_JSON_Controller {
 		}
 
 		return array(
+			'description'  => $post_type->description,
+			'hierarchical' => $post_type->hierarchical,
+			'labels'       => $post_type->labels,
 			'name'         => $post_type->label,
 			'slug'         => $post_type->name,
-			'description'  => $post_type->description,
-			'labels'       => $post_type->labels,
-			'hierarchical' => $post_type->hierarchical,
 		);
 	}
 
+	/**
+	 * Get the Post type's schema, conforming to JSON Schema
+	 *
+	 * @return array
+	 */
+	public function get_item_schema() {
+		$schema = array(
+			'$schema'              => 'http://json-schema.org/draft-04/schema#',
+			'title'                => 'type',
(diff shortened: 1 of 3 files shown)


rmccue:
As noted in #524, the format of data here changed from an array to a hash. Is there a reason this was done? Post types have a unique non-numeric ID, so it's better from a usability perspective to have a hash keyed by that.

(Compare looking up labels in JS: `data[ post_type ].labels` vs `_.findWhere( data, { slug: post_type } ).labels`, and that's assuming you have Underscore or similar. O(1) vs O(N) behaviour)

danielbachhuber:
> Post types have a unique non-numeric ID, so it's better from a usability perspective to have a hash keyed by that.

This seems reasonable to me. I wasn't aware of it as a design decision. I think the usability outweighs the inconsistency with other collections.

Would the key be the json_base for the post type? Should tax and post status endpoints follow suit?

I haven't explored the idea in depth, but I also started thinking that this informational data should be placed under a `/site` endpoint

danielbachhuber:
Feel free to create a new issue where we can discuss at depth.

wkw:
I think this was my pull request, so I'll weigh in on how I came upon it (as best I remember now a few months later). I was putting the Backbone wp-api classes through some paces (https://github.com/WP-API/client-js) and this endpoint broke them. Backbone expects arrays back from data APIs. To receive otherwise requires massaging when the data is received.

This endpoint was the only one that behaved this way, and logical as it may be, it seems inconsistent in comparison to all the others.

rmccue:
> Backbone expects arrays back from data APIs. To receive otherwise requires massaging when the data is received.

One key difference with this endpoint as compared with the others is that this isn't really *data*, it's server configuration.

That said, the Backbone client certainly shouldn't fail on that. Looping in @tlovett1 to double-check that one :)

tlovett1:
Yep this is a problem. The Backbone client PostType collection is not fetching properly because the response is an object and not an array. We can either change the response or hack the client a bit. I think the response should be changed.

joehoyle:
I think we should change it back to be an array - there are other areas for convenience where a hash might be easier but we don't do that, for example we have the same issue in `_embedded` for a post's terms, you have to do `_.findWhere( post._embedded['wp:term'], { taxonomy: 'category' } )`

PR #917, opened 2015-02-23
Use `wp_filter_post_kses()` instead of `wp_kses_post()` on insert

danielbachhuber:
The latter is intended for display, whereas the former is good for
sanitizing data


danielbachhuber:
@WP-API/amigos #reviewmerge

PR #918, opened 2015-02-23
If value is ID reference to object, append `_id`

danielbachhuber:
This more clearly denotes the special nature of the field.

Affects `featured_image_id`, `author_id` and `parent_id` on Post
controller.

See #605


danielbachhuber:
@WP-API/amigos #reviewmerge

rmccue:
FWIW, I'm -0 on this one. We need to be careful to not be overly verbose; if `author` is an integer, it should be pretty obvious that it refers to the author's ID without needing `_id`. (We don't want to use Hungarian notation for fields, but some fields might be better if they're clarified.)

Are there any cases where it's *not* obvious as to the field's usage?

danielbachhuber:
Neither Github nor Stripe follow this pattern, so I'm fine with dropping `_id` altogether #941

PR #927, opened 2015-02-25
WIP for Additional Fields in responses

joehoyle:


lib/endpoints/class-wp-json-controller.php
```diff
@@ -169,7 +169,120 @@ public function filter_response_by_context( $data, $context ) {
...
+			if ( ! $field_options['update_callback'] ) {
+				continue;
+			}
+
+			// Don't run the update callbacks if the data wasn't passed in the request
+			if ( ! isset( $request[$field_name] ) ) {
+				continue;
+			}
+
+			$result = call_user_func( $field_options['update_callback'], $request[$field_name], $object, $field_name, $request );
```
rmccue: We don't actually do anything with this result. If it's a `WP_Error`, we should return it upwards and break the loop immediately.

lib/endpoints/class-wp-json-users-controller.php
```diff
@@ -208,6 +208,9 @@ public function create_item( $request ) {
 			}
 			$user->ID = $user_id;
 		}
+
+		$this->update_additional_fields_for_object( $user, $request );
```
rmccue: Make sure we add checks here when we add `WP_Error` returning

joehoyle: I don't see why we'd listen for a WP_Error, as the update to the "main" object has been done - we wouldn't want to return an error anyway presumably?

rmccue: If an `update_option` or `update_post_meta` fails (DB is too busy, perhaps), it's important that we pass this data back up the entire stack. I'm fine with us treating each field as atomic by itself though, rather than the whole set of fields as atomic (i.e. one failure doesn't break immediately).

lib/endpoints/class-wp-json-controller.php
joehoyle: This seems to contradict your comment below:

>  I'm fine with us treating each field as atomic by itself though, rather than the whole set of fields as atomic (i.e. one failure doesn't break immediately).

I think the fields should be treat in insolation, if a given field fails to update, we shouldn't discriminate against other addition fields by other plugins etc.

lib/endpoints/class-wp-json-users-controller.php
joehoyle: Does this mean we'd return a 500 or something on an update of a post? If not, how if this error returned?

lib/endpoints/class-wp-rest-posts-controller.php
```diff
@@ -1150,7 +1156,7 @@ public function get_item_schema() {
 		$base = $this->get_post_type_base( $this->post_type );
 		$schema = array(
 			'$schema'    => 'http://json-schema.org/draft-04/schema#',
-			'title'      => $base,
+			'title'      => $this->post_type,
```
joehoyle: @danielbachhuber this was using `posts` before for the "title", however all our other endpoints use singular names. I presumed this `title` refers to the object name, not plural form, so i changed this to just the post type. Do you see issues with that?

danielbachhuber: Should be fine for now. We'll eventually need to revisit these schemas to dot our i's and cross our t's

rachelbaker:
Merged #927

Issue #444, opened 2014-08-19
Error on Post Update of featured_image


domenic95:
Fatal error:  Cannot use object of type WP_JSON_Response as array in /var/www/wordpress/wp-content/plugins/WP-API/lib/class-wp-json-media.php on line 437

danielbachhuber:
This was fixed in #1024

PR #930, opened 2015-02-26
Fix error on Post Update of featured_image (WP-API/WP-API#444)

rhurling:


rhurling:
there was already a pull request #727 , but for the develop branch.
this pull request is against the one-dot-one branch so that the version can be bumped and pushed to w.org

rmccue:
Thanks for the patch! This needs to wrap `$thumbnail` in `json_ensure_response`, as well as checking for an error before using it (`is_wp_error( $thumbnail )`)

rhurling:
Now that 1.2 is released, could someone move this Pull Request to the 1.2.1 Milestone?
Or should I create a new one (against the master branch)?

danielbachhuber:
This is already fixed on master https://github.com/WP-API/WP-API/blob/master/lib/class-wp-json-media.php#L436

PR #935, opened 2015-02-27
Switch CORS headers callback to new action

### LINKED ISSUE #932: When is json_serve_request filter applied for CORS?
I'm having trouble understanding when the `json_serve_request` filter is applied during a request. The filter is added in **json-rest-api/plugin.php:406** and points to a function called `json_send_cors_headers`, which is defined at **:694**. As far as I can tell though, at no time during the request is `apply_filters('json_serve_request')` called and as such, the CORS headers are never sent. Am I wrong about this?

rmccue:
Fixes #932


rmccue:
@WP-API/amigos #reviewmerge

PR #936, opened 2015-02-27
Tweak the post controller

rmccue:
This is a lot of whitespace changes, but mainly things that don't need to be on separate branches.

Non-whitespace changes only: https://github.com/WP-API/WP-API/pull/936/files?w=1


rmccue:
@WP-API/amigos #reviewmerge

Issue #843, opened 2015-02-05
Add `wp/users/{id}/posts` Route Referenced in _links[archive] 


rachelbaker:
Add a route to match the archive link (wp/users/{id}/posts) that basically is a redirect to wp/posts?author={id}

Discussed in Slack Channel: https://wordpress.slack.com/archives/core-restapi/p1419895443001608

danielbachhuber:
Shouldn't it be `wp/v2/users/<id>/wp/v2/posts`? Otherwise the nature of the redirect would change when the API is versioned.

rachelbaker:
Punted to Beta 4

danielbachhuber:
I think we should use `https://api.w.org/items` as the link.

The question is: which post types do we link to?

danielbachhuber:
Wontfix on this for now.

PR #937, opened 2015-02-27
Display links in collections

### LINKED ISSUE #859: Post _links are not available from collection
There's no way to "browse" from the post collection to the individual post right now.

As a side effect, this means you also can't get a collection of posts along with their comments via embedding, but I'm actually not sure if that'd work anyway.

rmccue:
Fixes #859.

@WP-API/amigos What do you think of the approach here? There's a wee bit of duplication needed in that we're now replicating a small part of the server, but I don't think that's too big of an issue.

If we're :+1: on the approach, I'll apply the same to our other collections too.

### DIFF
--- a/lib/endpoints/class-wp-json-attachments-controller.php
+++ b/lib/endpoints/class-wp-json-attachments-controller.php
@@ -144,7 +144,7 @@ class WP_JSON_Attachments_Controller extends WP_JSON_Posts_Controller {
 	 */
 	public function prepare_item_for_response( $post, $request ) {
 		$response = parent::prepare_item_for_response( $post, $request );
-		$data = $response->data;
+		$data = $response->get_data();
 
 		$data['alt_text']      = get_post_meta( $post->ID, '_wp_attachment_image_alt', true );
 		$data['caption']       = $post->post_excerpt;
--- a/lib/endpoints/class-wp-json-attachments-controller.php
+++ b/lib/endpoints/class-wp-json-attachments-controller.php
@@ -177,9 +177,7 @@ class WP_JSON_Attachments_Controller extends WP_JSON_Posts_Controller {
 
 		$links = parent::prepare_links( $post );
 		foreach ( $links as $rel => $attributes ) {
-			$other = $attributes;
-			unset( $other['href'] );
-			$data->add_link( $rel, $attributes['href'], $other );
+			$data->add_link( $rel, $attributes['href'], $attributes );
 		}
 
 		return apply_filters( 'json_prepare_attachment', $data, $post, $request );
--- a/lib/endpoints/class-wp-json-attachments-controller.php
+++ b/lib/endpoints/class-wp-json-attachments-controller.php
@@ -144,33 +144,45 @@ class WP_JSON_Attachments_Controller extends WP_JSON_Posts_Controller {
 	 */
 	public function prepare_item_for_response( $post, $request ) {
 		$response = parent::prepare_item_for_response( $post, $request );
+		$data = $response->data;
 
-		$response['alt_text']      = get_post_meta( $post->ID, '_wp_attachment_image_alt', true );
-		$response['caption']       = $post->post_excerpt;
-		$response['description']   = $post->post_content;
-		$response['media_type']    = wp_attachment_is_image( $post->ID ) ? 'image' : 'file';
-		$response['media_details'] = wp_get_attachment_metadata( $post->ID );
-		$response['post']          = ! empty( $post->post_parent ) ? (int) $post->post_parent : null;
-		$response['source_url']    = wp_get_attachment_url( $post->ID );
+		$data['alt_text']      = get_post_meta( $post->ID, '_wp_attachment_image_alt', true );
+		$data['caption']       = $post->post_excerpt;
+		$data['description']   = $post->post_content;
+		$data['media_type']    = wp_attachment_is_image( $post->ID ) ? 'image' : 'file';
+		$data['media_details'] = wp_get_attachment_metadata( $post->ID );
+		$data['post']          = ! empty( $post->post_parent ) ? (int) $post->post_parent : null;
+		$data['source_url']    = wp_get_attachment_url( $post->ID );
 
 		// Ensure empty details is an empty object
-		if ( empty( $response['media_details'] ) ) {
-			$response['media_details'] = new stdClass;
-		} elseif ( ! empty( $response['media_details']['sizes'] ) ) {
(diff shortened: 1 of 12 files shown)


rmccue:
Note also that you can't embed items from a collection, because this is handled outside of the server.

One alternative approach I've been considering: the server could check if the response data is purely a list of `WP_JSON_Response` instances, and embed from there instead. Does that sound like a better approach? I'm not fussed either way, but that does mean special-casing at the server level.

danielbachhuber:
My current thinking on how links should be added is that the Server should sniff [attributes with](https://github.com/WP-API/WP-API/pull/943/files#diff-28e9a2b4bbf4494812d6125821ff0a7dR206) `relation=<named-route>` and auto-populate `_links` based on those defined attributes. I think `relation=<named-route>` will scale well, and solve validation problems too, etc.

rachelbaker:
@rmccue The links cannot be embedded from collection responses with this PR. Should we remove the `embeddable` key when returning a collection response for now?

rachelbaker:
I added collection links to the following endpoints:

- [x] Media
- [x] Comments
- [x] Users

We currently don't appear to have any entity links for the Taxonomies or Terms endpoints.  Anything else we are missing here?

lib/endpoints/class-wp-json-attachments-controller.php
```diff
@@ -144,33 +144,45 @@ protected function prepare_item_for_database( $request ) {
...
-		$response['media_details'] = wp_get_attachment_metadata( $post->ID );
-		$response['post']          = ! empty( $post->post_parent ) ? (int) $post->post_parent : null;
-		$response['source_url']    = wp_get_attachment_url( $post->ID );
+		$data['alt_text']      = get_post_meta( $post->ID, '_wp_attachment_image_alt', true );
+		$data['caption']       = $post->post_excerpt;
+		$data['description']   = $post->post_content;
+		$data['media_type']    = wp_attachment_is_image( $post->ID ) ? 'image' : 'file';
+		$data['media_details'] = wp_get_attachment_metadata( $post->ID );
+		$data['post']          = ! empty( $post->post_parent ) ? (int) $post->post_parent : null;
+		$data['source_url']    = wp_get_attachment_url( $post->ID );
```
danielbachhuber: Hm. Why didn't this break before?

lib/endpoints/class-wp-json-attachments-controller.php
```diff
@@ -144,33 +144,45 @@ protected function prepare_item_for_database( $request ) {
...
+		$data = json_ensure_response( $data );
+
+		$links = parent::prepare_links( $post );
+		foreach ( $links as $rel => $attributes ) {
+			$other = $attributes;
+			unset( $other['href'] );
+			$data->add_link( $rel, $attributes['href'], $other );
+		}
+
+		return apply_filters( 'json_prepare_attachment', $data, $post, $request );
```
danielbachhuber: We're passing an array in 1.x: https://github.com/WP-API/WP-API/blob/master/lib/class-wp-json-media.php#L131

How are we versioning our filters?

lib/endpoints/class-wp-json-comments-controller.php
```diff
@@ -418,37 +419,68 @@ public function prepare_item_for_response( $comment, $request ) {
...
+
+		// Wrap the data in a response object
+		$data = json_ensure_response( $data );
+
+		$links = $this->prepare_links( $comment );
+		foreach ( $links as $rel => $attributes ) {
+			$other = $attributes;
+			unset( $other['href'] );
+			$data->add_link( $rel, $attributes['href'], $other );
+		}
```
danielbachhuber: This seems to be copy and paste of other methods. Can we do some abstraction?

lib/endpoints/class-wp-json-comments-controller.php
```diff
@@ -418,37 +419,68 @@ public function prepare_item_for_response( $comment, $request ) {
...
-				'href' => json_url( '/wp/users/' . $comment->user_id ),
+				'href'       => json_url( '/wp/users/' . $comment->user_id ),
+				'embeddable' => true,
 			);
 		}
 
 		if ( 0 !== (int) $comment->comment_post_ID ) {
 			$links['post'] = array(
-				'href' => json_url( '/wp/posts/' . $comment->comment_post_ID ),
+				'href'       => json_url( '/wp/posts/' . $comment->comment_post_ID ),
```
danielbachhuber: What if this post ID isn't a post, but a page? As it stands, the route will 404.

lib/endpoints/class-wp-json-users-controller.php
```diff
@@ -439,16 +437,36 @@ public function prepare_item_for_response( $user, $request ) {
...
+	 * @param WP_Post $user User object.
+	 * @return array Links for the given user.
+	 */
+	protected function prepare_links( $user ) {
+		$links = array(
+			'self' => array(
+				'href' => json_url( sprintf( '/wp/users/%d', $user->ID ) ),
 			),
-			'archives' => array(
-				'href' => json_url( '/wp/users/' . $user->ID . '/posts' ),
```
danielbachhuber: Why are you dropping and renaming here?

danielbachhuber:
Seems like this would be a good opportunity to add more test coverage.

lib/endpoints/class-wp-json-users-controller.php
rachelbaker: @danielbachhuber the archives link never worked in v1.0, and has an enhancement issue #843

lib/endpoints/class-wp-json-attachments-controller.php
rmccue: `WP_JSON_Response` has `ArrayAccess`, so you can set it directly fine

rachelbaker: Because the parent post controller `prepare_item_for_response` now returns an object instead of an array.  I renamed the variable because we are modifying the response data not the response itself.

rachelbaker:
I added support for Terms links, and am kicking this back to @rmccue to see what else he wants to do here.

lib/endpoints/class-wp-json-comments-controller.php
rmccue: We can probably just ignore `href` inside `add_link`, which would shrink this down to a single loop instead. Maybe add a `add_links` too. Will change.

lib/endpoints/class-wp-json-attachments-controller.php
rmccue: #1119 for discussion.

PR #939, opened 2015-02-27
Ensure password cookie is cleared after use

### LINKED ISSUE #938: Post password isn't reset after temporary set
I believe this is related to the order tests are run in `WP_Test_JSON_Posts_Controller`:
* `test_get_post_with_password` has permission to access the post, so we set the correct cookie value to work around core behaviour.
* `test_get_post_with_password_without_permisson` (apart from being misspelt) does not have permission to access the post, however the cookie is set previously, and the password is the same, so the post can be accessed.

This is pretty crazy. Maybe we need to kill the workaround and fix core instead?

(Also, nice work setting the password to `always$inthebananastand` for both posts. :D)

rmccue:
Fixes #938

Doesn't actually fix it yet, but here's a test case showing the failure.

### DIFF
--- a/lib/endpoints/class-wp-json-posts-controller.php
+++ b/lib/endpoints/class-wp-json-posts-controller.php
@@ -786,6 +786,11 @@ class WP_JSON_Posts_Controller extends WP_JSON_Controller {
 	 * @return array $data
 	 */
 	public function prepare_item_for_response( $post, $request ) {
+		$previous_post = null;
+		if ( ! empty( $GLOBALS['post'] ) ) {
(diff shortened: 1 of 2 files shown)


rmccue:
Turns out this is a *huge* rabbit hole. I'll get back to fixing it eventually, but suffice to say that the visibility issue I thought existed isn't quite what is actually happening.

danielbachhuber:
Didn't solve the problem entirely, but clearing cookie after creation in 475e041ff865ffe2f0102156d29f2d3e0cdfc3f1

danielbachhuber:
Closing in favor of #1055

PR #942, opened 2015-02-27
For clarity, rename `date` and `date_gmt` as `published` and `published_gmt`

danielbachhuber:
See #605


danielbachhuber:
@WP-API/amigos Would love opinions on this.

rmccue:
I'd prefer to leave this as `date`. Although i don't feel strongly about this, I think we need good reasons to deviate from core's naming (conceptually).

As an example, the main reason we deviated from core's naming for the user fields is because things like `display_name` and `nicename` are *terrible* naming when you expose the data. Ditto `name` to `slug` for the most part. I'm not sure the same exists for `date` vs `published`, but like I said, I don't feel strongly as long as we justify it :)

danielbachhuber:
@nacin @markjaquith Opinions?

danielbachhuber:
Punt on this — I'm not tied to it.

Issue #875, opened 2015-02-11
API for registering fields to a schema


danielbachhuber:
Something like:

`register_api_field( $resource_name, $attribute_name, $get_callback, $update_callback, $schema );`

Or whatever arguments we're keen on.

rmccue:
Should probably be an array for the callbacks, for forwards-compat if we need to add extra flags.

```php
register_api_field( $resource_name, $attribute_name, array(
    'get_callback' => $get_callback,
    'update_callback' => $update_callback,
    'schema' => $schema,
) );
```

Do we potentially need a `capability_callback` here?

danielbachhuber:
We could also use this API to register many internal fields too.

TimothyBJacobs:
+1 million. #560 might be worth looking at again now. I'd planned on contributing, but then life happened. But it seems like explicit versioning would work hand in hand with a schema. Especially for things like deprecating fields and the like.

joehoyle:
### Implementation Questions

So, it seems this would be best off all implemented in the `WP_JSON_Controller` base class for the most part, as it shouldn't need to interact with the rest of the server level stuff. The only part we need to solve there is how to get the extra fields added to the params array of the endpoint. 

Things like the permissions checks, update callbacks etc can all be done automatically from the `WP_JSON_Controller` classes. Currently our implementation for subclasses is override `get_item` / `get_item_schema` etc, which are then directly called by the route callback. The issue here is that we don't get to implement any DRY middleware for getting schema, checking callbacks etc. I'd rather not have to handle applying the extra fields to the schema, checking permissions callbacks of extra fields etc in all the JSON Controllers, rather just have it implemented once in the base controller and have that be called.

The best way of doing that I can see if either enforce `parent::get_item_schema()` in the subclasses, or introduce a middleware function in the base controller `item_schema_callback` (and for all the other methods invoked via an endpoint callback.) This function would apply any generic middleware it wanted, such as adding in the extra registered fields to the schema, and then internally call `get_item_schema`.

It seems also that the extra fields could be stored on the specific `WP_JSON_Controller` such as `WP_JSON_Posts_Controller::extra_fields` so though we could have a wrapper function, it would invoke `static WP_JSON_Posts_Controller::add_extra_field( $field, ... )` this would need to be static  as we don't have singleton / shared instances. Either that, or registered an additional field would need to be stored in some global until the time it's needed.

[...]

Issue #871, opened 2015-02-11
Validate request arguments before passing to the controller


danielbachhuber:
If a client provides an invalid `format` when creating or updating a Post, the current behavior is to:

1. Create the post with all other supplied details.
2. Silently fail setting the post format (we create a WP_Error but don't do anything with it).

See https://github.com/WP-API/WP-API/blob/develop/lib/class-wp-json-posts-controller.php#L156

As a client developer, I'd expect my data to be validated before any data transformation happens, so I don't have to worry about my Resource only partially being updated.

This applies more broadly to other attributes too, not just post formats.

danielbachhuber:
More thoughts from the chat:

* Post format needs to be validated against a set of potential options.
* parent_id and author_id need to be validated against their respective objects
* post_title and post_content also should check string types because saving an object would be bad
* Possible rule to follow for the API: if invalid data has been provided for a valid field, throw an error. If invalid data has been provided for an invalid field, we can silently ignore.

Need #844 to land, then we can work on a framework for validating data based on it.

rachelbaker:
@danielbachhuber With #913 merged, is there anything left to do here?

danielbachhuber:
Yes, we need to pre-validate all of our fields before saving any data to the database.

danielbachhuber:
Here's how Github returns validation errors: https://developer.github.com/v3/#client-errors

Issue #577, opened 2014-11-04
Add object type-to-route mapping to index for discoverability


danielbachhuber:
To enhance client discovery, we'd like for endpoints to be able to register a global name that will appear in the index. For instance, core will register `posts`, `users`, `comments`, `terms`. Jetpack may register `portfolio`.

Currently our index returns a list of routes, but the individual routes don't have much metadata about the *type* of resource each represents.

The degree to which we recognize these as named content objects is TBD.

danielbachhuber:
[Pushing this to future](https://wordpress.slack.com/archives/core-restapi/p1427759475000547). It doesn't seem like we have a clear direction on it, and we need more input from client authors.

danielbachhuber:
I want to revisit this in beta2 because I think it has real-world utility for us (e.g. documentation generation)

danielbachhuber:
Punting to Future. #1222 should do the trick for our more immediate needs.

rmccue:
Closing this out; namespaces are listed in the root for discoverability of extensions to the API, and schema data is exposed with `context=help`. Combined, these should let you do all the discovery your heart desires.

PR #943, opened 2015-02-27
Validate request arguments based on schema

danielbachhuber:
See #871
Blocked by #577, #875


danielbachhuber:
The validation needs to happen in the server, which means the server needs to be aware of named routes and schemas associated with those routes.

danielbachhuber:
Joe has a good abstraction for this in #989

PR #838, opened 2015-02-05
Expose basic author details when user has published posts

danielbachhuber:
See #837

Issue #702, opened 2014-12-09
Resolve relationship between context and capabilities


danielbachhuber:
While working on #701, I realized that `context = edit` can be passed to any `GET /wp/users` request to get full user details. Which, in hindsight, makes sense, because if a user has the `list_users` cap, they conceivably can get all of those details out of the Manage Users table.

However, we'll shortly need to remove the current `list_users` cap check at the beginning of each method because a basic amount of user data will need to be exposed if the user has at least one published post (#695).

The `master` branch has `json_check_post_permission()`, which no one is really a fan of.

In short, the relationship between contexts and capabilities is a bit muddy right now. We should make it much clearer.

Maybe we should reintroduce `check_context_permission()` for controllers? https://github.com/WP-API/WP-API/pull/421/files#diff-08b662d17a90b99fbd5dc8592dafd6c9R124

Related #605

danielbachhuber:
With #1049, I'm much happier with things now. Contexts are defined to a field in the schema, which makes it much easier to see when they're exposed. Also, with the abstraction we've been working on, contexts are much more closely correlated with capabilities.

danielbachhuber:
@joehoyle Any thoughts on remaining steps for this?

joehoyle:
I don't really have anything to add here - I like the abstraction with putting context in the schema, but beyond that I'm not convinced if we can "resolve the relationship" for much more of a degree. The more I work the capabilities, the more I come to the conclusion that they can't be done declaratively, despite the "declarative" api they are meant to do. Having declarative caps for different contexts isn't realistic I think, so I'm happy with the procedural checks we have, which is now a lot saner since your most recent PR on this ticket.

danielbachhuber:
I'm pretty happy with the abstractions we have now (permission callback and context as a part of a schema). I think we start the conversation again in the future as merited.

PR #948, opened 2015-02-28
Throw error if request uses unavailable `edit` context

### LINKED ISSUE #861: Passing context=edit to a user when unauthorized does not give an error
We should always be explicit if you try and access data without permission. Introduced in #838.

danielbachhuber:
Fixes #861

### DIFF
--- a/lib/endpoints/class-wp-json-users-controller.php
+++ b/lib/endpoints/class-wp-json-users-controller.php
@@ -53,6 +53,10 @@ class WP_JSON_Users_Controller extends WP_JSON_Controller {
 			return new WP_Error( 'json_user_invalid_id', __( 'Invalid user ID.' ), array( 'status' => 404 ) );
 		}
 
+		if ( ! empty( $request['context'] ) && 'edit' === $request['context'] && ! current_user_can( 'edit_user', $user->ID ) ) {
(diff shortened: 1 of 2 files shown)


danielbachhuber:
@rmccue How do you feel about this forced downgrade logic? https://github.com/WP-API/WP-API/pull/838/files#diff-6a242a8d43b6c0aa73da8f01c1324859R63

I think this is another instance of #702 — there's a loose coupling between contexts and capabilities right now that's causing regular problems.

rmccue:
> How do you feel about this forced downgrade logic?

I'd *much* rather return an error; that's the main point of contexts in the first place. As I said on #702:

> Additionally, right now setting `context` either succeeds (returns all data for that context) or fails (returns no data/an error), which is kind of nice in that you immediately know if you can't access the data.

rmccue:
With this pull request: this doesn't handle `/comments` as well, correct?

danielbachhuber:
Correct, it doesn't. I don't think the solution is to fix all of these context-related issues in a one-off manner. If we want contexts to have a strong correlation to capability checks, we need a better abstraction

danielbachhuber:
Continuing in #1033

Issue #894, opened 2015-02-18
JSON Schema: How should we report `context` arguments adding data to a response?


danielbachhuber:
Our `context` parameter can add data to User and Post responses. We should figure out how this variable data can be communicated in the JSON Schema representation.

danielbachhuber:
The leading idea I have right now is that we always include the fields in the response, but nullify the values when the data is supposed to be inaccessible.

rachelbaker:
@danielbachhuber Anything still open here, or can we close this issue?

danielbachhuber:
I still need to port Comments over, and Meta and Revisions when they're ready.

danielbachhuber:
This has been addressed in all of our endpoints.

Issue #893, opened 2015-02-18
Comment author email address probably shouldn't be exposed by default


danielbachhuber:
If we aren't exposing email address for users by default.

Only affects `develop`

rachelbaker:
@danielbachhuber Good point.  Should we return the `email` field as `false`?  

That would match the default behavior of the `users` endpoint response, and the .com API: https://developer.wordpress.com/docs/api/1.1/get/sites/%24site/comments/

Also needs unit tests so this "never happens again" :hammer:

danielbachhuber:
> Should we return the email field as false?

I think it's contingent on #894

On `master`, it's nice that the author data is normalized to some degree. But it doesn't really match our other resources.

danielbachhuber:
> But it doesn't really match our other resources.

Or, we could include `author_name`, `author_email`, and `author_avatar` as a part of the `Post` response.

rmccue:
@rachelbaker Not sure why this one's "Blocked on Core"?

rachelbaker:
@rmccue Me neither.  Label removed.

PR #945, opened 2015-02-28
POC query params declaration and validation for Comments and Terms

danielbachhuber:
* Describe collection query params in a structured manner
* Validate based on definition, and return error if arguments don't validate

See #871, #924

PR #949, opened 2015-02-28
POC Use schema to limit returned fields for context

danielbachhuber:
See #894, #893, #605

### DIFF
--- a/lib/infrastructure/class-wp-json-server.php
+++ b/lib/infrastructure/class-wp-json-server.php
@@ -638,6 +638,19 @@ class WP_JSON_Server {
 					$response = json_ensure_response( $response );
 				}
 
+				$data = $response->get_data();
+				$properties = ! empty( $handler['schema']['properties'] ) ? $handler['schema']['properties'] : array();
+				foreach( $properties as $key => $field_args ) {
+					if ( ! isset( $field_args['context'] ) ) {
+						continue;
+					}
+					$context = (array) $field_args['context'];
+					if ( ! in_array( $request['context'], $context ) ) {
+						$data[ $key ] = null;
+					}
+				}
+				$response->set_data( $data );
+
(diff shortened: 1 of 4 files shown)


lib/infrastructure/class-wp-json-server.php
```diff
@@ -638,6 +638,19 @@ public function dispatch( $request ) {
...
 				}
 
+				$data = $response->get_data();
+				$properties = ! empty( $handler['schema']['properties'] ) ? $handler['schema']['properties'] : array();
+				foreach( $properties as $key => $field_args ) {
+					if ( ! isset( $field_args['context'] ) ) {
+						continue;
+					}
+					$context = (array) $field_args['context'];
+					if ( ! in_array( $request['context'], $context ) ) {
```
danielbachhuber: There's a problem with this code in that, for create / update requests, the `$request['context']` is set within the controller's method, and not exposed back to the server:

```
NULL
status
array (
  0 => 'embed',
  1 => 'view',
  2 => 'edit',
)
NULL
type
array (
  0 => 'embed',
  1 => 'view',
  2 => 'edit',
)
```

For create / update requests, we should probably have `context=edit` by default. And it would be nice to get some final context from the response object.

plugin.php
```diff
@@ -367,6 +367,7 @@ function create_initial_json_routes() {
 			'methods'   => WP_JSON_Server::READABLE,
 			'callback'  => array( $controller, 'get_items' ),
 			'permission_callback' => array( $controller, 'get_items_permissions_check' ),
+			'schema'    => $controller->get_item_schema(),
```
danielbachhuber: This really sucks. It would be better if the server just knew which named resource it was dealing with, and how to find that schema.

Related #945, #577

danielbachhuber:
Much better implementation in #1049

Issue #950, opened 2015-02-28
WP_JSON_Comments_Controller::get_item() should error if user can't context=edit


danielbachhuber:
Related https://github.com/WP-API/WP-API/pull/948#issuecomment-76545260, #702

danielbachhuber:
This was fixed in #951 it appears:

```
local ➜  ~  http GET wordpress-develop.dev/wp-json/wp/v2/comments/1 context==edit --body
{
    "code": "rest_forbidden_context",
    "data": {
        "status": 401
    },
    "message": "Sorry, you cannot view this comment with edit context."
}
```

PR #951, opened 2015-02-28
Require `moderate_comments` capability to context=edit a Comment

danielbachhuber:
If a comment author created a comment without the `moderate_comments`
capability, use context=view

Don't use `edit_comment` capability because `moderate_comments` can be
used to view the Comment list table without granting edit access

See #950, #702

### DIFF
--- a/lib/endpoints/class-wp-json-comments-controller.php
+++ b/lib/endpoints/class-wp-json-comments-controller.php
@@ -316,7 +316,7 @@ class WP_JSON_Comments_Controller extends WP_JSON_Controller {
 	 * Check if a given request has access to read the comment
 	 * 
 	 * @param  WP_JSON_Request $request Full details about the request.
-	 * @return bool
+	 * @return bool|WP_Error
 	 */
 	public function get_item_permissions_check( $request ) {
 		$id = (int) $request['id'];
--- a/lib/endpoints/class-wp-json-comments-controller.php
+++ b/lib/endpoints/class-wp-json-comments-controller.php
@@ -47,6 +47,10 @@ class WP_JSON_Comments_Controller extends WP_JSON_Controller {
 			return new WP_Error( 'json_comment_invalid_id', __( 'Invalid comment ID.' ), array( 'status' => 404 ) );
 		}
 
+		if ( ! empty( $request['context'] ) && 'edit' === $request['context'] && ! current_user_can( 'moderate_comments' ) ) {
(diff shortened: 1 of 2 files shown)


danielbachhuber:
@WP-API/amigos #reviewmerge

tests/test-json-comments-controller.php
```diff
@@ -116,6 +117,13 @@ public function test_get_comment_invalid_id() {
 		$this->assertErrorResponse( 'json_comment_invalid_id', $response, 404 );
 	}
 
+	public function test_get_comment_invalid_context() {
+		$request = new WP_JSON_Request( 'GET', sprintf( '/wp/comments/%s', $this->approved_id ) );
+		$request->set_param( 'context', 'edit' );
+		$response = $this->server->dispatch( $request );
+		$this->assertErrorResponse( 'json_comment_cannot_view', $response, 403 );
```
joehoyle: With the latest merge of permissions checks, I'd expect this to return a generic `json_forbidden` now instead.

lib/endpoints/class-wp-json-comments-controller.php
```diff
@@ -47,6 +47,10 @@ public function get_item( $request ) {
 			return new WP_Error( 'json_comment_invalid_id', __( 'Invalid comment ID.' ), array( 'status' => 404 ) );
 		}
 
+		if ( ! empty( $request['context'] ) && 'edit' === $request['context'] && ! current_user_can( 'moderate_comments' ) ) {
```
joehoyle: We should move this check to the permissions check callback instead me things now

lib/endpoints/class-wp-json-comments-controller.php
```diff
@@ -329,6 +337,10 @@ public function get_item_permissions_check( $request ) {
 			return false;
 		}
 
+		if ( ! empty( $request['context'] ) && 'edit' === $request['context'] && ! current_user_can( 'moderate_comments' ) ) {
+			return new WP_Error( 'json_forbidden', __( 'Sorry, you cannot view this comment with edit context' ), array( 'status' => 403 ) );
```
danielbachhuber: Shouldn't this return `bool`?

joehoyle: The caps check can return either. I kept the `WP_Error` so we can have a custom error message in this case.

danielbachhuber: Can you update the PHPdoc then?

joehoyle: 804cadd

PR #952, opened 2015-02-28
Allow creating / updating posts without title, content, or excerpt

### LINKED ISSUE #745: Can't update the post with empty content, title or excerpt

danielbachhuber:
Fixes #745

### DIFF
--- a/lib/endpoints/class-wp-json-posts-controller.php
+++ b/lib/endpoints/class-wp-json-posts-controller.php
@@ -147,7 +147,9 @@ class WP_JSON_Posts_Controller extends WP_JSON_Controller {
 		}
 
 		$post->post_type = $this->post_type;
+		add_filter( 'wp_insert_post_empty_content', '__return_false', 100 );
 		$post_id = wp_insert_post( $post, true );
+		remove_filter( 'wp_insert_post_empty_content', '__return_false', 100 );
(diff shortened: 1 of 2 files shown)


danielbachhuber:
@WP-API/amigos #reviewmerge

lib/endpoints/class-wp-json-posts-controller.php
```diff
@@ -147,7 +147,9 @@ public function create_item( $request ) {
 		}
 
 		$post->post_type = $this->post_type;
+		add_filter( 'wp_insert_post_empty_content', '__return_false', 100 );
```
rmccue: I don't understand why we want this? The check in core exists for a reason.

danielbachhuber: For what reason?

rmccue: Disabling parts of core code isn't something that we should do without really good reasons. The code here ensures that the data can't be changed to an invalid state: that is, posts must have *some* data (title, content, or excerpt) in the database to be valid.

@iseulde's original issue states:
> I know you can't update a post without the content, title AND excerpt empty, but you can leave two fields blank if one is filled. The API fails to do that.

That's the only issue that needs solving there: **updating** one field without needing to send the rest of the data (partial updates). This PR doesn't solve that issue.

PR #953, opened 2015-02-28
Allow `title` to be set to empty string in request

### LINKED ISSUE #745: Can't update the post with empty content, title or excerpt

danielbachhuber:
Fixes #745


danielbachhuber:
@WP-API/amigos #reviewmerge

PR #955, opened 2015-03-01
Change comment author to structured data

### LINKED ISSUE #893: Comment author email address probably shouldn't be exposed by default

rmccue:
For comments, we currently have destructured author data in the following form:

```json
{
    "author": "admin",
    "author_email": "admin@example.com",
    "author_url": ""
}
```

Why did we switch from structured data (`author: { ... }`) to destructured data in the first place? Is anyone opposed to changing that back?

Also fixes #893.

### DIFF
--- a/lib/endpoints/class-wp-json-comments-controller.php
+++ b/lib/endpoints/class-wp-json-comments-controller.php
@@ -289,9 +289,11 @@ class WP_JSON_Comments_Controller extends WP_JSON_Controller {
 			'post'         => (int) $comment->comment_post_ID,
 			'parent'       => (int) $comment->comment_parent,
 			'user'         => (int) $comment->user_id,
-			'author'       => $comment->comment_author,
-			'author_email' => $comment->comment_author_email,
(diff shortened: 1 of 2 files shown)


danielbachhuber:
Destructured data is easier to use, and more consistent with the rest of the API

rmccue:
> Destructured data is easier to use

I'm not sure exactly what you mean by "easier to use". There are contexts certainly where destructured data is easier to use; Backbone models are a good example of that (with `model.get()`). However, structured data is a lot easier to use in other contexts, such as Handlebars templates:

```
{{#with author}}
    <h2>By <a href="{{ url }}">{{ name }}</a></h2>
{{/with}}
```

I don't think there's a clear benefit to destructured data in that context.

> and more consistent with the rest of the API

More consistent with what exactly? One advantage of having structured data for comment authors is that you can get a user-like object, which means you can share code for things like views.

rachelbaker:
Merged #955

PR #958, opened 2015-03-01
Use real URLs instead of query_params attribute

### LINKED ISSUE #954: Remove `query_params` link hack
Back in #836, we added `query_params` as a special field in links to allow passing extra parameters with a link. This solved the problem of passing through data for a link.

However, this breaks the links system when not embedding, as we're now not linking to the object correctly. For example, linking to a single post's comments:
```json
{
    "embeddable": true,
    "query_params": {
        "post_id": 20
    },
    "href": "http://api.local/wp-json/wp/comments"
}
```

Rather than using another system here, let's use `add_query_arg` to construct a URL properly, and ensure that gets passed through via embedding correctly.

(See also, #860)

rmccue:
Fixes #954


rmccue:
@rachelbaker This is undoing your change in #836, so I'll leave to you to #reviewmerge :)

rachelbaker:
Merged #958

PR #959, opened 2015-03-01
Ensure post deletion is idempotent

### LINKED ISSUE #789: Deleting an item should always delete an item
Core has this weird pattern where you can pass `force = true` when deleting a post or comment. However, `force = false` can produce this behavior:

* Post isn't yet trashed => Post is marked as "trash", but not deleted.
* Post is already trashed => Post is permanently deleted.

This behavior is inconsistent, and an idiosyncrasy of core internals.

In the API, `DELETE <id>` should always delete the item. To trash an item, clients should be expected to `PUT <id> status=trash`, just like any other status.

Related #618.

rmccue:
Practically speaking: a DELETE request to a post resource will ensure the post is trashed. If the post is already trashed, no change will occur. To permanently delete, pass force=true, and it will be permanently deleted regardless of status.

This is a change from the previous behaviour of the first DELETE trashing the post, and the second DELETE permanently deleting it.

See #789.

### DIFF
--- a/lib/endpoints/class-wp-json-posts-controller.php
+++ b/lib/endpoints/class-wp-json-posts-controller.php
@@ -59,7 +59,9 @@ class WP_JSON_Posts_Controller extends WP_JSON_Controller {
 				'callback' => array( $this, 'delete_item' ),
 				'permission_callback' => array( $this, 'delete_item_permissions_check' ),
 				'args'     => array(
-					'force'    => array(),
+					'force'    => array(
+						'default'      => false,
+					),
 				),
 			),
 		) );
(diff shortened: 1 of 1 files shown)


rmccue:
The second commit here also returns a 410 Gone if you call `DELETE` when it's already trashed. Anyone have opinions on this? 410 might be a bit too permanent, so maybe a 400 or 404 is a better code here.

rmccue:
As reflected by the low line-count change, this is a pretty simple fix for the bug. This isn't the full solution to #789, as it's clear we need more than this in any case, but the bug needs to be fixed before we can ship beta 1.

PR #960, opened 2015-03-01
[WIP] Convert meta endpoints to new-style

### LINKED ISSUE #928: Add New Style Meta Endpoints
For backwards compatibility.  

Assigning to @rmccue

rmccue:
Fixes #928.


rmccue:
Alright, the controllers are migrated, but still needs unit tests converted across. (I suspect I'll find further bugs as those are migrated.)

plugin.php
```diff
@@ -115,6 +113,11 @@ function create_initial_json_routes() {
 		}
 
 		$controller->register_routes();
+
+		if ( post_type_supports( $post_type->name, 'custom-fields' ) ) {
+			$meta_controller = new WP_JSON_Meta_Posts_Controller( $post_type->json_base );
```
rmccue: There should probably be a way to override this; I'm thinking `$post_type->json_meta_controller_class`, which defaults to the above?

plugin.php
```diff
@@ -198,13 +201,6 @@ function json_api_maybe_flush_rewrites() {
  * @param WP_JSON_Server $server Server object.
  */
 function json_api_default_filters( $server ) {
-
-	// Post meta.
-	$wp_json_post_meta = new WP_JSON_Meta_Posts();
-	add_filter( 'json_endpoints',    array( $wp_json_post_meta, 'register_routes'    ), 0 );
-	add_filter( 'json_prepare_post', array( $wp_json_post_meta, 'add_post_meta_data' ), 10, 3 );
-	add_filter( 'json_insert_post',  array( $wp_json_post_meta, 'insert_post_meta'   ), 10, 2 );
```
rmccue: Not sure how best to integrate the controllers. Maybe we don't?

We need to link from the post to meta in any case, which is probably best handled through having meta poke into the main post controller and add the data, but I can't think of a nice implementation there.

rmccue:
@WP-API/amigos There's one major thing missing from this with regards to Feature Parity right now: meta is no longer exposed on the post (with edit context) at all. Are we happy to keep it this way (separate endpoint only), or should we re-add that?

Test coverage is looking good, apart from no coverage on the inline methods mentioned above; I'll remove those if we're happy to, otherwise add in tests and actual usage.

rmccue:
(I have user meta ready to go as a separate PR, but I'll file that after this one's merged to keep the PR clean, as it branches from this. That'll test the ID column method correctly.)

BenjaminBarnett:
Ryan, haven't been able to get an answer w/r/t multisite reintegration with
the current build. Is this on the short list? Thank you.


______________________
Benjamin Barnett, co-owner
media schmedia LLC
media-schmedia.com
mightylittlewebsites.com
216-291-4313

rmccue:
> Ryan, haven't been able to get an answer w/r/t multisite reintegration

Answered on WP-API/multisite#2.

danielbachhuber:
> Are we happy to keep it this way (separate endpoint only), or should we re-add that?

Keep them separate. We can revisit when we do custom taxonomies.

lib/endpoints/class-wp-json-meta-posts-controller.php
```diff
@@ -1,19 +1,24 @@
 <?php
 
-class WP_JSON_Meta_Posts extends WP_JSON_Meta {
+class WP_JSON_Meta_Posts_Controller extends WP_JSON_Meta_Controller {
```
danielbachhuber: Rather than subclassing, let's pass `$type` to `WP_JSON_Meta_Controller`, similar to `WP_JSON_Posts_Controller`

lib/endpoints/class-wp-json-meta-controller.php
```diff
@@ -33,44 +33,53 @@ public function __construct() {
...
+		$single_args = $base_args + array(
+			'mid'   => array(
+				'required' => true,
+			),
+		);
+
+		$data_args = array(
+			'key'   => array(),
+			'value' => array(),
+		);
```
danielbachhuber: Can you add a defined schema please? These values can be derived from that.

plugin.php
danielbachhuber: I think we can just register by default for posts, users and comments, and figure out the abstraction in a later beta.

danielbachhuber:
Couple of important notes from me.

rachelbaker:
Still needed here:

- [ ] Meta schema & schema route
- [ ] Meta links in endpoint responses
- [ ] Links to the meta routes in Posts responses

rachelbaker:
I was hoping to include a _link in the Posts response to the meta endpoint, but cannot seem to find the appropriate relationship: http://www.iana.org/assignments/link-relations/link-relations.xhtml#link-relations-1

rachelbaker:
I am going to be out for the next 2-3 days, and haven't found a great approach for the remaining open items here.  @rmccue or @joehoyle welcome to pick this up, if you can

rmccue:
> Meta schema & schema route

Done.

> Meta links in endpoint responses
> Links to the meta routes in Posts responses

Not sure what the difference between these is? I've added the link from the post out to the meta items, was the other just the reverse of that?

rmccue:
This is no longer `[WIP]` and is up for review.

One note: @danielbachhuber said above:
> Rather than subclassing, let's pass `$type` to `WP_JSON_Meta_Controller`, similar to `WP_JSON_Posts_Controller`

Unlike the posts controller, we have two fundamentally different objects (posts and users, with potentially comments and terms) that have different permissions checks and such. For example, comment meta would need to check the comment and potentially also the parent post.

@WP-API/amigos #reviewmerge

danielbachhuber:
> Unlike the posts controller, we have two fundamentally different objects (posts and users, with potentially comments and terms) that have different permissions checks and such.

Couldn't we have some conditional logic based on `$type` inside of the permissions callback?

Pages are fundamentally different than Posts, and we've accommodated them both just fine in the Posts controller.

@joehoyle thoughts?

rmccue:
> Couldn't we have some conditional logic based on `$type` inside of the permissions callback?

We have a small amount right now, yeah (`get_id_column`), but I'm tempted to even move that to the subclasses.

Pages and posts are both fundamentally CPTs, so they follow a bunch of similar patterns. Both store meta in the database equally, both have the same system to check permissions, etc. Posts and users however are a fair bit different.

We *could* do it, but we'll end up with some kinda nasty conditionals in the permission checks. As is, the permission checks are essentially the only place we need to subclass right now, so I think it's fine as-is. We can always revisit and refactor if we need to when we add user meta.

PR #966, opened 2015-03-03
Add missing core path to post endpoint link hrefs.

rachelbaker:
Fixes missing `/wp/` path in the 'href' values for the collection, up, and version-history `_links`.


rachelbaker:
@WP-API/amigos #reviewmerge

PR #967, opened 2015-03-03
Allow HTTP method to be overwritten by HTTP_X_HTTP_METHOD_OVERRIDE

tlovett1:
Simply check `$_SERVER['HTTP_X_HTTP_METHOD_OVERRIDE']` and if it is set, use it as the request method.

### DIFF
--- a/lib/infrastructure/class-wp-json-server.php
+++ b/lib/infrastructure/class-wp-json-server.php
@@ -200,7 +200,12 @@ class WP_JSON_Server {
 			}
 		}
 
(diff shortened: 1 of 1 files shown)


rmccue:
@tlovett1 :+1:, but can you unify the handling with the `$_GET['method']` check please? (I don't mind if you change this one to match existing or change the existing to match this, but both should be consistent.)

tlovett1:
How's that look?

rmccue:
> How's that look?

Perfect, thanks!

PR #968, opened 2015-03-04
Fix attachment caption and description fields

danielbachhuber:
`caption` is stored in `post_excerpt`, `description` is stored in
`post_content`. They were switched.

See
https://core.trac.wordpress.org/browser/trunk/src/wp-includes/media.php#L2701


danielbachhuber:
@WP-API/amigos #reviewmerge

rachelbaker:
Merged #968

PR #970, opened 2015-03-04
Move the Route Registering to the Controllers

joehoyle:


joehoyle:
@WP-API/amigos #reviewmerge

lib/endpoints/class-wp-json-posts-controller.php
```diff
@@ -9,6 +9,76 @@ public function __construct( $post_type ) {
 	}
 
 	/**
+	 * Register the routes for the objects of the controller
+	 * 
+	 */
+	public function register_routes() {
+
+		parent::register_routes();
```
danielbachhuber: If someone misses this call, the sky falls on their head?

lib/endpoints/class-wp-json-post-types-controller.php
```diff
@@ -3,6 +3,33 @@
...
+
+		register_json_route( 'wp', '/types/schema', array(
+			'methods'         => WP_JSON_Server::READABLE,
+			'callback'        => array( $this, 'get_item_schema' ),
+		) );
+
+		register_json_route( 'wp', '/types/(?P<type>[\w-]+)', array(
+			'methods'         => WP_JSON_Server::READABLE,
+			'callback'        => array( $this, 'get_item' ),
+		) );
```
danielbachhuber: Should we make each of these register instances separate callbacks, so someone could subclass and override? How would someone go about disabling the registry of a route?

lib/endpoints/class-wp-json-posts-controller.php
danielbachhuber: Do we even want them to necessarily re-register the parent's routes?

joehoyle: Perhaps not actually

lib/endpoints/class-wp-json-post-types-controller.php
```diff
@@ -3,6 +3,33 @@
 class WP_JSON_Post_Types_Controller extends WP_JSON_Controller {
 
 	/**
+	 * Register the routes for the objects of the controller
+	 * 
+	 */
+	public function register_routes() {
+
+		parent::register_routes();
```
rmccue: Let's remove this unless it's needed. The base controller shouldn't need to register endpoints, right?

lib/endpoints/class-wp-json-posts-controller.php
rmccue: Let's remove this unless it's needed. The base controller shouldn't need to register endpoints, right?

joehoyle: The idea was more to potentially have some DRY code in the base for things like schema endpoints, but after thinking on that I don't think we can do it automajically very well anyway

lib/endpoints/class-wp-json-post-types-controller.php
rmccue: I think it's fine as-is here, but we probably need a `deregister_json_route` as well.

One thing I did like from v1 was that you took the `$routes` array and returned your modified version (filter-style), which meant you could `unset` anything you didn't want. Don't know how that would work with `register_json_route` though.

lib/endpoints/class-wp-json-comments-controller.php
```diff
@@ -6,6 +6,123 @@
 class WP_JSON_Comments_Controller extends WP_JSON_Controller {
 
 	/**
+	 * Register the routes for the objects of the controller
+	 *
```
rmccue: Extra line in the docblock should be trimmed here, and the description should have a trailing period.

lib/endpoints/class-wp-json-post-types-controller.php
joehoyle: Created #972 for the deregistering

joehoyle:
:clap:

PR #971, opened 2015-03-04
Move validation to the `WP_JSON_Request` class

danielbachhuber:
It's more semantic, and makes validation more accessible

See #871


danielbachhuber:
Build passes locally.

Issue #719, opened 2014-12-11
Refresh documentation


danielbachhuber:
In the near future, we should do a refresh of the documentation, including:

* [ ] Our new Controller pattern, with associated Controller test pattern.
* [ ] Request class.
* [ ] Enveloping / batch requests.
* [ ] Updating all endpoint arguments.

(Feel free to edit this checklist).

danielbachhuber:
@pollyplummer any opinions on what should be added to that list, and how we should go about this? I'm ready to kick things off

pollyplummer:
@danielbachhuber Most of those listed seem like they'd fall under the Guides section on wp-api.org. Those are fairly easy to add as new files. The site currently uses the "master" branch of WP-API to pull from, but we probably don't want to add new guides or update the routes.md file in the master until the develop branch is pushed through to the plugin. In the meantime, we could create a new docs-develop branch in the wp-api.org repo for adding new guides that pertain to the upcoming changes to the plugin. Thoughts on this approach? @rachelbaker @rmccue

danielbachhuber:
Documentation is now covered by the docs repo https://github.com/WP-API/docs-v2/issues

PR #983, opened 2015-03-12
Update documentation

govaleriia:
Documentation is lacking some essential information (use of 'page' instead of '[page]' for example).


rachelbaker:
@govaleriia Documentation for the `develop` branch is being tracked with #719.  I am closing this PR because we don't want to merge master into develop.

PR #989, opened 2015-03-16
Support for validation / sanitize callbacks in arguments

joehoyle:
This PR adds support for sanitizing and validating of arguments in the route registration. The two are of course providing different purposes:

- **Sanitize Callback** - this is called as early as it can be on the args in `WP_JSON_Request`, it's primary focus is to clean the input data. For example, simple things: `absint`, `sanitize_key`, `sanitize_text_field` etc. Another potentially nice thing that sanitize_callback could do is expand objects, for example if i have an arg `post_id` my `sanitize_callback` could transfor that into a `WP_Post` object to be used downstream.
- **Validate Callback** - this is use to flag a param value as being invalid before the main callback for the endpoint is called. This should allow us to pull a fair amount of loginc out of the main callback. For example, validate callback would check if a post actually exists (after the supplied id has been cleaned by the `sanitize_callback`. 

Looking for feedback on this idea generally :)


danielbachhuber:
My primary concern is that it isn't declarative, and we're trying to avoid procedural code as much as possible in the design pattern. How could it incorporate some ideas from #945?

joehoyle:
@danielbachhuber I don't see at as procedural vs. declarative. Procedural code is always needed to build a declarative abstraction, which is what I see this PR as being. Whether we build (which I think we maybe should) a declarative way to specify arg types, we need the underlying API to build off.

Declarative is fine 90% of the time, until you get something very custom - for example, the 'per_page' arg is of type `integer` but also can't be above 100; having the lower level API underneath to support it is valuable.

I don't want to get down the declarative rabbit whole ATM, but I don't see this as an opposition to that either way.

danielbachhuber:
> Procedural code is always needed to build a declarative abstraction, which is what I see this PR as being.

Sounds good.

joehoyle:
@rmccue @rachelbaker I added the code for the terms controller so show how this would be utilized in practice. There wasn't much use for validation here - and so it happens pretty much the only thing I wanted to validate was the taxonomy name, term id etc - however those vars are part of the URL (for arguments) which this PR does not address.

The "next step" that was decided in the last meeting was to see what this looks like in use - so that's what this now is.

danielbachhuber:
Hm, now that I think of it, much (if not all) of this sanitization happens in core internals already (e.g. `sanitize_term_field()`. Do we even need sanitization in the API, or just validation?

joehoyle:
I think we should have sanitization for completeness either way. We are also already doing a lot of sanitizing in the API code, so we should either remove that, or move it to the api provided here.

I think also 3rd parties will appreciate it.

joehoyle:
- [ ] Add unit tests

lib/endpoints/class-wp-json-terms-controller.php
```diff
@@ -16,12 +16,29 @@ public function register_routes() {
...
+						'sanitize_callback' => 'absint',
+						'default'           => 10,
+					),
+					'page'     => array(
+						'sanitize_callback' => 'absint',
+						'default'           => 1,
+					),
+					'order'    => array(
+						'sanitize_callback' => 'sanitize_key',
+						'default'           => '',
```
rachelbaker: @joehoyle Any reason why the default here isn't 'ASC' the default for `get_terms()`?

lib/endpoints/class-wp-json-terms-controller.php
```diff
@@ -16,12 +16,29 @@ public function register_routes() {
...
+					),
+					'orderby'  => array(
```
rachelbaker: @joehoyle Any reason why the default here isn't 'name' the default for `get_terms()`?

rachelbaker:
This PR currently only adds this feature to the Terms routes.  We still have to add for:

- [ ] Posts
- [ ] Comments
- [ ] Users
- [ ] Taxonomies
- [ ] Post Types

joehoyle:
@rachelbaker I'm happy to do it for those two - though I'd rather merge this and then do those one at a time otherwise I feel like we'll be fixing conflicts for a while!

Issue #961, opened 2015-03-02
Better HTTP status code when `wp_insert_post()` or `wp_update_post()` fails


danielbachhuber:
Currently, we get a non-descriptive 500 status code when `wp_insert_post()` or `wp_update_post()` fails. We should make the potential error states more descriptive.

From #952

danielbachhuber:
This was fixed in beta1

PR #999, opened 2015-03-23
Return correct response code from wp_insert_post() error

joehoyle:
From #961 we always trigger a 500 error is wp_insert_post() / wp_update_post() fails, however it's not neccesarily a problem with "us" it ould also be a problem with the request data.

I went through wp_insert_post() and put a check in for the only one which should be considered a 500 - which is a DB insert / update error. Also I made a tricky unit test to test this scenario.


joehoyle:
@WP-API/amigos #reviewmerge

PR #1000, opened 2015-03-23
If an invalid date is supplied to create / update post, return an error

joehoyle:
Currently we silently fail to not updating the date.


rachelbaker:
Merged #1000

PR #1013, opened 2015-03-25
Add alt property field to attachment_meta

### LINKED ISSUE #1011: Missing alt property field from attachment_meta
Could not retrieve the alt property for an attached image.

cies:
We found no other way to retrieve the alt property of image attachment.

This fixes #1011.

### DIFF
--- a/lib/class-wp-json-media.php
+++ b/lib/class-wp-json-media.php
@@ -95,9 +95,10 @@ class WP_JSON_Media extends WP_JSON_Posts {
 		}
 
 		// $thumbnail_size = current_theme_supports( 'post-thumbnail' ) ? 'post-thumbnail' : 'thumbnail';
-		$data['source']          = wp_get_attachment_url( $post['ID'] );
-		$data['is_image']        = wp_attachment_is_image( $post['ID'] );
-		$data['attachment_meta'] = wp_get_attachment_metadata( $post['ID'] );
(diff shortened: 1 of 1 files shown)


swissspidy:
Haven't tested this, but there's an `$data['attachment_meta']` array – why do you need to extra add `$data['attachment_meta']['alt']`?
Can't you just access it from the `attachment_meta` array? If not, why only add the `alt` field and not other meta data as well?

cies:
> Haven't tested this, but there's an $data['attachment_meta'] array – why do you need to extra add $data['attachment_meta']['alt']? Can't you just access it from the attachment_meta array?

Nope, `alt` is not a member of `attachment_meta`.

> why only add the alt field and not other meta data as well?

@swissspidy What other meta data would qualify to be added in your opinion?

swissspidy:
Well, then never mind my comment, sorry! Don't know other fields, would have to look into it.

cies:
@swissspidy No problem. If you find time to look into it, feel free to comment further.

danielbachhuber:
We aren't taking enhancements against the `master` branch now that v1.2 is released.

`alt_text` is present on the `develop` branch: https://github.com/WP-API/WP-API/blob/develop/lib/endpoints/class-wp-json-attachments-controller.php#L148

PR #1017, opened 2015-03-26
Initiated WP_JSON_Menus

ctimoteo:
Hello,

i am using Wordpress with Menu plugin (https://wordpress.org/plugins/menu/) to create menus,

i developed initial code to retrieve menus data with WP-API,

for now, only permits to get menu data from given menu name, but i will share my code anyway.

for example calling WP api with /wp-json/menu/main_menu will retrieve data from a menu called main_menu

Thanks for your work on WP-API, it's very useful!

I hope to help more on future...

Bye

### DIFF
--- a/lib/class-wp-json-menus.php
+++ b/lib/class-wp-json-menus.php
@@ -1,0 +1,102 @@
+<?php
+
+/**
+ * Plugin Name: WP_JSON_Menus
+ * Description: WP REST API Plugin to retrieve menus from wordpress
+ * Version: 1.0.0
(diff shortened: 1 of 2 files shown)


danielbachhuber:
Thanks for your work on this, @ctimoteo. We'll be going a bit different of a direction on menus when we come around to them. Stay tuned!

unfulvio:
@ctimoteo  a bit old issue, but if you're still looking for a menu solution I hope this could work for you: https://github.com/unfulvio/wp-api-menus

Issue #1009, opened 2015-03-25
Cannot use object of type WP_JSON_Response as array in class-wp-json-media.php on line 437


jchoe:
I am using version 1.2.

When I pass 'featured_image' with a post, I get an error

Cannot use object of type WP_JSON_Response as array in .... class-wp-json-media.php on line 437

It is at method, attach_thumbnail, which is invoked from class-json-rest-api/lib/class-wp-json-posts.php:967 (do_action( 'json_insert_post', $post, $data, $update ))

This error occurs because class-json-rest-ap's get_post method returns a WP Response object.  class-json-rest-ap's get_post is invoked by class-wp-json-media's get_post which in turn called within attach_thumbnail.  

I got around to it by not invoking class-json-rest-ap's get_post method, instead of returning the local $post object from class-wp-json-media's get_post.  (line 81 of lass-wp-json-media)

However, this could not be a general solution.

rachelbaker:
Fixed in #1024

PR #1024, opened 2015-03-28
Bug fix: upload attachment error

jvandijk:
The get_post method does not return an array, but a WP_JSON_Response object. Because of this, the ID of the thumbnail is only to be found in the data variable of the class.

This PR fixes issue : #1009.


rachelbaker:
@jvandijk Thank you for the bugfix.

Merged #1024

PR #1026, opened 2015-03-29
Add embeddable attachments to Post response _links

rachelbaker:
Adds the `attachments` key to the `_links` object in a Post response for discoverability, which is also embeddable.  This does require allowing the `post_parent` query parameter to be public, which the WordPress.com API already allows: https://developer.wordpress.com/docs/api/1.1/get/sites/%24site/posts/.

### DIFF
--- a/lib/endpoints/class-wp-json-posts-controller.php
+++ b/lib/endpoints/class-wp-json-posts-controller.php
@@ -1124,7 +1124,7 @@ class WP_JSON_Posts_Controller extends WP_JSON_Controller {
 			);
 		}
 
-		if ( 'attachment' !== $post->post_type ) {
+		if ( ! in_array( $post->post_type, array( 'attachment', 'nav_menu_item', 'revision' ) ) ) {
 			$attachments_url = json_url( 'wp/media' );
 			$attachments_url = add_query_arg( 'post_parent', $post->ID, $attachments_url );
 			$links['attachments'] = array(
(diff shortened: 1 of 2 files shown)


lib/endpoints/class-wp-json-posts-controller.php
```diff
@@ -1124,6 +1124,15 @@ protected function prepare_links( $post ) {
 			);
 		}
 
+		if ( 'attachment' !== $post->post_type ) {
```
danielbachhuber: Do all post types support attachments? Revisions? Menu items?

Or, phrased another way, does core have an explicit way of denoting a post supports attachments?

rachelbaker: @danielbachhuber Core doesn't have an explicit way to note if a post supports attachments.  A non-explict option would be to check if a post_type supports either `thumbnail` OR `editor`.

danielbachhuber: > Core doesn't have an explicit way to note if a post supports attachments.

I've opened a new issue so it doesn't block this PR #1029 

> A non-explict option would be to check if a post_type supports either thumbnail OR editor.

Seems a bit prone to breakage. Let's instead also exclude `nav_menu_item` and `revision`, and I think that'll be fine for now.

rachelbaker:
@danielbachhuber #reviewmerge

Issue #1023, opened 2015-03-28
Fix comment schema; return error when response data doesn't match schema


danielbachhuber:
In #955, @rmccue was able to change the response format for a comment without updating the schema. The API shouldn't ever return a response that doesn't match the declared schema.

In this issue, we should:

1. Fix the schema, somewhat contingent on #1022
1. Return a 500 error when the produced response doesn't match the schema, instead of a malformed object.

Note for the second, I'm just referring to fields in the response that are missing or of a different format than in the declared schema. Additional fields can be ignored.

danielbachhuber:
> Return a 500 error when the produced response doesn't match the schema, instead of a malformed object.

I think there are two compelling reasons for this:

1. It's much easier for a client to interpret a 500 error than handle a response formatted differently than expected. In fact, the latter likely means a fatal exception for most clients. By returning an understandable 500 error, we're significantly reducing client developer pain.
1. I can't think of a way a client could successfully handle data of an unexpected format, other than to degrade the user experience in some way. But this would be substantially more code for each client, and the end user should be expected to understand their degraded user experience is because of an API application misconfiguration.

We'd want to use the same request validation code as in #971, so it probably makes sense to create some validator class.

danielbachhuber:
Discussed this some today, and a 203 error seems most appropriate. We'll need to break out the validator code.

danielbachhuber:
The consensus on our call today is that a 203 code isn't appropriate and that there's not much we can do.

PR #1028, opened 2015-03-30
Re-enable the Post endpoint filters

rachelbaker:
These were temporarily commented out in #684. Now that the Comments and Media endpoints have moved over to the new style, we can re-add them.

Based on the request from @kadamwhite in #399 the filters vary on the post_type property.


danielbachhuber:
These filters should only permit _adding_ fields, not removing fields. Removing fields would mean the response deviates from the declared, fixed schema. #1023

danielbachhuber:
Needs a refresh, then it's good to go.

PR #1034, opened 2015-04-01
Add missing response fields to the user schema.

rachelbaker:
The existing schema returned from `wp/users/schema` only included the following properties:

  - id
  - name
  - email
  - link

This adds the following missing properties to the schema:

  - first_name
  - last_name
  - nickname
  - slug
  - url
  - avatar_url
  - description

### DIFF
--- a/lib/endpoints/class-wp-json-users-controller.php
+++ b/lib/endpoints/class-wp-json-users-controller.php
@@ -500,31 +500,61 @@ class WP_JSON_Users_Controller extends WP_JSON_Controller {
 	public function get_item_schema() {
 
 		$schema = array(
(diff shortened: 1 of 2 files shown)


rachelbaker:
There are three additional properties for the `view` context:

  - roles
  - capabilities
  - registered_date

@danielbachhuber How are we noting the different context fields in the schema?

danielbachhuber:
> How are we noting the different context fields in the schema?

TBD #894

I've filed it against b1 so we make a decision on it. I'm open to ideas — https://github.com/WP-API/WP-API/pull/949/files#diff-cba72a4ad6a60d3df6e2bc3284e46a55R557 is my leading candidate right now.

Issue #1035, opened 2015-04-02
Pretty print 5.2-compatible way


mnelson4:
I see #1027 and #835 were on this same topic. I have another pull request for the same feature, but different implementation that IS compatible with PHP 5.2.

danielbachhuber:
I think this makes most sense in a supplemental plugin for now.

mnelson4:
yeah ok although the wpcom API provides this.

If this is to be done by a supplemental plugin we should add a filter for it, or is there one already?

nerrad:
I think it'd be valuable for debugging to have this included.  Especially when this eliminates the sole reason given in #835 for not doing something similar (PHP5.2).

mnelson4:
Also thinking about it: an API client developer might not have access to the server to install another plugin, in which case they're going to need to go to some 3rd party service like @rmccue pointed out on #835. I think given the frequency this is requested merits NOT putting it in another plugin

danielbachhuber:
> I think given the frequency this is requested merits NOT putting it in another plugin

But, it can easily be replicated with a browser extension, no?

mnelson4:
I think the time required to merge this in is much less than the time it will take nearly ALL api clients to figure out how to pretty-up the json in their client (which of course won't always be a browser, or command line; but will often be javascript code, server-side code, etc).
for the event espresso 3 api, we included it because our iOS developer requested it, and it only took a few minutes for me to grab the snippet and implement it, rather than needing to explain all the different ways to pretty-up the JSON in your client to ALL the different developers.

danielbachhuber:
This isn't a priority for us right now.

mnelson4:
Ok so maybe I'll bring it up again in 2.0, at least adding a filter so a plugin can add this

danielbachhuber:
> at least adding a filter so a plugin can add this

You can use the `json_pre_serve_request` filter: https://github.com/WP-API/WP-API/blob/develop/lib/infrastructure/class-wp-json-server.php#L281

mnelson4:
Ya ok that works, and I have a plugin located here: https://github.com/eventespresso/wp-api-pretty-json/blob/master/wp-api-pretty-json.php
However, it seems like that short-circuits a lot more code than what's necessary in this case: my filter would need to basically duplicate the code between lines 283 and 307 (which is difficult because many of those methods are protected), but just pass the final result through the newly added function `prettyJson` method in the pull request. 
[...]

PR #1036, opened 2015-04-02
adds pretty printing json for #1035

mnelson4:
this is how we've been handling pretty printing in the event espresso 3 json api for a few years, and from my testing it works. Just add the GET parameter "pretty=1" and you get back pretty json


danielbachhuber:
We're not taking any new enhancements against master at this time.

PR #1042, opened 2015-04-03
Turn post types from array to object, with name as key

### LINKED ISSUE #934: Index post types by type slug
See https://github.com/WP-API/WP-API/pull/914#issuecomment-76316952

@danielbachhuber:
> Would the key be the json_base for the post type? Should tax and post status endpoints follow suit?

IMO, it should be the type slug itself; you'll typically be indexing by the actual type in the main use case for the post types endpoint (building editor UI).

danielbachhuber:
Fixes #934


rachelbaker:
Merged #1042

Issue #762, opened 2015-01-04
Match WP Core's permission and sticky handling for post_password


rachelbaker:
A new (created) or updated Post with a `post_password` property requires `publish` permission.  We currently only check 'edit' permission.   

A new or updated Post with a `post_password` parameter set cannot be `sticky`

joehoyle:
@rachelbaker why does someone need `publish` to add a password to a post? Is that just a core parity thing?

We currently check `publish` permission on create post, so I think the first condition is already done.

joehoyle:
I don't see any evidence in core that you need publish to add a password, though let me know if I'm wrong there.

joehoyle:
I left the published cap check in - that still needs deciding

rachelbaker:
On the Edit Posts screen, a user has to have publish permission to be able to view and set the post_password.  https://core.trac.wordpress.org/browser/tags/4.1.1/src/wp-admin/includes/meta-boxes.php#L137

Closing this for now.

joehoyle:
Interestingly, I couldn't find the permissions checks in the submission - core security issue maybe?

PR #1044, opened 2015-04-03
Setting a post to be sticky AND password protected should fail

joehoyle:
See #762

### DIFF
--- a/lib/endpoints/class-wp-json-posts-controller.php
+++ b/lib/endpoints/class-wp-json-posts-controller.php
@@ -639,6 +639,20 @@ class WP_JSON_Posts_Controller extends WP_JSON_Controller {
 			if ( ! current_user_can( $post_type->cap->publish_posts ) ) {
 				return new WP_Error( 'json_forbidden', __( 'Sorry, you are not allowed to create password protected posts in this post type' ), array( 'status' => 403 ) );
 			}
+
+			if ( ! empty( $schema['properties']['sticky'] ) && ! empty( $request['sticky'] ) ) {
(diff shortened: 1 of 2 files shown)


lib/endpoints/class-wp-json-posts-controller.php
```diff
@@ -639,6 +639,20 @@ protected function prepare_item_for_database( $request ) {
 			if ( ! current_user_can( $post_type->cap->publish_posts ) ) {
 				return new WP_Error( 'json_forbidden', __( 'Sorry, you are not allowed to create password protected posts in this post type' ), array( 'status' => 403 ) );
 			}
+
+			if ( ! empty( $schema['properties']['sticky'] ) && ! empty( $request['sticky'] ) ) {
```
danielbachhuber: Where is the password conditional in this statement?

Also should check that an already sticky post can't be assigned a password.

danielbachhuber: Erp, nevermind :) missed the rest of the code.

lib/endpoints/class-wp-json-posts-controller.php
```diff
@@ -639,6 +639,20 @@ protected function prepare_item_for_database( $request ) {
...
+				return new WP_Error( 'json_invalid_field', __( 'A post can not be sticky and have a password.' ), array( 'status' => 400 ) );
+			}
+
+			if ( ! empty( $prepared_post->ID ) && is_sticky( $prepared_post->ID ) ) {
```
danielbachhuber: Need WordPress objects here :) `$obj->is_sticky()`

johnbillion: Tell me about it

PR #1048, opened 2015-04-03
Add Embeddable Taxonomy Term Links to the Post Response

### LINKED ISSUE #900: Posts should include `_links` to associated taxonomy terms
The flip side of #898

This will make it possible to include include categories and tags in the Post response, etc.

rachelbaker:
Fixes #900 

The _links in the Post response now include embeddable references for each object taxonomy.


rachelbaker:
In v1 the taxonomy terms were included in the Post response, this moves them to embeddable resource links. Any taxonomy that is returned by `wp_get_object_taxonomies()` is shown, which matches v1 behavior.

lib/endpoints/class-wp-json-terms-controller.php
```diff
@@ -83,12 +84,17 @@ public function get_items( $request ) {
 			return $taxonomy;
 		}
 
-		$terms = get_terms( $taxonomy, $prepared_args );
+		if ( isset( $request['post'] ) ) {
+			$post_id = absint( $request['post'] );
+			$terms = wp_get_object_terms( $post_id, $taxonomy, $prepared_args );
```
danielbachhuber: Shouldn't we validate the taxonomy for the post type?

rachelbaker:
@danielbachhuber  db2c077 adds checks to be sure a post exists and that the taxonomy is related to the post.

lib/endpoints/class-wp-json-terms-controller.php
```diff
@@ -368,4 +379,24 @@ protected function check_valid_taxonomy( $taxonomy ) {
...
+	 */
+	protected function check_valid_taxonomy_for_post( $taxonomy, $post_id ) {
+		$post = get_post( $post_id );
+		if ( empty( $post->ID ) ) {
+			return new WP_Error( 'json_post_invalid_id', __( 'Invalid post ID.' ), array( 'status' => 404 ) );
+		}
+
+		$valid_taxonomies = get_object_taxonomies( $post->post_type );
+		if ( in_array( $taxonomy, $valid_taxonomies ) ) {
+			return $taxonomy;
```
danielbachhuber: Why isn't this just returning `bool`? I don't understand why you'd need `$taxonomy` returned when you're passing it in the first place.

tests/test-json-posts-controller.php
```diff
@@ -103,6 +103,14 @@ public function test_get_item_links() {
...
 		$this->assertEquals( $attachments_url, $links['attachments'][0]['href'] );
+
+		$tags_url = json_url( 'wp/terms/post_tag' );
+		$tags_url = add_query_arg( 'post', $this->post_id, $tags_url );
+		$this->assertEquals( $tags_url, $links['post_tag'][0]['href'] );
+
+		$category_url = json_url( 'wp/terms/category' );
+		$category_url = add_query_arg( 'post', $this->post_id, $category_url );
+		$this->assertEquals( $category_url, $links['category'][0]['href'] );
 	}
```
danielbachhuber: We should have a test for a custom taxonomy too.

lib/endpoints/class-wp-json-posts-controller.php
```diff
@@ -1133,6 +1133,19 @@ protected function prepare_links( $post ) {
 			);
 		}
 
+		$taxonomies = get_object_taxonomies( $post->post_type );
+		if ( ! empty( $taxonomies ) ) {
+			foreach ( $taxonomies as $tax ) {
```
danielbachhuber: Need to check the taxonomy isn't private. This would benefit from a test.

danielbachhuber:
Just a few nits.

lib/endpoints/class-wp-json-posts-controller.php
rachelbaker: WordPress doesn't have private taxonomies: https://core.trac.wordpress.org/ticket/21949

danielbachhuber: We still need to check. Otherwise we will be exposing private data.

rachelbaker:
I found a big hole in my logic here, I also need to check that a user has permission to read a post before we return the terms associated with it. Still a WIP.

lib/endpoints/class-wp-json-posts-controller.php
rachelbaker: Added in line 1156

rachelbaker:
Includes:
- checks to make sure a user can read a post before returning terms related to a given post_id
- checks to make sure a taxonomy is public
- added permission_callback methods.
- unit tests
- wraps responses in json_ensure_response

lib/endpoints/class-wp-json-terms-controller.php
```diff
@@ -365,18 +434,53 @@ public function get_item_schema() {
...
+
+		return sanitize_key( $taxonomy );
 	}
 
+	/**
+	 * Check that the taxonomy is valid for a given post type.
+	 *
+	 * @param string  $taxonomy
+	 * @param integer $post_id
+	 * @return null|WP_Error
```
danielbachhuber: Why return `null`?

tests/test-json-posts-controller.php
```diff
@@ -105,6 +105,14 @@ public function test_get_item_links() {
 		$attachments_url = json_url( 'wp/media' );
 		$attachments_url = add_query_arg( 'post_parent', $this->post_id, $attachments_url );
 		$this->assertEquals( $attachments_url, $links['attachments'][0]['href'] );
+
+		$tags_url = json_url( 'wp/terms/post_tag' );
```
danielbachhuber: Why is this `post_tag`, and it's `tag` further down?

lib/endpoints/class-wp-json-terms-controller.php
rachelbaker: It just returns....what would you do?

danielbachhuber: True?

rachelbaker:
@danielbachhuber Both pieces of recent feedback have been addressed in `82ac7d2` and `de7b818`

Issue #839, opened 2015-02-05
Expose basic details for users with published posts


danielbachhuber:
`GET /wp/users` currently requires a user to have `list_users` cap. We should expose all users who have any published content for public post types.

Core doesn't have a sane way to query this, me thinks :(

#838 implements this for `GET /wp/users/<id>` for published posts, but we should support all post types.

TimothyBJacobs:
I think something like this could work, basic and untested.

````php
$types = get_post_types( array( 'public' => true ) );
$types = implode( $types, ', ' );
````
and the sql.
````sql
SELECT `post_author` FROM `wp_posts` WHERE `post_type` IN ( $types );
````

danielbachhuber:
Another consideration: performance. Our queries should be behind some basic object cache so we don't make the same query over and over

rmccue:
> Core doesn't have a sane way to query this, me thinks :(

Can we use `get_users( [ 'who' => 'authors' ] )` ? Not sure if that's robust enough.

rmccue:
> Not sure if that's robust enough.

Ah, checking the source reveals that it's any user with a user level > 0, so could be too much.

rachelbaker:
Punting to Beta 2.

joehoyle:
Currently we allow `/users/<id>` for any author that has a published post of post_type = post. I'd recommend we use `get_posts_by_author_sql` to get those users and roll with that.

rmccue:
Our issue is that we don't really want to do SQL queries in the API, since it doesn't belong there. What we need is a way for `WP_User_Query` to use `get_posts_by_author_sql` for the query rather than just allowing sorting by it. This also means that it works nicely with `wp_is_large_network`, any caching layers that may be in place with user query filters, etc.

joehoyle:
Trac ticket: https://core.trac.wordpress.org/ticket/32250#comment:8 looks to be making some progress.

rachelbaker:
Trac Ticket is making progress, but punting to Beta 3 to see how this pans out.

johnbillion:
https://core.trac.wordpress.org/ticket/32250 is in.

joehoyle:
Will handle a PR for this by tomorrow, so leaving in b3 for now.

PR #1049, opened 2015-04-03
Use schema abstraction to limit which user fields are exposed per context

danielbachhuber:
See #894

### DIFF
--- a/tests/test-json-users-controller.php
+++ b/tests/test-json-users-controller.php
@@ -38,6 +38,7 @@ class WP_Test_JSON_Users_Controller extends WP_Test_JSON_Controller_Testcase {
 		wp_set_current_user( $this->user );
 
 		$request = new WP_JSON_Request( 'GET', '/wp/users' );
+		$request->set_param( 'context', 'view' );
 		$response = $this->server->dispatch( $request );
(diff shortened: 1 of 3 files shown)


danielbachhuber:
@joehoyle @rachelbaker #reviewmerge

joehoyle:
Looks good to me

tests/test-json-users-controller.php
```diff
@@ -38,6 +38,7 @@ public function test_get_items() {
 		wp_set_current_user( $this->user );
 
 		$request = new WP_JSON_Request( 'GET', '/wp/users' );
+		$request->set_param( 'context', 'view' );
```
joehoyle: @danielbachhuber do we have tests for the other contexts too?

danielbachhuber: `GET /wp/users?context=embed` would actually error right now because of #839

Why do you think we should have tests for other contexts?

joehoyle: > Why do you think we should have tests for other contexts?

I think we should have tests for as much things as possible. Do we not need to test both `/wp/users?context=view` and `/wp/users?context=edit`?

rachelbaker:
Merged #1049

PR #1053, opened 2015-04-03
Use full Post schema to filter fields based on context

danielbachhuber:
See #894


lib/endpoints/class-wp-json-attachments-controller.php
```diff
@@ -167,6 +167,9 @@ public function prepare_item_for_response( $post, $request ) {
 		    $response['media_details']['sizes'] = new stdClass;
 		}
 
+		$context = ! empty( $request['context'] ) ? $request['context'] : 'view';
```
joehoyle: @danielbachhuber can this be handled by a default param in the `args` in the route registration?

lib/endpoints/class-wp-json-posts-controller.php
```diff
@@ -964,60 +964,53 @@ public function prepare_item_for_response( $post, $request ) {
...
+
 			$data['content'] = array(
+				'raw'      => $post->post_content,
 				'rendered' => apply_filters( 'the_content', $post->post_content ),
 			);
-			if ( 'edit' === $request['context'] ) {
-				$data['content']['raw'] = $post->post_content;
+
+			if ( ! empty( $post->post_password ) ) {
+				$_COOKIE['wp-postpass_' . COOKIEHASH] = '';
```
joehoyle: Can we get a comment in here to explain this crazyness?

lib/endpoints/class-wp-json-posts-controller.php
```diff
@@ -1080,6 +1073,9 @@ public function prepare_item_for_response( $post, $request ) {
 			) );
 		}
 
+		$context = ! empty( $request['context'] ) ? $request['context'] : 'view';
```
joehoyle: Again, seems like this should be a registered arg with a default value?

lib/endpoints/class-wp-json-attachments-controller.php
danielbachhuber: > can this be handled by a default param in the args in the route registration?

Maybe. But, conceivably, another controller can call this directly and omit `context`

joehoyle: Fair enough

lib/endpoints/class-wp-json-posts-controller.php
```diff
@@ -1198,24 +1191,49 @@ public function get_item_schema() {
...
+					'description' => 'A password to protect access to the post.',
+					'type'        => 'string',
+					'context'     => array( 'edit' ),
+				),
+				'slug'            => array(
+					'description' => 'An alphanumeric identifier for the object unique to its type.',
+					'type'        => 'string',
+					'context'     => array( 'view', 'edit' ),
+				),
+				'status'          => array(
```
joehoyle: an we pass enums here?

lib/endpoints/class-wp-json-posts-controller.php
danielbachhuber: 8d53589

lib/endpoints/class-wp-json-posts-controller.php
danielbachhuber: dd05e31

PR #1054, opened 2015-04-03
Move the permissions checks for password and author into the permissions callback

joehoyle:


joehoyle:
@WP-API/amigos #reviewmerge

PR #1059, opened 2015-04-05
Allow WP_JSON_Server::send_header()/send_headers() to be accessed publicly

johnbillion:
When using the `json_pre_serve_request` hook, it's not possible to call `$server->send_header()` or `$server->send_headers()` because they're protected methods.

Let's make them public so a callback on the `json_pre_serve_request` action can use these methods instead of using `header()` directly.

### DIFF
--- a/lib/infrastructure/class-wp-json-server.php
+++ b/lib/infrastructure/class-wp-json-server.php
@@ -727,7 +727,7 @@ class WP_JSON_Server {
 	 * @param string $key Header key
 	 * @param string $value Header value
 	 */
-	protected function send_header( $key, $value ) {
+	public function send_header( $key, $value ) {
 		// Sanitize as per RFC2616 (Section 4.2):
 		//   Any LWS that occurs between field-content MAY be replaced with a
 		//   single SP before interpreting the field value or forwarding the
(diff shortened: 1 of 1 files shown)


rachelbaker:
@joehoyle From what I saw, the `json_pre_serve_request` hook is too late to use the Response object.  The headers are already sent from the Response object to the Server a few lines above: https://github.com/WP-API/WP-API/blob/develop/lib/infrastructure/class-wp-json-server.php#L263

Looking at this request again, I see the `json_post_dispatch` filter is a better location for setting or removing headers in the Response object.  

What do you think (also would be nice for @rmccue to chime in here)?  Should we keep these methods protected in the Server class?

rmccue:
> What do you think (also would be nice for @rmccue to chime in here)? Should we keep these methods protected in the Server class?

I'd say yes (they should remain protected); if you want to send headers, do it in `json_post_dispatch`. `json_pre_serve_request` is intended for you to handle output for the body directly.

@johnbillion Was there a specific use case you needed here?

johnbillion:
I can indeed use the `json_post_dispatch` filter, so feel free to revert this.

Issue #1063, opened 2015-04-06
Evaluate all protected controller methods


danielbachhuber:
We have a variety of protected controller methods. Once we release WP-API to the wild, we're committing to keeping those methods around forever.

Before we do, we should some methods as private so they aren't subclassed, and abstract others out to dedicated functions so they can be used without controllers. `check_is_post_type_allowed()` is one candidate.

Related #1062

danielbachhuber:
I'll own this in b3

rachelbaker:
Punted to Beta 4

danielbachhuber:
Because we don't want to globally make methods private, we need to evaluate our existing protected methods one by one. Once we've done so, this issue can be closed.

BE-Webdesign:
Is there a list of these that need to be discussed about?

danielbachhuber:
> Is there a list of these that need to be discussed about?

Nope, but I wrote a script to produce a list :) https://gist.github.com/danielbachhuber/5fb19ff57ba273348c17

```
WP_REST_Controller
+-----------------------------------+-------------------------------------+---------+
| public                            | protected                           | private |
+-----------------------------------+-------------------------------------+---------+
| register_routes                   | prepare_item_for_database           |         |
| get_items_permissions_check       | add_additional_fields_to_object     |         |
| get_items                         | update_additional_fields_for_object |         |
| get_item_permissions_check        | add_additional_fields_schema        |         |
| get_item                          | get_additional_fields               |         |
| create_item_permissions_check     | get_object_type                     |         |
| create_item                       |                                     |         |
| update_item_permissions_check     |                                     |         |
| update_item                       |                                     |         |
| delete_item_permissions_check     |                                     |         |
| delete_item                       |                                     |         |
| prepare_item_for_response         |                                     |         |
| prepare_response_for_collection   |                                     |         |
| filter_response_by_context        |                                     |         |
| get_item_schema                   |                                     |         |
| get_public_item_schema            |                                     |         |
| get_collection_params             |                                     |         |
| get_context_param                 |                                     |         |
| get_endpoint_args_for_item_schema |                                     |         |
+-----------------------------------+-------------------------------------+---------+
WP_REST_Posts_Controller
+------------------------------------------+----------------------------+---------+
| public                                   | protected                  | private |
+------------------------------------------+----------------------------+---------+
| __construct                              | prepare_items_query        |         |
| register_routes                          | get_allowed_query_vars     |         |
[...]

PR #1064, opened 2015-04-06
Remove unnecessary sticky posts abstraction

danielbachhuber:
protected methods are committments we don't want to make lightly.

See #1063

### DIFF
--- a/lib/endpoints/class-wp-json-posts-controller.php
+++ b/lib/endpoints/class-wp-json-posts-controller.php
@@ -219,8 +219,11 @@ class WP_JSON_Posts_Controller extends WP_JSON_Controller {
 		$schema = $this->get_item_schema();
 
 		if ( ! empty( $schema['properties']['sticky'] ) ) {
-			$sticky = isset( $request['sticky'] ) ? (bool) $request['sticky'] : false;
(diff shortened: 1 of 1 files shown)


danielbachhuber:
@WP-API/amigos #reviewmerge

lib/endpoints/class-wp-json-posts-controller.php
```diff
@@ -293,8 +296,11 @@ public function update_item( $request ) {
 		}
 
 		if ( ! empty( $schema['properties']['sticky'] ) && isset( $request['sticky'] ) ) {
-			$sticky = isset( $request['sticky'] ) ? (bool) $request['sticky'] : false;
-			$this->handle_sticky_posts( $sticky, $post_id );
+			if ( ! empty( $request['sticky'] ) ) {
```
joehoyle: @danielbachhuber does this mean if you don't pass `sticky` _at all_ in an update request, it will remove `sticky` from it? (this might have been the behaviour before too - which is incorrect).

danielbachhuber: There's an `isset()` check in the line before.

joehoyle: my bad!

PR #1065, opened 2015-04-06
Allow empty titles to be set after initial title is set.

tlovett1:
Once a title has been set as non-empty further edits cannot set the title as empty. This fixes the problem with an `isset` check instead of `! empty`.


danielbachhuber:
@tlovett1 Can you add a test for this please?

tlovett1:
Done :)

PR #1066, opened 2015-04-07
Fix the format of the args when building them from the Schema

joehoyle:
Currently we pass the args as `array( 'title', 'author', ... );` which is not
the correct format. This Pull Request changes that to be `array( 'title' =>
array( ... ), ... )` which can / will be extended to add arg validation per
the enums / datatype defines in the schema.

### DIFF
--- a/lib/endpoints/class-wp-json-controller.php
+++ b/lib/endpoints/class-wp-json-controller.php
@@ -103,4 +103,21 @@ abstract class WP_JSON_Controller {
 		return array();
 	}
 
+	/**
+	 * Get an array of endpoint arguments from the item schema for the controller.
+	 * 
+	 * @return array
+	 */
+	public function get_endpoint_args_for_item_schema() {
+
+		$schema                = $this->get_item_schema();
+		$post_type_fields      = ! empty( $schema['properties'] ) ? $schema['properties'] : array();
+		$post_type_fields_args = array();
+
+		foreach ( $post_type_fields as $field_id => $params ) {
+			$post_type_fields_args[$field_id] = array();
+		}
+
+		return $post_type_fields_args;
+	}
 }
--- a/lib/endpoints/class-wp-json-posts-controller.php
+++ b/lib/endpoints/class-wp-json-posts-controller.php
@@ -15,8 +15,7 @@ class WP_JSON_Posts_Controller extends WP_JSON_Controller {
 
 		$base = $this->get_post_type_base( $this->post_type );
 
-		$schema = $this->get_item_schema();
-		$post_type_fields = ! empty( $schema['properties'] ) ? array_keys( $schema['properties'] ) : array();
+		$post_type_fields = $this->get_endpoint_args_for_item_schema();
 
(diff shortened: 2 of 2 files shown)


joehoyle:
@WP-API/amigos #reviewmerge

lib/endpoints/class-wp-json-controller.php
```diff
@@ -103,4 +103,21 @@ public function get_item_schema() {
...
+	 * @return array
+	 */
+	public function get_endpoint_args_for_item_schema() {
+
+		$schema                = $this->get_item_schema();
+		$post_type_fields      = ! empty( $schema['properties'] ) ? $schema['properties'] : array();
+		$post_type_fields_args = array();
+
+		foreach ( $post_type_fields as $field_id => $params ) {
+			$post_type_fields_args[$field_id] = array();
```
danielbachhuber: Huh?

lib/endpoints/class-wp-json-controller.php
```diff
@@ -103,4 +103,21 @@ public function get_item_schema() {
 		return array();
 	}
 
+	/**
+	 * Get an array of endpoint arguments from the item schema for the controller.
+	 * 
+	 * @return array
+	 */
+	public function get_endpoint_args_for_item_schema() {
```
danielbachhuber: Why the method abstraction? Does this code need to be reusable?

joehoyle: Currently posts is the only one that is using args from the schema so far I believe, but this is in the base controller so all others can user it (so that's at least 3 times per object type). Also, from my commit message: This Pull Request changes that to be `array( 'title' => array( ... ), ... )` which can / will be extended to add arg validation per the enums / datatype defines in the schema.

lib/endpoints/class-wp-json-controller.php
joehoyle: This array is currently empty as we are not specifying an options for the args. Does this not make sense?

danielbachhuber: > Does this not make sense?

Nope — validation is supposed to be happening :) #971

Something must be broken.

joehoyle: I don't understand what you are referring to. Validation works off the `args` param - we are now taking that from the schema; which prior to this PR was specifying the args incorrectly. I don't see how moving the validation per #971 is relevant.

danielbachhuber: > I don't understand what you are referring to. Validation works off the args param - we are now taking that from the schema; which prior to this PR was specifying the args incorrectly.

Right — so there should've been some tests failing because of this break.

joehoyle: I don't think we have any tests specifically testing that - none of the `args` have any options currently, so no `required` flags to break. Also, I believe if a value is in `$_POST`, it still gets handed to the callback as an arg, even though it wasn't specified in the `args` options.

PR #1089, opened 2015-04-13
Check edit permission for context before returning created post

miguelpeixe:
After creating the post, the returned post object is always on "edit" context. In my project, the user can publish a post in a certain post type, but not edit afterwards. That causes a crash, since the returned object for `json_ensure_response( $this->get_post( $result, 'edit' ) )` is a WP_Error and it tries to `set_status` on this WP_Error.

The best solution, I figured, is to check the user permission to send the proper context as parameter.

### DIFF
--- a/lib/class-wp-json-posts.php
+++ b/lib/class-wp-json-posts.php
@@ -234,7 +234,7 @@ class WP_JSON_Posts {
 
 		$context = 'edit';
 
-		if ( !json_check_post_permission( $post, 'edit' ) ) {
+		if ( !json_check_post_permission( get_post( $result, ARRAY_A ), 'edit' ) ) {
 			$context = 'view';
 		}
 
--- a/lib/class-wp-json-posts.php
+++ b/lib/class-wp-json-posts.php
@@ -232,7 +232,13 @@ class WP_JSON_Posts {
 			return $result;
 		}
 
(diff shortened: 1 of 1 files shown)


danielbachhuber:
>  In my project, the user can publish a post in a certain post type, but not edit afterwards

Just for the record, could you explain this use case in a bit more detail?

> The best solution, I figured, is to check the user permission to send the proper context as parameter.

Can you add some test coverage for this?

miguelpeixe:
My project is a collaborative platform for mapping and reviewing bike-friendly city spots. "Place" is one of the post types. A place can be registered by any contributor, using OSM database to fill most of the fields. After a place is registered, only the administrators can edit it (fix house number, etc). That means after a place is published, the contributor does not have permission to edit it.

You can see [here](https://github.com/codigourbano/bikeit/blob/2f71271845e2c8da985818d8b5f78aca2953c75c/inc/place.php#L82-L158) that I'm using meta capabilities to get that working.

I'm performing the tests now, soon I'll have it done

danielbachhuber:
This isn't a problem on `develop`, which is what we're more actively maintaining. On `develop`, a `POST /wp/v2/posts` will return all of the post data in edit context, which is reasonable because it will only be data the user has supplied in the Create request.

PR #1091, opened 2015-04-14
Fix error with OPTIONS requests

### LINKED ISSUE #1057: OPTIONS requests are broken
`Notice:  Undefined offset: 1 in json-rest-api/plugin.php on line 523`

`json_handle_options_request()` triggers the above PHP notice on an OPTIONS request due to [this line](https://github.com/WP-API/WP-API/blob/e61ce72688268bd4df63b7d7233cdb1166f7855c/plugin.php#L523) looking for an array key that doesn't exist.

This breaks OPTIONS requests completely.

rmccue:
Fixes #1057 and adds testing to ensure it doesn't happen again.

### DIFF
--- a/plugin.php
+++ b/plugin.php
@@ -507,10 +507,6 @@ function json_handle_options_request( $response, $handler, $request ) {
 
 	$accept = array();
 
-	$handler_class = get_class( $handler );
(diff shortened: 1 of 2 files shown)


rmccue:
@WP-API/amigos #reviewmerge

Issue #250, opened 2014-05-29
Filter params not working?


lapidus:
I am trying to filter /posts with for example ?filter[offset]=2 but nothing seems to happen. Is this the right syntax?

imjared:
Just got so hung up on this. `filter[page]=number` does NOT seem to work nor does `paged` or anything else. 

Simply passing `?page=3` saved me.

`/wp-json/posts?filter[order]=DESC&filter[posts_per_page]=2&page=3`. Seems like there are several issues that point to this functionality as confusing so hopefully the documentation improves over time. In the meantime, hopefully some poor soul unlike me finds this issue through the Google.

tomconroy:
Thank you @imjared ! :raised_hands:

jcutrell:
Question on this thread - I'm wondering if support was ever added for allowing a specific offset, rather than simply using the page param? My page starts with 2 items, and I would like to load in 6 when the user clicks a button. Doing so requires an offset, or some strange workarounds.

pbrisson:
+1 jcutrell question! I'm in the same situation. Support for the offset variable should be properly implemented.

pbrisson:
@jcutrell You need to be [authenticated](http://wp-api.org/guides/authentication.html) to use offset, as stated in [WP API documentation](http://wp-api.org/#posts_retrieve-posts):

> In addition, the following are available when authenticated as a user with edit_posts:
> * offset

PR #1095, opened 2015-04-15
Add note in routes.md about `page` parameter

danielma:
`?filter[paged]=2` does not work for pagination (although according to the current docs, it should), `?page=2` does work

I was only able to find this info in #250 even though it should be available in the main docs


danielbachhuber:
We are actively maintaining v2.wp-api.org documentation at this time, which corresponds to the `develop` branch.

danielma:
that may be the case, but until that's considered stable, I think the V1 documentation should be kept accurate for the people trying to use it

PR #1098, opened 2015-04-15
Correct a bunch of filter docs in WP_JSON_Server

johnbillion:
The `json_pre_dispatch` and `json_post_dispatch` filters both have missing docs.

The `$result` parameter in the `json_post_dispatch` and `json_pre_serve_request` filters can be any object that implements `WP_HTTP_ResponseInterface`, not just a `WP_JSON_Response`, so that should be used as the parameter type in the docblock.


rachelbaker:
Merged #1098 

Thanks again @johnbillion

PR #1103, opened 2015-04-17
Prefer destructured objects over nested sub-objects

### LINKED ISSUE #1022: Establish criteria for when data should be nested in sub-objects on resources
I'd still like to keep an issue open on #955 for discussion. The pull request changes:

```
{
    "author": "admin",
    "author_email": "admin@example.com",
    "author_url": ""
}
```

Into:

```
{
    "author": {
        "name": "admin",
        "email": "admin@example.com",
        "url": ""
    } 
}
```

I'd argue for the former, as it's less code to write for create and update requests, and the latter implies some form of consistent structured object. @rmccue [argues for the latter because it's in v1](https://wordpress.slack.com/archives/core-restapi/p1427325812000068), although it's not clear why it made into v1 in the first place. According to @rmccue, [Backbone is one place where destructured data is better](https://wordpress.slack.com/archives/core-restapi/p1427325682000055).

More broadly speaking, though, we should establish some criteria for when we put data in sub-objects on a resource, and when we don't. If we don't have such criteria, then we will introduce inconsistencies to our API. Inconsistent behaviors, arguments, and formats are one reason developers don't like working with WordPress (and PHP). WP-API provides us the opportunity to provide a new, consistent, durable interface to WordPress, let it be through JSON REST, PHP, CLI, or other usage implementations.

#848 would be impacted by this criteria.

danielbachhuber:
Converts Comment author object to flat Comment attributes

Also changes response object based on declared context

Fixes #1022
See #894

### DIFF
--- a/lib/endpoints/class-wp-json-comments-controller.php
+++ b/lib/endpoints/class-wp-json-comments-controller.php
@@ -414,7 +414,7 @@ class WP_JSON_Comments_Controller extends WP_JSON_Controller {
 				'rendered'     => apply_filters( 'comment_text', $comment->comment_content, $comment ),
 				'raw'          => $comment->comment_content,
 			),
-			'karma'        => $comment->comment_karma,
+			'karma'        => (int) $comment->comment_karma,
 			'link'         => get_comment_link( $comment ),
 			'status'       => $this->prepare_status_response( $comment->comment_approved ),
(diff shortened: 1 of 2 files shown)


danielbachhuber:
@WP-API/amigos #reviewmerge

PR #1105, opened 2015-04-17
Rename the `wp_json_server_before_serve` to `wp_json_init`

joehoyle:
Currently, developers must use the `wp_json_server_before_serve` hook to add
any new endpoints. This is very verbose and difficult to remember, I think it
would be better to have this named akin to `init` and `admin_init` hooks. In
the future, if the plugin switches to `wp_api` prefix, rather than `json` I
think we'd change this hook to `api_init` but for now I think it's best kept
inline with the rest of the naming conventions.

### DIFF
--- a/plugin.php
+++ b/plugin.php
@@ -155,7 +155,7 @@ function create_initial_json_routes() {
 	$controller->register_routes();
 
 }
-add_action( 'wp_json_server_before_serve', 'create_initial_json_routes', 0 );
+add_action( 'wp_json_init', 'create_initial_json_routes', 0 );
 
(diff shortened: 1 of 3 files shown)


joehoyle:
@WP-API/amigos any thoughts?

danielbachhuber:
:+1:

danielbachhuber:
Do we need a compatibility shim though?

rmccue:
:+1:

> Do we need a compatibility shim though?

Nope, 2.0 isn't compatible with anything, so not sure why we would.

PR #1106, opened 2015-04-17
`author` is the Comment attribute with user ID

danielbachhuber:


danielbachhuber:
@WP-API/amigos #reviewmerge

PR #1107, opened 2015-04-17
Only use if it's not WP_Error, and get data from Response object

### LINKED ISSUE #1099: Post thumbnail error
When sending PUT request to /wp-json/posts/{{post id}} with {"featured_image": {{media_id}} }, API response is "Internal server error 500". The problem is that wp-content/plugins/json-rest-api/lib/class-wp-json-media.php on line 437 assumes that $thumbnail is a PHP array, but it is an instance of WP_JSON_Response.

error log details are: 
PHP Deprecated:  Automatically populating $HTTP_RAW_POST_DATA is deprecated and will be removed in a future version. To avoid this warning set 'always_populate_raw_post_data' to '-1' in php.ini and use the php://input stream instead. in Unknown on line 0
PHP Fatal error:  Cannot use object of type WP_JSON_Response as array in /Applications/MAMP/htdocs/wordpress/wp-content/plugins/json-rest-api/lib/class-wp-json-media.php on line 437

My plugin version is: '1.2.1'
PHP version is: 5.6.2
Apache: 2.2.29

danielbachhuber:
Fixes #1099

### DIFF
--- a/lib/class-wp-json-media.php
+++ b/lib/class-wp-json-media.php
@@ -433,8 +433,10 @@ class WP_JSON_Media extends WP_JSON_Posts {
 		if ( ! empty( $data['featured_image'] ) ) {
 			// Already verified in preinsert_check()
 			$thumbnail = $this->get_post( $data['featured_image'], 'child' );
-
-			set_post_thumbnail( $post['ID'], $thumbnail->data['ID'] );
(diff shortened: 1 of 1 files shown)


rachelbaker:
Merged #1107

Issue #855, opened 2015-02-08
Conditionally register revisions endpoint


danielbachhuber:
Revisions endpoint should only be registered if post type supports revisions.

rachelbaker:
We should make sure this has tests too, so it doesn't happen again.

danielbachhuber:
Related #869

danielbachhuber:
:hammer: :hammer: :hammer:

PR #1110, opened 2015-04-17
Abstract revisions to dedicated controller; only include revisioned fields

### LINKED ISSUE #869: Create WP_JSON_Revisions_Controller
Because the schema for Revisions will be quite different than Posts and Pages, we should just create a separate controller for it.

This way, we don't need to register `/wp/posts/<id>/revisions` and can instead use `/wp/revisions/?parent=<id>`

After #844

danielbachhuber:
Fixes #869, #855 
See #894, #314

### DIFF
--- a/lib/endpoints/class-wp-json-revisions-controller.php
+++ b/lib/endpoints/class-wp-json-revisions-controller.php
@@ -60,6 +60,12 @@ class WP_JSON_Revisions_Controller extends WP_JSON_Controller {
 	 * @return mixed WP_Error or WP_JSON_Response.
 	 */
 	public function get_items( $request ) {
+
+		$parent = get_post( $request['parent_id'] );
+		if ( ! $request['parent_id'] || ! $parent || $this->parent_post_type !== $parent->post_type ) {
+			return new WP_Error( 'json_post_invalid_parent_id', __( 'Invalid post parent ID.' ), array( 'status' => 404 ) );
+		}
+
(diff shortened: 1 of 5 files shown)


danielbachhuber:
@WP-API/amigos #reviewmerge

lib/endpoints/class-wp-json-revisions-controller.php
```diff
@@ -0,0 +1,318 @@
...
+	public function get_item_permissions_check( $request ) {
+
+		$response = $this->get_items_permissions_check( $request );
+		if ( ! $response || is_wp_error( $response ) ) {
+			return $response;
+		}
+
+		$post = get_post( $request['id'] );
+		if ( ! $post || 'revision' !== $post->post_type ) {
+			return new WP_Error( 'json_post_invalid_id', __( 'Invalid revision ID.' ), array( 'status' => 404 ) );
```
joehoyle: Elsewhere we have not thrown invalid errors in the permissions checks. We'd typically not return any error in the case of the post not existing, or being the wrong post type - and handle that check in the `get_item` callback it's self.

joehoyle:
Just a concern around the error checking done in the permissions callbacks, as I'm not sure if that is the best place for those checks. Apart from that - I'm happy.

lib/endpoints/class-wp-json-revisions-controller.php
danielbachhuber: d3ff8cd

PR #1112, opened 2015-04-17
When `context=edit`, confirm user can `manage_comments`

danielbachhuber:
Otherwise supplying `context=edit` grants unauthorized access to comment secrets


danielbachhuber:
@WP-API/amigos #reviewmerge

PR #1128, opened 2015-04-23
Automatically validate params based of the Schema data

joehoyle:
The basic idea here is to provide to "for free" validation based off the item schema. Looking at enums and data types to start with.


joehoyle:
@WP-API/amigos #reviewmerge

danielbachhuber:
Does it return all potential failures or just the first?

joehoyle:
@danielbachhuber The implementation of the `validate_callback` is to fail on the first, see https://github.com/WP-API/WP-API/pull/989/files#diff-887d62e7b2ec4bf35ee08d99831b5b4bR681

We could change that and bulk them all or multiple - though that should be a PR on top of https://github.com/WP-API/WP-API/pull/989/files#diff-887d62e7b2ec4bf35ee08d99831b5b4bR681 I think.

I think it could be a good idea - the only downside is slight performance hit I suppose.

danielbachhuber:
I think it makes sense to validate all at once. Consider submitting a form. If you did them one at a time, then the user would only get one error at a time and potentially hit 10+ error states before successfully submitting. 

If it's too much of a lift, we can come back to it later. At the very least though, it would be nice if the response could support one or more error messages.

joehoyle:
As the response is a `WP_Error` I think multiple error is already supported from the server part.

joehoyle:
Created https://github.com/WP-API/WP-API/pull/1131 to address multiple errors at once.

joehoyle:
I also added the same for sanitizing defaults for the schema to this PR

lib/endpoints/class-wp-json-controller.php
```diff
@@ -196,4 +200,111 @@ public function get_endpoint_args_for_item_schema( $add_required_flag = true ) {
...
+						return sanitize_email( $value );
+					}
+					return sanitize_text_field( $value );
+
+				case 'uri' :
+					return esc_url_raw( $value );
+			}
+		}
+
+		if ( $property['type'] === 'string' ) {
```
danielbachhuber: What about strings with HTML in them?

lib/endpoints/class-wp-json-controller.php
```diff
@@ -196,4 +200,111 @@ public function get_endpoint_args_for_item_schema( $add_required_flag = true ) {
...
+
+	/**
+	 * Validate an parameter value that's based on a property from the item schema.
+	 * 
+	 * @param  mixed $value
+	 * @param  WP_JSON_Request $request
+	 * @param  string $parameter
+	 * @return WP_Error|bool
+	 */
+	public function validate_schema_property( $value, $request, $parameter ) {
```
danielbachhuber: It would be good to have a strong set of tests for this `validate_schema_property()` and the `sanitize_schema_property()` methods, independently of how they're used in our controllers.

lib/endpoints/class-wp-json-controller.php
joehoyle: Good point, will remove this default.

lib/endpoints/class-wp-json-controller.php
joehoyle: agreed

rachelbaker:
I like the approach here, but I don't think we can introduce another enhancement this late into Beta 1.  Moving to Beta 2

lib/endpoints/class-wp-json-controller.php
joehoyle: Fixed in b344f5b

lib/endpoints/class-wp-json-controller.php
joehoyle: Added in ca7d320

joehoyle:
@WP-API/amigos ready for a #reviewmerge again

danielbachhuber:
Looks good to me. I'll wait for an assist from Ryan or Rachel.

`WP_REST_Controller` is getting a bit hairy though. I think it's time for more abstraction.

joehoyle:
@rmccue @rachelbaker fancy merging / any other thoughts? :)

rachelbaker:
@joehoyle is going to resolve the merge conflicts so this can be merged.

rachelbaker:
Merged #1128

PR #1129, opened 2015-04-23
Sanitize args using new args API

joehoyle:
This completes the sanitizing outlined in https://github.com/WP-API/WP-API/pull/989


lib/endpoints/class-wp-json-posts-controller.php
```diff
@@ -25,8 +25,10 @@ public function register_routes() {
 					'context'          => array(
 						'default'      => 'view',
 					),
-					'type'            => array(),
```
joehoyle: Removed `type` as it's not used in the controller at all - this was from before the posts controller abstraction.

joehoyle:
@WP-API/amigos #reviewmerge

PR #1131, opened 2015-04-23
Show all the invalid param errors at once

joehoyle:
This means clients can get lots of info at once, rather than one at a time,
see https://github.com/WP-API/WP-API/pull/1128#issuecomment-95405670


joehoyle:
@WP-API/amigos #reviewmerge

PR #1132, opened 2015-04-23
Use the `required` flags from the schema for CREATE post

joehoyle:
We have to use a switch on the getting args from schema, as the `required`
flags in schemas should be ignored for update requests


joehoyle:
@WP-API/amigos #reviewmerge

PR #1133, opened 2015-04-24
Readonly attribute in schema to exclude from args array

joehoyle:
Picking up from #892 - now the endpoint args are specified from the schema, it pays to declare the readonly attribute, as these are ignored in the route's arg registration.


joehoyle:
@WP-API/amigos #reviewmerge

PR #1142, opened 2015-04-27
Only return 201 on Create. Update should be 200

### LINKED ISSUE #1141: Only return 201 on creation
Updating posts [returns a 201 right now](https://github.com/WP-API/WP-API/blob/e24264fa965c3f012f9c43a272768d27120b01f6/lib/endpoints/class-wp-json-posts-controller.php#L283), which is incorrect.

The point of a `201 Created` is that it requires a `Location` header as well, to indicate where the new resource lives. This is because you `POST /wp/posts`, but end up with a new resource at `/wp/posts/42` instead.

We need to remove the Location and 201 from updating.

danielbachhuber:
Fixes #1141

### DIFF
--- a/lib/endpoints/class-wp-json-posts-controller.php
+++ b/lib/endpoints/class-wp-json-posts-controller.php
@@ -275,14 +275,10 @@ class WP_JSON_Posts_Controller extends WP_JSON_Controller {
 		 * do_action( 'json_insert_post', $post, $request );
 		 */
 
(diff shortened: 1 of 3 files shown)


danielbachhuber:
@WP-API/amigos #reviewmerge

rachelbaker:
This was my fault, thanks for the fix @danielbachhuber

Issue #1146, opened 2015-04-27
Bring back descriptive errors for permissions failures


joehoyle:
When we merged the permissions callbacks, we lost some of the finer grained error codes and messages. We should typically have each `false` return from a permissions callback actually returning a `WP_Error` object.

rmccue:
Essentially: `json_forbidden` doesn't tell us much, since it's really just duplicating the 403 that you get already. The code is instead meant to be a machine-readable version of the plain-text message you get to allow checking fine-grained errors in clients.

(You can still do coarse checks via HTTP status codes, of course.)

joehoyle:
PR merged, closing

PR #1148, opened 2015-04-27
Specific error codes for permissions failures

joehoyle:
See #1146


joehoyle:
Expecting to fail until merging @rachelbaker's meta PR in

Issue #1119, opened 2015-04-19
Version our Endpoints and Filters


rachelbaker:
As we prepare to release version v2.0 Beta 1 we still need to add versioning to the core endpoints.  Also, as noted by @danielbachhuber we may want to consider versioning our filters as well: https://github.com/WP-API/WP-API/pull/937#discussion_r28469875

Does anyone have an idea of how we can add versioning to our route registrations?

danielbachhuber:
> we still need to add versioning to the core endpoints

We should discuss the purpose of versioning too. To me, having versioned routes set the expectation responses will be consistent. Given we're letting plugin authors do whatever, responses can't ever be expected to be consistent. I'm not sure what versioned routes are intended to solve.

rachelbaker:
1. We need to add support for versioning to our endpoint registration function so that plugins can version their own endpoints.  Example: /gravity-forms/v2.3/forms
2. Having endpoint versions allows us to make breaking changes with new API versions.
3. Support for endpoint versions was a requirement for the v2.0 rewrite. https://github.com/WP-API/WP-API/issues/571

apokalyptik:
To the point of not being able to version because we can't expect responses to be consistent due to plugins changing data around. 

I understand what you mean. But I think we don't need to get too carried away with being too literal. We've had some discussions like this in the past about trying to police what devs to, and it's not a good idea. What we can and should do is police ourselves, and give devs the ability to do the same.  Worrying about the possibility of a plugin breaking the site (or in this case versioning of the API) isn't a super good use of our time. Doing the right thing for ourselves and providing others the tools to walk in our footsteps certainly is.

rmccue:
Agreed with @apokalyptik. The version number is a strong indicator of what to expect (i.e. "which API handling class should I use?"), but clients still need to be robust enough to know the response might change independently of that.

With regards to versioning internally: it's always been my intent to change the prefixes on everything for v2. This sounds like a good time to do it. V2 can use `WP_REST_` for class names, and `rest_` for filters. If/When we do a v3, we can pick a different prefix then, possibly just `WP_REST_V3` and `rest_v3`, but I don't think we should be too tied to that.

[...]

PR #1153, opened 2015-04-28
Move all core routes to v2 namespace.

rachelbaker:
Adds the endpoint versioning piece of #1119 

Core routes now use the `wp/v2` namespace.


danielbachhuber:
> Core routes now use the wp/v2 namespace.

Do we need `wp/v2` if the `wp` route never existed in the first place? Could we just start with `wp/v1` ?

rachelbaker:
>>> Do we need wp/v2 if the wp route never existed in the first place? Could we just start with wp/v1

My understanding is that we are versioning the API (version 2.0) not the specific endpoint.

PR #1154, opened 2015-04-29
Inline docs for rest_query_navigation_headers action

Shelob9:


danielbachhuber:
Looks like the build failed for an unrelated reason.

PR #1158, opened 2015-04-29
json_ to rest_

quasel:
Not sure but i think this should be renamed too #1151


rachelbaker:
@quasel Good catch.  Thanks.

Issue #716, opened 2014-12-10
Accept JSON by default for update and create item routes.


rachelbaker:
When registering a `POST` or `PUT` route you have to explicitly set "'accept_json'     => true".  Let's set this to a default of `true`.

PR #1160, opened 2015-04-29
accept_json got renamed to accept_rest

quasel:
but isn't used anway afaik as of #716 this got removed entirely ?


rachelbaker:
Good catch again.

Issue #557, opened 2014-10-29
Add PHP Codesniffer to make sure we're adhering to coding standards


danielbachhuber:
To make this the best code it can be, let's sniff it out automatically.

danielbachhuber:
@westonruter Up for taking another swing at this?

westonruter:
Refreshed PR #558. PHPCS is still complaining, but there aren't as many issues as before: https://travis-ci.org/WP-API/WP-API/jobs/51793479

danielbachhuber:
:ship:

PR #1168, opened 2015-05-02
Run WPCS on every build

danielbachhuber:
See #557

### DIFF
--- a/phpcs.ruleset.xml
+++ b/phpcs.ruleset.xml
@@ -5,6 +5,7 @@
 	<rule ref="WordPress-VIP">
 		<exclude name="WordPress.VIP.FileSystemWritesDisallow" />
 		<exclude name="WordPress.VIP.RestrictedFunctions" />
+		<exclude name="WordPress.VIP.RestrictedVariables" />
 		<exclude name="WordPress.VIP.SuperGlobalInputUsage" />
 		<exclude name="WordPress.VIP.ValidatedSanitizedInput" />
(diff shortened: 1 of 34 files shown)


phpcs.ruleset.xml
```diff
@@ -0,0 +1,9 @@
+<?xml version="1.0"?>
+<ruleset name="WP-API">
+	<description>Sniffs for the coding standards of the WP-API plugin</description>
+
+	<rule ref="WordPress-VIP">
```
rmccue: Looks like the VIP rules are too strict here; we're getting `get_term_by is prohibited, please use wpcom_vip_get_term_by() instead.`

I'm not familiar enough with the VIP sniffs, but can we keep the security/etc ones but ditch the VIP-only ones?

rmccue: The ones I'm seeing that we could exclude:
* get_term_by is prohibited, please use wpcom_vip_get_term_by() instead.
* file_get_contents is highly discouraged, please use wpcom_vip_file_get_contents() instead
* switch_to_blog is not something you should ever need to do in a VIP theme context. Instead use an API (XML-RPC, REST) to interact with other sites if needed

Possibly also "Due to using Batcache, server side based client related logic will not work, use JS instead."

danielbachhuber: > I'm not familiar enough with the VIP sniffs, but can we keep the security/etc ones but ditch the VIP-only ones?

We need to exclude more. I haven't gotten that far yet.

PR #1174, opened 2015-05-03
Fix extra carriage return issue

morwalz:
This is fix to issues of extra carriage retrun. Without it in many cases it gives json validation error Expecting 'STRING', 'NUMBER', 'NULL', 'TRUE', 'FALSE', '{', '[', got 'undefined'

### DIFF
--- a/lib/infrastructure/class-wp-rest-server.php
+++ b/lib/infrastructure/class-wp-rest-server.php
@@ -290,8 +290,9 @@ class WP_REST_Server {
 			// Embed links inside the request
 			$result = $this->response_to_data( $result, isset( $_GET['_embed'] ) );
 
-			$result = json_encode( $result );
-
+			//$result = json_encode( $result );
+			//to remove carriage returns. This is fix to issues of json validation
+			$result = preg_replace( '/\s+/', '\n',json_encode( $this->prepare_response( $result ) ));
 			$json_error_message = $this->get_json_last_error();
 			if ( $json_error_message ) {
 				$json_error_obj = new WP_Error( 'rest_encode_error', $json_error_message, array( 'status' => 500 ) );


rmccue:
Can you give specific data examples here? What are you using to validate the JSON output?

morwalz:
Please look at this data set generate from wp json api plugin. (now after applying this patch this works fine). 
When there is no utf-8 characters then plugin output works very fine. 
I am validating this output in javascript using JSON.parse
http://www.examsbook.com/hi/wp-json/posts?filter[post_status]=publish&filter[cat]=8&filter[posts_per_page]=10&page=1&filter[orderby]=date&filter[order]=desc

rmccue:
Could be possible that `json_encode` is failing to encode the characters correctly.

@pento From memory, is this something that `wp_json_encode` handles?

rmccue:
> I am validating this output in javascript using JSON.parse

Also note that if you're passing this to `JSON.parse` that it expects a string, and it's possible that the data you're passing in is already decoded into native JS objects. I don't *think* that's the issue here, but it has popped up before.

pento:
`wp_json_encode()` doesn't have a specific thing for handling extra carriage returns, no.

It sounds like a thing that we could handle, though. Is there a smaller test case of an object that can be passed to `json_encode()` to generate invalid JSON?

morwalz:
@rmccue no i verified this clearly. It took me 6 hours to verify whats going wrong. I am using this in ionic framework. First i run the code using forking the ionic. But i think source of APIs should send correct response. So i tried to found fix here. In case of special char it returns 'LR CF ' at many places in output string.

joehoyle:
@morwalz could you provide some sample data that causes this issue? It would be easier for us to debug this then.

morwalz:
yes. i can provide dump of the site. you can install db locally and add wp-json api and test the output with JSON.parse. Let me know.

danielbachhuber:
We're using `wp_json_encode()` now that the infrastructure is in core.

@morwalz if you still have access to test data, it would be great if you could open a ticket on [core Trac](https://core.trac.wordpress.org/) so we can see if this problem is something we can fix in `wp_json_encode()`

Issue #1143, opened 2015-04-27
Multiple errors via param validation should be programmatically readable


joehoyle:
Currently if a request contains multiple invalid request args, we compile an error string showing all the issues. However, we don't pass all the different fields that errored. From @rmccue in slack:

> `new WP_Error( 'json_missing_field', 'Missing $field', array( 'key' => $field ) )`
> Would output as `[ { "code": "json_missing_field", "message": "Missing abc", "data": { "key": "abc" } }]`

joehoyle:
Fixed by #1175

PR #1175, opened 2015-05-04
Include programmatically readable erros for validation

joehoyle:
This was the client know exactly which fields failed. See #1143


joehoyle:
@WP-API/amigos #reviewmerge

Issue #924, opened 2015-02-25
Query params for collections


danielbachhuber:
_If you have a bug to report relating to querying for posts, terms, users, comments, etc., please open a new issue — this issue is intended to be a master reference._

This issue is a master issue representing work on query parameters in collections. Here's a table of the currently supported parameters:

```
+----------------+-------+-------+-------+----------+------------+-------+-------+
| param          | posts | pages | media | comments | categories | tags  | users |
+----------------+-------+-------+-------+----------+------------+-------+-------+
| after          |   x   |   x   |   x   |   x      |            |       |       |
| author         |   x   |   x   |   x   |   x      |            |       |       |
| author_email   |       |       |       |   x      |            |       |       |
| author_exclude |   x   |   x   |   x   |   x      |            |       |       |
| before         |   x   |   x   |   x   |   x      |            |       |       |
| context        |   x   |   x   |   x   |   x      |   x        |   x   |   x   |
| exclude        |   x   |   x   |   x   |   x      |   x        |   x   |   x   |
| filter         |   x   |   x   |   x   |          |            |       |       |
| hide_empty     |       |       |       |          |   x        |   x   |       |
| include        |   x   |   x   |   x   |   x      |   x        |   x   |   x   |
| karma          |       |       |       |   x      |            |       |       |
| media_type     |       |       |   x   |          |            |       |       |
| menu_order     |       |   x   |       |          |            |       |       |
| mime_type      |       |       |   x   |          |            |       |       |
| offset         |   x   |   x   |   x   |   x      |            |   x   |   x   |
| order          |   x   |   x   |   x   |   x      |   x        |   x   |   x   |
| orderby        |   x   |   x   |   x   |   x      |   x        |   x   |   x   |
| page           |   x   |   x   |   x   |   x      |   x        |   x   |   x   |
| parent         |       |   x   |   x   |   x      |   x        |       |       |
| parent_exclude |       |   x   |   x   |   x      |            |       |       |
| per_page       |   x   |   x   |   x   |   x      |   x        |   x   |   x   |
| post           |       |       |       |   x      |   x        |   x   |       |
| search         |   x   |   x   |   x   |   x      |   x        |   x   |   x   |
| slug           |   x   |   x   |   x   |          |   x        |   x   |   x   |
| status         |   x   |   x   |   x   |   x      |            |       |       |
| type           |       |       |       |   x      |            |       |       |
+----------------+-------+-------+-------+----------+------------+-------+-------+
```

Related #605

[...]

Issue #1047, opened 2015-04-03
Reconsider how schemas are exposed


danielbachhuber:
Currently, we're registering paths like `/wp/posts/schema`. I'm not a huge fan of this though, seems to be special-casing too much.

Maybe `_schema`?

rmccue:
`_links.profile` would be the correct place to link the schema. Alternatively, we could expose it as the body on an OPTIONS request? (I think we're allowed to do that.)

danielbachhuber:
> Alternatively, we could expose it as the body on an OPTIONS request? (I think we're allowed to do that.)

:boom: #1222

PR #1181, opened 2015-05-05
Declare term collection params so they can be exposed

danielbachhuber:
See #924

### DIFF
--- a/lib/endpoints/class-wp-rest-controller.php
+++ b/lib/endpoints/class-wp-rest-controller.php
@@ -206,19 +206,19 @@ abstract class WP_REST_Controller {
 				'type'               => 'integer',
 				'default'            => 1,
 				'sanitize_callback'  => 'absint',
-				),
+			),
 			'per_page'               => array(
 				'description'        => 'Maximum number of items to be returned in result set.',
 				'type'               => 'integer',
 				'default'            => 10,
 				'sanitize_callback'  => 'absint',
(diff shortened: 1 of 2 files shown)


danielbachhuber:
@WP-API/amigos Can I get some validation on this concept? If it looks good, I'd like to merge, and then handle each controller separately.

rmccue:
Looks fine by me.

lib/endpoints/class-wp-rest-controller.php
```diff
@@ -173,6 +173,33 @@ public function get_item_schema() {
...
+	 * @return array
+	 */
+	public function get_collection_params() {
+		return array(
+			'page'                   => array(
+				'description'        => 'Current page of the collection.',
+				'type'               => 'integer',
+				'default'            => 1,
+				'sanitize_callback'  => 'absint',
+				),
```
rmccue: Watch the indentation on your closing parens.

danielbachhuber:
@joehoyle gave a [little pushback](https://wordpress.slack.com/archives/core-restapi/p1430867483005306) on this yesterday.

I still think it's a good idea for this metadata to be stored on the controller instead of global state, but it's worth discussing in tandem with #1047

danielbachhuber:
> I still think it's a good idea for this metadata to be stored on the controller instead of global state

And, it makes subclassing possible.

PR #1182, opened 2015-05-06
Add pagination headers to term collection response

### LINKED ISSUE #534: Paginate terms queries and limit by default
#347 was just fixed, providing the ability to filter the collection retrieved by `/taxonomies/<taxonomy>/terms` endpoints. With this, we now have the ability to set the `number` filter value to specify the number of terms to retrieve. 

When the number of terms is limited, we should provide pagination headers for terms collections in the same manner as we do with `query_navigation_headers` for posts collections.

Once pagination headers are available we should also limit the number of terms retrieved by default, in a manner similar to the way `/posts` defaults to the number of posts per page specified in settings. Without an imposed limit, unfiltered posts queries on sites with tons of tags, such as `http://www.wired.com/wp-json/taxonomies/post_tag/terms`, will become so large that they timeout fail when there are too many terms available.

danielbachhuber:
Fixes #534

### DIFF
--- a/lib/endpoints/class-wp-rest-terms-controller.php
+++ b/lib/endpoints/class-wp-rest-terms-controller.php
@@ -151,6 +151,9 @@ class WP_REST_Terms_Controller extends WP_REST_Controller {
 		$base = add_query_arg( $request->get_query_params(), rest_url( '/wp/v2/terms/' . $this->get_taxonomy_base( $this->taxonomy ) ) );
 		if ( $request['page'] > 1 ) {
 			$prev_page = $request['page'] - 1;
+			if ( $prev_page > $max_pages ) {
+				$prev_page = $max_pages;
+			}
 			$prev_link = add_query_arg( 'page', $prev_page, $base );
 			$response->link_header( 'prev', $prev_link );
 		}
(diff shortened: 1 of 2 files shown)


danielbachhuber:
@WP-API/amigos #reviewmerge

lib/endpoints/class-wp-rest-terms-controller.php
```diff
@@ -140,7 +140,30 @@ public function get_items( $request ) {
 			$response[] = $this->prepare_response_for_collection( $data );
 		}
 
-		return rest_ensure_response( $response );
+		$response = rest_ensure_response( $response );
+		unset( $prepared_args['number'] );
+		unset( $prepared_args['offset'] );
+		$total_terms = wp_count_terms( $this->taxonomy, $prepared_args );
+		$response->header( 'X-WP-Total', (int) $total_terms );
```
rachelbaker: @danielbachhuber Many hosts limit the size of HTTP headers.  I don't know if we ran into this issue with v1.0, but I believe @rmccue and I discussed moving these values into the collection response itself.  @rmccue am I recalling correctly??

danielbachhuber: I'd like to implement as headers for now, as it follows the existing implementation, and we can reformat at a later date in this milestone.

rmccue: > I don't know if we ran into this issue with v1.0

We definitely did, WP Engine limits more aggressively than other hosts and we saw problems there. That was only because we sent one-link-per-item though.

Should be fine here (we do the same for posts) for now, I think the plan was to review when we look at the links-for-collections stuff

rachelbaker:
Merged #1182

PR #1191, opened 2015-05-06
Pagination headers for `GET /wp/v2/users`

### LINKED ISSUE #537: Paginate user endpoint and limit by default
The users endpoint collection should be returned with pagination headers

(This may [i.e. will] depend on WP_User_Query updates)

As with #534 for taxonomy terms, once pagination is available the users endpoint should be limited by default to a set number of users to prevent timeout errors when hitting /users on sites with many users. It may be appropriate to use the same number of posts per page specified in the admin.

danielbachhuber:
Fixes #537


danielbachhuber:
@WP-API/amigos #reviewmerge

rachelbaker:
Merged #1191

Issue #1145, opened 2015-04-27
Correct phpDoc for WP_JSON_Controller


rmccue:
There's a few things in here right now that are annoying Scrutinizer as well as my autocomplete in my editor:

* `@return mixed WP_Error or WP_JSON_Response.` uses an incorrect expression for types, this should be `WP_JSON_Response|WP_Error`
* `get_item`, `create_item`, `update_item`, `delete_item` are missing param/return doc
* `@return mixed WP_Error|bool.` shouldn't have the trailing dot (only needed with a description)

joehoyle:
Fixed

PR #1200, opened 2015-05-07
Fix lots of doc blocks

joehoyle:
Mainly return types in a standardized format


joehoyle:
See @1145

rachelbaker:
Merged #1200 

Fixes #1145

PR #1203, opened 2015-05-07
Only link `author` when there's a valid author

### LINKED ISSUE #1192: Don't link `author` when `author=0`
image

danielbachhuber:
Fixes #1192

### DIFF
--- a/lib/endpoints/class-wp-rest-posts-controller.php
+++ b/lib/endpoints/class-wp-rest-posts-controller.php
@@ -1074,7 +1074,8 @@ class WP_REST_Posts_Controller extends WP_REST_Controller {
 			),
 		);
 
-		if ( in_array( $post->post_type, array( 'post', 'page' ) ) || post_type_supports( $post->post_type, 'author' ) ) {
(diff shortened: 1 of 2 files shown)


danielbachhuber:
@WP-API/amigos #reviewmerge

lib/endpoints/class-wp-rest-posts-controller.php
```diff
@@ -1074,7 +1074,8 @@ protected function prepare_links( $post ) {
 			),
 		);
 
-		if ( in_array( $post->post_type, array( 'post', 'page' ) ) || post_type_supports( $post->post_type, 'author' ) ) {
+		if ( ( in_array( $post->post_type, array( 'post', 'page' ) ) || post_type_supports( $post->post_type, 'author' ) )
```
joehoyle: Out of curiosity, why do we whitelist the post types here?

danielbachhuber: To force the data to always be present, even if the developer has removed post type supports.

Our posts and pages are fixed schemas. Except not, of course.

PR #1209, opened 2015-05-07
Set pagination headers even when no found posts

danielbachhuber:
Also handles adding `prev` link when outside of bounds

Removes `query_navigation_headers()` from `WP_REST_Response`, which was an oddball.

Fixes #890

### DIFF
--- a/tests/test-rest-posts-controller.php
+++ b/tests/test-rest-posts-controller.php
@@ -78,6 +78,70 @@ class WP_Test_REST_Posts_Controller extends WP_Test_REST_Post_Type_Controller_Te
 		}
 	}
 
+	/**
+	 * @group test
+	 */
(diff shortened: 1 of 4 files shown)


danielbachhuber:
@WP-API/amigos #reviewmerge

tests/test-rest-posts-controller.php
```diff
@@ -78,6 +78,70 @@ public function test_get_items_status_without_permissions() {
 		}
 	}
 
+	/**
+	 * @group test
```
joehoyle: #protip `phpunit --filter test_get_items_pagination_headers`

danielbachhuber: Derp. Can you remove?

PR #1210, opened 2015-05-07
Use Coveralls for code coverage

rmccue:
Don't merge this one yet, just testing on this branch.

### DIFF
--- a/.coveralls.yml
+++ b/.coveralls.yml
@@ -1,2 +1,3 @@
 src_dir: lib
 coverage_clover: clover.xml
+json_path: tests/coveralls.json
--- a/.coveralls.yml
+++ b/.coveralls.yml
@@ -1,0 +1,2 @@
+src_dir: lib
+coverage_clover: clover.xml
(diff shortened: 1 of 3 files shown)


danielbachhuber:
These messages will get annoying. Can we pipe to Slack instead?

rmccue:
Will switch to Slack. Do we want to leave the build status one as well?

danielbachhuber:
> Do we want to leave the build status one as well?

Sorry, not sure I follow?

rmccue:
screenshot 2015-05-10 23 02 35

Right now, it's just set to report here (should do it straight after Travis is done, so shouldn't block). We can set a minimum threshold (e.g. fail if it drops below x%), or a relative minimum threshold (e.g. fail if it drops by x%), but dunno whether we want that.

danielbachhuber:
> Right now, it's just set to report here (should do it straight after Travis is done, so shouldn't block).

Oh, nifty. Works for me.

rmccue:
Comments off. Still on me to sort out getting Slack hooked up with someone with Slack admin.

rachelbaker:
@rmccue Sounds like we aren't connecting this to Slack, okay to merge?

rmccue:
Forgot to merge this after talking with Sam, let's hit it.

Issue #1199, opened 2015-05-07
Add and remove taxonomy terms from a post


danielbachhuber:
As a WP-API client, I'd expect to be able to assign and unassign any custom taxonomy terms from my posts.

danielbachhuber:
:ship:

Issue #1214, opened 2015-05-07
Respect `term_order` for taxonomies


danielbachhuber:
`register_taxonomy()` supports a `sort=true` flag, which means terms should be returned in the order they were assigned.

Related #1199

danielbachhuber:
This can go in beta3

danielbachhuber:
> register_taxonomy() supports a sort=true flag, which means terms should be returned in the order they were assigned.

As it turns out, I was a crazy person when I wrote this. `wp_get_object_terms()` supports an `orderby` argument, but there's no metadata registered on the taxonomy object. We'll need to support query params for querying a post's terms.

PR #1170, opened 2015-05-03
Retool `WP_REST_Terms_Controller` to follow Posts controller pattern

danielbachhuber:
By passing the taxonomy to the controller, we can more easily allow
custom controllers for taxonomies, and for those custom taxonomies to
have custom schemas

Fixes #1061
See also #899

PR #1216, opened 2015-05-08
Add controller for a post's terms

joehoyle:
Should we allow `DELETE /wp/v2/posts/1/terms/post_tag` to remove all tags from the post?

Also should we allow `POST /wp/v2/posts/1/terms/post_tag` with body `{ ids: 1,2,3,4 }` to assign more than 1 term to a post?

See #1199

### DIFF
--- a/lib/endpoints/class-wp-rest-posts-terms-controller.php
+++ b/lib/endpoints/class-wp-rest-posts-terms-controller.php
@@ -60,7 +60,7 @@ class WP_REST_Posts_Terms_Controller extends WP_REST_Controller {
 
 		$post = get_post( $request['id'] );
 
-		$is_request_valid = $this->validate_request();
+		$is_request_valid = $this->validate_request( $request );
 		if ( is_wp_error( $is_request_valid ) ) {
 			return $is_request_valid;
 		}
@@ -88,7 +88,7 @@ class WP_REST_Posts_Terms_Controller extends WP_REST_Controller {
 		$post     = get_post( $request['id'] );
 		$term_id  = absint( $request['term_id'] );
 
-		$is_request_valid = $this->validate_request();
+		$is_request_valid = $this->validate_request( $request );
 		if ( is_wp_error( $is_request_valid ) ) {
 			return $is_request_valid;
 		}
@@ -116,7 +116,7 @@ class WP_REST_Posts_Terms_Controller extends WP_REST_Controller {
 		$post     = get_post( $request['id'] );
 		$term_id  = absint( $request['term_id'] );
 
-		$is_request_valid = $this->validate_request();
+		$is_request_valid = $this->validate_request( $request );
 		if ( is_wp_error( $is_request_valid ) ) {
 			return $is_request_valid;
 		}
@@ -145,7 +145,7 @@ class WP_REST_Posts_Terms_Controller extends WP_REST_Controller {
 		$post     = get_post( $request['id'] );
 		$term_id  = absint( $request['term_id'] );
 
-		$is_request_valid = $this->validate_request();
+		$is_request_valid = $this->validate_request( $request );
 		if ( is_wp_error( $is_request_valid ) ) {
 			return $is_request_valid;
 		}
--- a/lib/endpoints/class-wp-rest-posts-terms-controller.php
+++ b/lib/endpoints/class-wp-rest-posts-terms-controller.php
@@ -204,13 +204,20 @@ class WP_REST_Posts_Terms_Controller extends WP_REST_Controller {
 	 * @return bool|WP_Error
 	 */
 	public function get_items_permissions_check( $request ) {
-		$post_check = $this->posts_controller->get_item_permissions_check( $request );
+
+		$post_request = new WP_REST_Request();
+		$post_request->set_param( 'id', $request['post_id'] );
+
+		$post_check = $this->posts_controller->get_item_permissions_check( $post_request );
 
 		if ( ! $post_check || is_wp_error( $post_check ) ) {
 			return $post_check;
 		}
 
-		$terms_check = $this->terms_controller->get_item_permissions_check( $request );
+		$term_request = new WP_REST_Request();
+		$term_request->set_param( 'id', $request['term_id'] );
+
+		$terms_check = $this->terms_controller->get_item_permissions_check( $term_request );
 
 		if ( ! $terms_check || is_wp_error( $terms_check ) ) {
 			return $terms_check;
@@ -227,7 +234,9 @@ class WP_REST_Posts_Terms_Controller extends WP_REST_Controller {
 	 */
 	public function create_item_permissions_check( $request ) {
 
-		$post_check = $this->posts_controller->update_item_permissions_check( $request );
+		$post_request = new WP_REST_Request();
+		$post_request->set_param( 'id', $request['post_id'] );
+		$post_check = $this->posts_controller->update_item_permissions_check( $post_request );
 
 		if ( ! $post_check || is_wp_error( $post_check ) ) {
 			return $post_check;
--- a/lib/endpoints/class-wp-rest-posts-terms-controller.php
+++ b/lib/endpoints/class-wp-rest-posts-terms-controller.php
@@ -55,8 +55,9 @@ class WP_REST_Posts_Terms_Controller extends WP_REST_Controller {
 
 		$post = get_post( $request['id'] );
 
-		if ( empty( $post->ID ) || $this->post_type !== $post->post_type ) {
-			return new WP_Error( 'rest_post_invalid_id', __( 'Invalid post ID.' ), array( 'status' => 404 ) );
+		$post_check = $this->posts_controller->get_item( $request );
+		if ( is_wp_error( $post_check ) ) {
+			return $post_check;
 		}
 
 		$valid_taxonomies = get_object_taxonomies( $post->post_type );
@@ -87,8 +88,9 @@ class WP_REST_Posts_Terms_Controller extends WP_REST_Controller {
 		$post     = get_post( $request['id'] );
 		$term_id  = absint( $request['term_id'] );
 
-		if ( empty( $post->ID ) || $this->post_type !== $post->post_type ) {
-			return new WP_Error( 'rest_post_invalid_id', __( 'Invalid post ID.' ), array( 'status' => 404 ) );
+		$post_check = $this->posts_controller->get_item( $request );
+		if ( is_wp_error( $post_check ) ) {
+			return $post_check;
 		}
 
 		$valid_taxonomies = get_object_taxonomies( $post->post_type );
@@ -123,8 +125,9 @@ class WP_REST_Posts_Terms_Controller extends WP_REST_Controller {
 		$post     = get_post( $request['id'] );
 		$term_id  = absint( $request['term_id'] );
 
-		if ( empty( $post->ID ) || $this->post_type !== $post->post_type ) {
-			return new WP_Error( 'rest_post_invalid_id', __( 'Invalid post ID.' ), array( 'status' => 404 ) );
+		$post_check = $this->posts_controller->get_item( $request );
+		if ( is_wp_error( $post_check ) ) {
+			return $post_check;
 		}
 
 		$valid_taxonomies = get_object_taxonomies( $post->post_type );
@@ -160,8 +163,9 @@ class WP_REST_Posts_Terms_Controller extends WP_REST_Controller {
 		$post     = get_post( $request['id'] );
 		$term_id  = absint( $request['term_id'] );
 
-		if ( empty( $post->ID ) || $this->post_type !== $post->post_type ) {
-			return new WP_Error( 'rest_post_invalid_id', __( 'Invalid post ID.' ), array( 'status' => 404 ) );
+		$post_check = $this->posts_controller->get_item( $request );
+		if ( is_wp_error( $post_check ) ) {
+			return $post_check;
 		}
 
 		$valid_taxonomies = get_object_taxonomies( $post->post_type );
--- a/lib/endpoints/class-wp-rest-posts-terms-controller.php
+++ b/lib/endpoints/class-wp-rest-posts-terms-controller.php
@@ -18,7 +18,7 @@ class WP_REST_Posts_Terms_Controller extends WP_REST_Controller {
 
 		$base = $this->posts_controller->get_post_type_base( $this->post_type );
 
-		register_rest_route( 'wp/v2', sprintf( '/%s/(?P<id>[\d]+)/terms/%s', $base, $this->taxonomy ), array(
+		register_rest_route( 'wp/v2', sprintf( '/%s/(?P<post_id>[\d]+)/terms/%s', $base, $this->taxonomy ), array(
 			array(
 				'methods'             => WP_REST_Server::READABLE,
 				'callback'            => array( $this, 'get_items' ),
@@ -26,7 +26,7 @@ class WP_REST_Posts_Terms_Controller extends WP_REST_Controller {
 			),
 		) );
 
-		register_rest_route( 'wp/v2', sprintf( '/%s/(?P<id>[\d]+)/terms/%s/(?P<term_id>[\d]+)', $base, $this->taxonomy ), array(
+		register_rest_route( 'wp/v2', sprintf( '/%s/(?P<post_id>[\d]+)/terms/%s/(?P<term_id>[\d]+)', $base, $this->taxonomy ), array(
 			array(
 				'methods'             => WP_REST_Server::READABLE,
 				'callback'            => array( $this, 'get_item' ),
@@ -44,7 +44,7 @@ class WP_REST_Posts_Terms_Controller extends WP_REST_Controller {
 			),
 		) );
 
-		register_rest_route( 'wp/v2', sprintf( '/%s/(?P<id>[\d]+)/terms/%s', $base, $this->taxonomy ) . '/schema', array(
+		register_rest_route( 'wp/v2', sprintf( '/%s/(?P<post_id>[\d]+)/terms/%s', $base, $this->taxonomy ) . '/schema', array(
 			'methods'         => WP_REST_Server::READABLE,
 			'callback'        => array( $this, 'get_item_schema' ),
 		) );
@@ -58,7 +58,7 @@ class WP_REST_Posts_Terms_Controller extends WP_REST_Controller {
 	 */
 	public function get_items( $request ) {
 
-		$post = get_post( $request['id'] );
+		$post = get_post( absint( $request['post_id'] ) );
 
 		$is_request_valid = $this->validate_request( $request );
 		if ( is_wp_error( $is_request_valid ) ) {
@@ -85,7 +85,7 @@ class WP_REST_Posts_Terms_Controller extends WP_REST_Controller {
 	 * @return WP_Error|WP_REST_Response
 	 */
 	public function get_item( $request ) {
-		$post     = get_post( $request['id'] );
+		$post     = get_post( absint( $request['post_id'] ) );
 		$term_id  = absint( $request['term_id'] );
 
 		$is_request_valid = $this->validate_request( $request );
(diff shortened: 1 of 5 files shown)


joehoyle:
@WP-API/amigos #review

lib/endpoints/class-wp-rest-posts-terms-controller.php
```diff
@@ -0,0 +1,229 @@
+<?php
+
+class WP_REST_Posts_Terms_Controller extends WP_REST_Controller {
+
+	protected $post_type;
+
+	public function __construct( $post_type ) {
```
danielbachhuber: Need to pass `taxonomy` here too, as the route should be explicitly defined based on the taxonomy. See #1170

lib/endpoints/class-wp-rest-posts-terms-controller.php
```diff
@@ -0,0 +1,229 @@
...
+	/**
+	 * Get a collection of posts
+	 *
+	 * @param WP_REST_Request $request Full details about the request
+	 * @return WP_Error|WP_REST_Response
+	 */
+	public function get_items( $request ) {
+
+		$post = get_post( $request['id'] );
+		$taxonomy = $request['taxonomy'];
```
danielbachhuber: `$taxonomy` should be a property of the controller, not a passed argument.

lib/endpoints/class-wp-rest-posts-terms-controller.php
joehoyle: :+1:

lib/endpoints/class-wp-rest-posts-terms-controller.php
```diff
@@ -0,0 +1,229 @@
...
+	 * @return WP_Error|WP_REST_Response
+	 */
+	public function get_item( $request ) {
+		$post     = get_post( $request['id'] );
+		$taxonomy = $request['taxonomy'];
+		$term_id  = absint( $request['term_id'] );
+
+		if ( empty( $post->ID ) || $this->post_type !== $post->post_type ) {
+			return new WP_Error( 'rest_post_invalid_id', __( 'Invalid post ID.' ), array( 'status' => 404 ) );
+		}
```
danielbachhuber: Seems like this check could be abstracted (or, use the check in the posts controller)

lib/endpoints/class-wp-rest-posts-terms-controller.php
```diff
@@ -0,0 +1,229 @@
...
+
+		$valid_taxonomies = get_object_taxonomies( $post->post_type );
+		if ( ! in_array( $taxonomy, $valid_taxonomies ) ) {
+			return new WP_Error( 'rest_post_taxonomy_invalid', __( 'Invalid taxonomy for post ID.' ), array( 'status' => 404 ) );
+		}
+
+		if ( ! get_term_by( 'term_taxonomy_id', $term_id, $taxonomy ) ) {
```
danielbachhuber: This should probably use the Terms controller.

lib/endpoints/class-wp-rest-posts-terms-controller.php
```diff
@@ -0,0 +1,229 @@
...
+		) );
+	}
+
+	/**
+	 * Get a collection of posts
+	 *
+	 * @param WP_REST_Request $request Full details about the request
+	 * @return WP_Error|WP_REST_Response
+	 */
+	public function get_items( $request ) {
```
danielbachhuber: Pagination headers?

lib/endpoints/class-wp-rest-posts-terms-controller.php
```diff
@@ -0,0 +1,229 @@
...
+		if ( empty( $post->ID ) || $this->post_type !== $post->post_type ) {
+			return new WP_Error( 'rest_post_invalid_id', __( 'Invalid post ID.' ), array( 'status' => 404 ) );
+		}
+
+		$valid_taxonomies = get_object_taxonomies( $post->post_type );
+		if ( ! in_array( $taxonomy, $valid_taxonomies ) ) {
+			return new WP_Error( 'rest_post_taxonomy_invalid', __( 'Invalid taxonomy for post ID.' ), array( 'status' => 404 ) );
+		}
+
+		$terms = wp_get_object_terms( $post->ID, $taxonomy );
```
danielbachhuber: Shouldn't we error if it's not in the set?

lib/endpoints/class-wp-rest-posts-terms-controller.php
```diff
@@ -0,0 +1,229 @@
...
+		$term = $terms_controller->prepare_item_for_response( get_term_by( 'term_taxonomy_id', $term_id, $taxonomy ), $request );
+
+		$response = rest_ensure_response( $term );
+		$response->set_status( 201 );
+
+		return $term;
+	}
+
+	/**
+	 * Delete a single post.
```
danielbachhuber: Copy pasta

lib/endpoints/class-wp-rest-posts-terms-controller.php
```diff
@@ -0,0 +1,229 @@
...
+		$terms_controller = new WP_REST_Terms_Controller( $request['taxonomy'] );
+		$term = $terms_controller->prepare_item_for_response( get_term_by( 'term_taxonomy_id', $term_id, $taxonomy ), $request );
+
+		$response = rest_ensure_response( $term );
+
+		return $response;
+	}
+
+	/**
+	 * Create a single post
```
danielbachhuber: Copy pasta

lib/endpoints/class-wp-rest-posts-terms-controller.php
```diff
@@ -0,0 +1,229 @@
...
+			$response[] = $this->prepare_response_for_collection( $data );
+		}
+
+		$response = rest_ensure_response( $response );
+
+		return $response;
+	}
+
+	/**
+	 * Get a single post
```
danielbachhuber: Copy pasta

lib/endpoints/class-wp-rest-posts-terms-controller.php
```diff
@@ -0,0 +1,229 @@
...
+			array(
+				'methods'         => WP_REST_Server::DELETABLE,
+				'callback'        => array( $this, 'delete_item' ),
+				'permission_callback' => array( $this, 'create_item_permissions_check' ),
+			),
+		) );
+	}
+
+	/**
+	 * Get a collection of posts
```
danielbachhuber: Copy pasta

danielbachhuber:
Some nits.

lib/endpoints/class-wp-rest-posts-terms-controller.php
joehoyle: Added in 20621e8100a7e1089a43cbbfd201950ade65475a

lib/endpoints/class-wp-rest-posts-terms-controller.php
joehoyle: Fixed in 20621e8100a7e1089a43cbbfd201950ade65475a

lib/endpoints/class-wp-rest-posts-terms-controller.php
joehoyle: Fixed in e695af3

lib/endpoints/class-wp-rest-posts-terms-controller.php
joehoyle: As this requires setting up a new Request object, I don't know how pretty that's going to be

joehoyle: Created https://github.com/WP-API/WP-API/issues/1218 to make this a bit easier hopefully.

lib/endpoints/class-wp-rest-posts-terms-controller.php
joehoyle: Not 100% sure. I guess for consistency we should add it? I do wonder whether typically developers would expect this or not. Pagination on object propper is somewhat expected, but how many people doing a simple GET that returns 10 tags would think to look for more?

This may be made a bit better with the wrapper object for collection responses to make it more obvious there is more results available.

lib/endpoints/class-wp-rest-posts-terms-controller.php
joehoyle: Done!

lib/endpoints/class-wp-rest-posts-terms-controller.php
joehoyle: 40e9c22

lib/endpoints/class-wp-rest-posts-terms-controller.php
joehoyle: 40e9c22

lib/endpoints/class-wp-rest-posts-terms-controller.php
joehoyle: 40e9c22

lib/endpoints/class-wp-rest-posts-terms-controller.php
joehoyle: 40e9c22

joehoyle:
Ready for another look!

lib/endpoints/class-wp-rest-posts-terms-controller.php
```diff
@@ -0,0 +1,225 @@
...
+	public function get_items( $request ) {
+
+		$post = get_post( $request['id'] );
+
+		$post_check = $this->posts_controller->get_item( $request );
+		if ( is_wp_error( $post_check ) ) {
+			return $post_check;
+		}
+
+		$valid_taxonomies = get_object_taxonomies( $post->post_type );
```
danielbachhuber: Do we need to check `show_in_rest` or `public` values for this taxonomy?

lib/endpoints/class-wp-rest-posts-terms-controller.php
```diff
@@ -0,0 +1,225 @@
...
+	 * Get all the terms that are attached to a post
+	 *
+	 * @param WP_REST_Request $request Full details about the request
+	 * @return WP_Error|WP_REST_Response
+	 */
+	public function get_items( $request ) {
+
+		$post = get_post( $request['id'] );
+
+		$post_check = $this->posts_controller->get_item( $request );
```
danielbachhuber: This is likely affected by https://github.com/WP-API/WP-API/issues/1218#issuecomment-103655685

lib/endpoints/class-wp-rest-posts-terms-controller.php
danielbachhuber: I think we can punt for now. Term collections aren't typically paginated at the moment — not sure core's APIs support it either.

lib/endpoints/class-wp-rest-posts-terms-controller.php
```diff
@@ -0,0 +1,225 @@
...
+
+	/**
+	 * Get a term that is attached to a post
+	 *
+	 * @param WP_REST_Request $request Full details about the request
+	 * @return WP_Error|WP_REST_Response
+	 */
+	public function get_item( $request ) {
+		$post     = get_post( $request['id'] );
+		$term_id  = absint( $request['term_id'] );
```
danielbachhuber: Why do you sanitize one and not the other?

Also, would it make more sense to flip these? e.g. `term_id` is `id` and we use `post` or something similar to how we've been naming other fields.

lib/endpoints/class-wp-rest-posts-terms-controller.php
```diff
@@ -0,0 +1,225 @@
...
+		}
+
+		$valid_taxonomies = get_object_taxonomies( $post->post_type );
+		if ( ! in_array( $this->taxonomy, $valid_taxonomies ) ) {
+			return new WP_Error( 'rest_post_taxonomy_invalid', __( 'Invalid taxonomy for post ID.' ), array( 'status' => 404 ) );
+		}
+
+		if ( ! get_term_by( 'term_taxonomy_id', $term_id, $this->taxonomy ) ) {
+			return new WP_Error( 'rest_term_invalid', __( "Term doesn't exist." ), array( 'status' => 404 ) );
+		}
```
danielbachhuber: Seems like the previous ~10 lines of checks should be abstracted in some form.

lib/endpoints/class-wp-rest-posts-terms-controller.php
```diff
@@ -0,0 +1,225 @@
...
```
danielbachhuber: Same comments as above.

lib/endpoints/class-wp-rest-posts-terms-controller.php
```diff
@@ -0,0 +1,225 @@
...
+
+		$tt_ids = wp_set_object_terms( $post->ID, $term_id, $this->taxonomy, true );
+
+		if ( is_wp_error( $tt_ids ) ) {
+			return $tt_ids;
+		}
+
+		$term = $this->terms_controller->prepare_item_for_response( get_term_by( 'term_taxonomy_id', $term_id, $this->taxonomy ), $request );
```
danielbachhuber: Same comment as above.

lib/endpoints/class-wp-rest-posts-terms-controller.php
```diff
@@ -0,0 +1,225 @@
...
+		}
+
+		$valid_taxonomies = get_object_taxonomies( $post->post_type );
+		if ( ! in_array( $this->taxonomy, $valid_taxonomies ) ) {
+			return new WP_Error( 'rest_post_taxonomy_invalid', __( 'Invalid taxonomy for post ID.' ), array( 'status' => 404 ) );
+		}
+
+		$terms = wp_get_object_terms( $post->ID, $this->taxonomy );
+		$terms = wp_list_filter( $terms, array( 'term_taxonomy_id' => $term_id ), 'NOT' );
+		$tt_ids = wp_set_object_terms( $post->ID, $terms, $this->taxonomy );
```
danielbachhuber: Worth opening a core ticket for a function to remove a term from an object? I think you might lose term order with this #1214

lib/endpoints/class-wp-rest-posts-terms-controller.php
```diff
@@ -0,0 +1,225 @@
...
+
+		if ( is_wp_error( $tt_ids ) ) {
+			return $tt_ids;
+		}
+
+		return null;
```
danielbachhuber: Should this return the original set, per #1224 ?

plugin.php
```diff
@@ -168,6 +169,16 @@ function create_initial_rest_routes() {
 			$revisions_controller = new WP_REST_Revisions_Controller( $post_type->name );
 			$revisions_controller->register_routes();
 		}
+
+		foreach ( get_object_taxonomies( $post_type->name, 'objects' ) as $taxonomy ) {
+
+			if ( ! $taxonomy->public ) {
```
danielbachhuber: Should we defer to `show_in_rest` ?

tests/test-rest-posts-terms-controller.php
```diff
@@ -0,0 +1,235 @@
...
+	public function test_update_item() {
+		/** Can't update a relationship **/
+	}
+
+	public function test_prepare_item() {
+		/** Post types can't be updated **/
+	}
+
+	public function test_get_item_schema() {
+		/** Post types can't be deleted **/
```
danielbachhuber: Copy pasta?

danielbachhuber:
@joehoyle left some comments

lib/endpoints/class-wp-rest-posts-terms-controller.php
joehoyle: This this should use `wp_remove_object_terms`, my bad

lib/endpoints/class-wp-rest-posts-terms-controller.php
joehoyle: Yes I think so

lib/endpoints/class-wp-rest-posts-terms-controller.php
joehoyle: Based off https://github.com/WP-API/WP-API/blob/develop/lib/endpoints/class-wp-rest-taxonomies-controller.php#L92-93 I think no (or at least that this point anyway). If we want to adopt `show_in_rest` for taxonomies, it should probably be applied universally and start at the taxonomies controller.

joehoyle: Oh, maybe I need to check public though

joehoyle: Actually, turns out we don't need this now that the controller is a post_type / taxonomy pair. This checking is done already on route registration in `plugin.php`.

lib/endpoints/class-wp-rest-posts-terms-controller.php
joehoyle: I used `id` in this case so we can pass the `$request` to the `WP_REST_Posts_Controller` as that used the `id` param too.

lib/endpoints/class-wp-rest-posts-terms-controller.php
joehoyle: Fixed in 36c8073

lib/endpoints/class-wp-rest-posts-terms-controller.php
joehoyle: Fixed in 36c8073

lib/endpoints/class-wp-rest-posts-terms-controller.php
joehoyle: Fixed in 65f0698

lib/endpoints/class-wp-rest-posts-terms-controller.php
joehoyle: Fixed in 65f0698

plugin.php
joehoyle: We don't use `show_in_rest` for taxonomies (yet) I think.

tests/test-rest-posts-terms-controller.php
joehoyle: Implemented test and added the schema endpoint in d3d6d6e3b2bc0f99f3de38f0bd2a84093a4d8ab8

joehoyle:
@danielbachhuber want another bash?

lib/endpoints/class-wp-rest-posts-terms-controller.php
danielbachhuber: Hm. I'd think `id` should always represent the primary resource.

joehoyle: I think the post _is_ the primary resource

lib/endpoints/class-wp-rest-posts-terms-controller.php
danielbachhuber: Can we do internal REST request here to make it more loosely coupled?

plugin.php
danielbachhuber: #1278

lib/endpoints/class-wp-rest-posts-terms-controller.php
joehoyle: We already have the tight coupling to the `terms_controller` here in several other places, so I think this should remain consistent with using that. For example we use `get_item` in the validation method. Also, if a route is removed for `/terms/post_tag` for whatever reason, internally routing will not return what is expected.

plugin.php
danielbachhuber: Er: 

image

joehoyle: Ah, I was looking at the taxonomies controller, I think need to update that too in that case.

lib/endpoints/class-wp-rest-posts-terms-controller.php
danielbachhuber: We use `id` and `parent_id` for meta: https://github.com/WP-API/WP-API/blob/develop/lib/endpoints/class-wp-rest-meta-controller.php#L61

joehoyle: I changed `id` to `post_id` in f82d7c2

plugin.php
joehoyle: Fixed in b661b99

danielbachhuber:
:cake:

Issue #1204, opened 2015-05-07
Support changing parent on hierarchical taxonomies.


danielbachhuber:
On hierarchical taxonomies, a client should be able to update the `parent` of a term. We'll need to make sure the supplied `term_taxonomy_id` is transformed to the `term_id` stored in the database.

PR #1219, opened 2015-05-08
Only permit querying by parent term when tax is hierarchical

danielbachhuber:
Related #1204


danielbachhuber:
@WP-API/amigos #reviewmerge

PR #1220, opened 2015-05-08
Add meta tag to head that contains the path URL for WP-API.

superphly:
This might be helpful to build in now, so that in the future if the API url is changed or maybe on a different server altogether, it can be relayed to the frontend.


rmccue:
Sorry for missing this PR! We already link with the `<link>` tag, as well as with a `Link` HTTP header (and in the RSD too), so you should be able to parse those out. See http://v2.wp-api.org/guide/discovery/ for documentation on that.

Closing this, since the `<meta>` tag is a bit redundant with `<link>`, but if there's a reason `<link>` doesn't suffice, let us know. :)

PR #1221, opened 2015-05-08
Handle term parent on update / create term

### LINKED ISSUE #1204: Support changing parent on hierarchical taxonomies.

joehoyle:
Fixes #1204

### DIFF
--- a/lib/endpoints/class-wp-rest-terms-controller.php
+++ b/lib/endpoints/class-wp-rest-terms-controller.php
@@ -59,7 +59,6 @@ class WP_REST_Terms_Controller extends WP_REST_Controller {
 				'permission_callback' => array( $this, 'update_item_permissions_check' ),
 				'args'       => array(
 					'name'        => array(
-						'required'          => true,
 						'sanitize_callback' => 'sanitize_text_field',
 					),
 					'description' => array(
--- a/lib/endpoints/class-wp-rest-terms-controller.php
+++ b/lib/endpoints/class-wp-rest-terms-controller.php
@@ -184,6 +184,20 @@ class WP_REST_Terms_Controller extends WP_REST_Controller {
 			$args['slug'] = $request['slug'];
 		}
 
+		if ( isset( $request['parent'] ) ) {
+			if ( ! is_taxonomy_hierarchical( $this->taxonomy ) ) {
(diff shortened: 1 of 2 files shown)


lib/endpoints/class-wp-rest-terms-controller.php
```diff
@@ -193,6 +192,20 @@ public function create_item( $request ) {
 			$args['slug'] = $request['slug'];
 		}
 
+		if ( isset( $request['parent'] ) ) {
+			if ( ! is_taxonomy_hierarchical( $this->taxonomy ) ) {
+				return new WP_Error( 'rest_taxonomy_not_hierarchical', __( "Can not set term parent, taxonomy is not hierarchical." ), array( 'status' => 400 ) );
+			}
```
danielbachhuber: `parent` isn't present in the schema, so we don't need to worry about the taxonomy check.

lib/endpoints/class-wp-rest-terms-controller.php
```diff
@@ -224,6 +237,20 @@ public function update_item( $request ) {
 			$prepared_args['slug'] = $request['slug'];
 		}
 
+		if ( isset( $request['parent'] ) ) {
+			if ( ! is_taxonomy_hierarchical( $this->taxonomy ) ) {
+				return new WP_Error( 'rest_taxonomy_not_hierarchical', __( "Can not set term parent, taxonomy is not hierarchical." ), array( 'status' => 400 ) );
+			}
```
danielbachhuber: Same as above.

lib/endpoints/class-wp-rest-terms-controller.php
joehoyle: This is here to provide feedback to someone sending the parent and not realizing. I think we shouldn't proesume everyone is reading the schema

danielbachhuber: With #1223, the argument won't be included.

joehoyle: As #1223 hasn't been decided o nor included, I'm going to leave this in.

PR #1244, opened 2015-05-17
Fix bug in creating comments if comment_type was not set

### LINKED ISSUE #1230: New Comments Not Returned in /comments Collection
This is for v2.0 Beta 1

When I submit and approve a new comment it doesn't get returned in the `/comments` collection, but it does return when queried directly by its id `/comments/<id>`

rachelbaker:
Don't assume the `type` value is set since we don't require it.  

Fixes #1230  and adds test coverage.


rachelbaker:
@WP-API/amigos #reviewmerge

danielbachhuber:
:pineapple:

Issue #940, opened 2015-02-27
WP_JSON_Comments_Controller::prepare_item_for_update() is too restrictive


danielbachhuber:
All fields should be updateable, and `WP_JSON_Comments_Controller` should have one method for preparing data for the database.

rachelbaker:
@danielbachhuber this is an issue with `wp_update_comment()` not handling all of the comment object properties.  How do you suggest we handle this?

danielbachhuber:
> How do you suggest we handle this?

Probably need a corresponding core patch then.

rmccue:
Which specific fields are we missing support for here?

joehoyle:
I'm not sure I follow the description and title on this ticker. How is `prepare_item_for_update` "too restrictive"?

Is the issue that `WP_REST_Comments` needs to prepare all it's data, then once it's run `wp_update_comment` it needs to manually set some other fields?

danielbachhuber:
Sorry. Seems like the original description was unclear. The intent of this issue is that I should be able to update _any_ comment fields, which core doesn't currently allow.

rmccue:
Looks like `wp_update_comment` accepts updating:
```
'comment_post_ID', 'comment_content', 'comment_author', 'comment_author_email', 'comment_approved', 'comment_karma', 'comment_author_url', 'comment_date', 'comment_date_gmt', 'comment_type', 'comment_parent', 'user_id'
```

Do you mean the core of the API (i.e. *we* don't support updating all of these)? I'm still kinda unclear here.

danielbachhuber:
> Do you mean the core of the API (i.e. we don't support updating all of these)?

Er, I mean we only support updating a subset: https://github.com/WP-API/WP-API/blob/develop/lib/endpoints/class-wp-rest-comments-controller.php#L669

It looks like we also have an odd `prepare_item_for_update()` method we should probably break down.

rachelbaker:
I will pick this up.

rachelbaker:
Started in #1245, and should make Beta 2

rachelbaker:
Fixed in #1245

Issue #1205, opened 2015-05-07
Comments and Terms controllers should use get_endpoint_args_for_item_schema()


danielbachhuber:
They don't yet.

Related #1135

rachelbaker:
I will pick this one up.

rachelbaker:
I will wrap this one up tonight for Beta 2.

rachelbaker:
After looking more into the Comments endpoints I feel like setting the endpoint_args_from_schema is too rigid.  

1. We cannot set a default value on create.  (Which means for other endpoints where we use this method we don't set defaults).  
2. Inability to separate out create vs update arguments.  Example: you cannot set the comment `status` on create, but you can on update.  Another example: `you can set the comment `type` on create, but you cannot change it on update.  
3. Inability to set specific string sanitization like `wp_filter_post_kses` in the schema.

I thought about adding `sanitize_callback` directly into the schema but that seems very *un*schema-like.  

Some discussion @joehoyle and I had on this: https://github.com/WP-API/WP-API/pull/1245/files#r30509829

Leaving this open if someone else wants to jump in here.

danielbachhuber:
> Example: you cannot set the comment status on create, but you can on update.

Why can't you?

rachelbaker:
Punted to Beta 4

rachelbaker:
Done in #1399 and #1422

PR #1245, opened 2015-05-18
Expand args for updating comments

rachelbaker:
Consistent support for Comment create and update arguments:
  - Expands the update comment args to match the create arguments (including special handling for `date_gmt`)
  - Define each update argument and set sanitization callbacks
  - Removes the `prepare_item_for_update` method in favor of re-using the `prepare_item_for_database` method
  - Allow users with `moderate_comments` to set comment karma
  - Return an error when comment type is being changed, since `wp_update_comment` does not support this arg

Some other minor changes:
  - Each create argument has set default and sanitization callbacks
  - Return an error if the user is creating a comment tries to set the author to an invalid value
  - Don't set a default value for the comment_type
  - Remove double sanitization on create and update comment args

All with test coverage!!!

This sets up #1205 for the Comments endpoints.


rachelbaker:
Fixes #940 

@WP-API/amigos #reviewmerge

lib/endpoints/class-wp-rest-comments-controller.php
```diff
@@ -73,12 +76,36 @@ public function register_routes() {
 				'callback' => array( $this, 'update_item' ),
 				'permission_callback' => array( $this, 'update_item_permissions_check' ),
 				'args'     => array(
-					'post'         => array(),
-					'status'       => array(),
-					'content'      => array(),
-					'author'       => array(),
-					'author_email' => array(),
-					'author_url'   => array(),
+					'post'         => array(
```
joehoyle: @rachelbaker we have an inconsistency here in the endpoints - some use the Schema to define all the `update_item` / `create_item` arguments, but here we are not doing that. One reason that may be is because we want to specify the `sanitize_callback` for each argument, which using the Schema does not allow us to do. Welcome to my recurring issue I've been wanting to solve for a while!

rachelbaker: @joehoyle Hmm...another issue I can see with using the Schema is that on comment update you can change the 'status' (which you cannot set on create), so that field isn't completely `readonly`.  What do you think we should do here?

rachelbaker:
I need to refresh this so we can add handling to accept `date_gmt` on update, even though `wp_update_comment()` doesn't accept the parameter.

lib/endpoints/class-wp-rest-comments-controller.php
joehoyle: @rachelbaker we probably shouldn't mark it as `readonly` in that case I think.

rachelbaker: @joehoyle We also cannot set a `default` argument in the schema.  Feels like setting the endpoint arguments from the schema is too rigid at this point to fit all our routes.

rachelbaker: Related: #1205

rachelbaker:
Remaining:

- Throw error for comments that are trying to change the type.
- Add unit test for updating comment date_gmt.

I am still hoping to get this in Beta 3

rachelbaker:
@WP-API/amigos #reviewmerge

lib/endpoints/class-wp-rest-comments-controller.php
```diff
@@ -625,53 +678,61 @@ protected function prepare_status_response( $comment_approved ) {
...
-			$prepared_comment['comment_author_url'] = esc_url_raw( $request['author_url'] );
+			$prepared_comment['comment_author_url'] = $request['author_url'];
+		}
+
+		if ( isset( $request['type'] ) ) {
+			$prepared_comment['comment_type'] = $request['type'];
+		}
+
+		if ( isset( $request['karma'] ) ) {
+			$prepared_comment['comment_karma'] = (int) $request['karma'] ;
```
joehoyle: I think casting shouldn't be needed as this is sanitized with `absint`

joehoyle:
+1 from me on this, though I think @rmccue was going to look over it too.

lib/endpoints/class-wp-rest-comments-controller.php
rachelbaker: @joehoyle addressed in 15df46e

tests/test-rest-comments-controller.php
```diff
@@ -388,6 +391,96 @@ public function test_create_comment_other_user() {
...
 
+	public function test_create_comment_other_user_without_permission() {
+		wp_set_current_user( $this->subscriber_id );
+
+		$params = array(
+			'post'         => $this->post_id,
+			'author_name'  => 'Homer Jay Simpson',
+			'author_email' => 'chunkylover53@aol.com',
+			'author_url'   => 'http://compuglobalhypermeganet.com',
+			'content'      => 'Here’s to alcohol: the cause of, and solution to, all of life’s problems.',
```
rmccue: Note the smart quotes here in the string, might cause problems, since it relies on editors using UTF-8/etc. Can we use normal ' or escape sequences please? :)

(Also, they have the internet on computers now?)

rmccue:
:+1: on merge, just need to fix those couple of strings. :)

Issue #417, opened 2014-08-04
Hide "username", "First Name" & "Last Name" for embedded/non-auth'd author responses?


kadamwhite:
When dealing with themes, it's general practice to show the user's name as the value clearly labeled as "Display name publicly as":

![image](https://cloud.githubusercontent.com/assets/442115/3804240/860115f2-1c23-11e4-8ba9-64d738838869.png)

The default JSON response for embedded users (e.g. authors in posts) includes all of these values:

```json
{"author": {
  "ID": 4,
  "username": "kadamnation",
  "name": "K. Adam White",
  "first_name": "K. Adam",
  "last_name": "White",
  "nickname": "KAdam",
  "slug": "kadamnation",
  "URL": "http://www.mydomain.com",
  "avatar": "http://0.gravatar.com/avatar/6cd1cf6de05952505a165425efaf06b4?s=96",
  "description": "Some short biographical description about the user",
  "registered": "2012-06-17T15:41:32+00:00",
  "meta": {
    "links": {
      "self": "http://www.mydomain.com/wp-json/users/4",
      "archives": "http://www.mydomain.com/wp-json/users/4/posts"
    }
  }
}}
```
Of these, we discussed at WCNYC that we may want to limit the information about a user that we expose. I propose limiting embedded response data (and responses for non-authenticated requests, see #297) to these values:

* **name**: (the name explicitly selected for display on the site)
* **slug**: The public slug representation of the user, for use with routes etc
* **nickname**: If this gets populated, I think it'd make sense to expose it since I've seen it commonly used on things like author archive pages

This would mean we would *omit* the following:

* **username**: Even though this is probably the same as slug (need to confirm this), we probably don't want to give away internal account names where possible
* **first_name**: Superseded by *name*
* **last_name**: Superseded by *name*
* We also discussed removing **registered**, unless anyone can think of a compelling reason why it should be in the public responses

PR #1251, opened 2015-05-19
Limit the Users response fields in the embed context

rachelbaker:
Changed the following fields to only display in the view or edit contexts:
    - first_name
    - last_name
    - slug
    - nickname 

Includes test coverage.

The `embed` context now only includes the following response fields:

`avatar_url`
`id`
`link`
`name`
`url`
`description`

### DIFF
--- a/lib/endpoints/class-wp-rest-users-controller.php
+++ b/lib/endpoints/class-wp-rest-users-controller.php
@@ -594,7 +594,7 @@ class WP_REST_Users_Controller extends WP_REST_Controller {
 				'description' => array(
 					'description' => 'Description of the object.',
 					'type'        => 'string',
-					'context'     => array( 'view', 'edit' ),
+					'context'     => array( 'embed', 'view', 'edit' ),
 				),
 				'email'       => array(
 					'description' => 'The email address for the object.',
@@ -662,7 +662,7 @@ class WP_REST_Users_Controller extends WP_REST_Controller {
 					'description' => 'URL of the object.',
 					'type'        => 'string',
 					'format'      => 'uri',
-					'context'     => array( 'view', 'edit' ),
+					'context'     => array( 'embed', 'view', 'edit' ),
 					'readonly'    => true,
 				),
 				'username'    => array(
--- a/lib/endpoints/class-wp-rest-users-controller.php
+++ b/lib/endpoints/class-wp-rest-users-controller.php
@@ -594,7 +594,7 @@ class WP_REST_Users_Controller extends WP_REST_Controller {
 				'description' => array(
 					'description' => 'Description of the object.',
 					'type'        => 'string',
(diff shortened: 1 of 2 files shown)


rachelbaker:
Fixes #417

@WP-API/amigos #reviewmerge

danielbachhuber:
Why are you removing `url` and `description` ?

rachelbaker:
>>> Why are you removing url and description ?

@danielbachhuber because those details aren't always publically available. Some themes expose them, but we shouldn't assume all sites want that data public.

danielbachhuber:
> Some themes expose them, but we shouldn't assume all sites want that data public.

I thought we decided in NYC that we'd be exposing the data?

rachelbaker:
@danielbachhuber You are right, confirmed with the logic laid out in #417.

lib/endpoints/class-wp-rest-users-controller.php
```diff
@@ -656,7 +656,7 @@ public function get_item_schema() {
 				'slug'        => array(
 					'description' => 'An alphanumeric identifier for the object unique to its type.',
 					'type'        => 'string',
-					'context'     => array( 'embed', 'view', 'edit' ),
+					'context'     => array( 'view', 'edit' ),
```
danielbachhuber: Isn't `slug` exposed in the URL?

danielbachhuber: And by URL I mean `link`

rachelbaker: @danielbachhuber It is, but it also defaults to the `user_login` so it makes sense to me to not include it in the embedded response.

danielbachhuber: > slug also defaults to the user_login so I don't see a reason to expose it in the embedded response.

Why not? Most other resources all have `slug`: Post, Term

rachelbaker: @danielbachhuber Is there a use case for having the Users slug without authentication?  Keeping in mind that `get_author_posts_url()` is filterable to not use `user_nicename` (or `slug`)

danielbachhuber: > Is there a use case for having the Users slug without authentication?

Consistency.

rachelbaker: I feel very strongly that `slug` is a User field I would not expect to be available in the embed context.  The field is available in the `view` and `edit` context.  

Is there a reason besides consistency?

danielbachhuber: > Is there a reason besides consistency?

Not at the top of my head, but I'm sleep deprived.

We can just see if it comes up again.

Issue #1224, opened 2015-05-08
Return resource data instead of random message when deleting Post, Comment, Term, User


danielbachhuber:
This will let the client restore the resource if they need to.

danielbachhuber:
This is continued in #1260

PR #1253, opened 2015-05-19
Return full user object when deleting user

danielbachhuber:
This will let the client more easily restore the data, should they need
to

See #1224

### DIFF
--- a/lib/endpoints/class-wp-rest-users-controller.php
+++ b/lib/endpoints/class-wp-rest-users-controller.php
@@ -312,12 +312,16 @@ class WP_REST_Users_Controller extends WP_REST_Controller {
 			}
 		}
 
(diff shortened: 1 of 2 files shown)


danielbachhuber:
@WP-API/amigos #reviewmerge

lib/endpoints/class-wp-rest-users-controller.php
```diff
@@ -312,12 +312,16 @@ public function delete_item( $request ) {
 			}
 		}
 
+		$mock_request = new WP_REST_Request( 'GET', '/wp/users/' . $id );
```
danielbachhuber: Hmm, this should fail?

danielbachhuber: Oh. Because it's not routing through the server, the request can be whatever. I guess ultimately it should route through the server, so it's not hard-coupled to this controller.

PR #1254, opened 2015-05-19
Return full term object when deleting term

danielbachhuber:
See #1224


danielbachhuber:
@WP-API/amigos #reviewmerge

PR #1255, opened 2015-05-19
Return full comment object when deleting comment

danielbachhuber:
See #1224


danielbachhuber:
@WP-API/amigos #reviewmerge

PR #1256, opened 2015-05-19
Return full post object when deleting post

danielbachhuber:
See #1224

### DIFF
--- a/lib/endpoints/class-wp-rest-posts-controller.php
+++ b/lib/endpoints/class-wp-rest-posts-controller.php
@@ -327,6 +327,10 @@ class WP_REST_Posts_Controller extends WP_REST_Controller {
 			return new WP_Error( 'rest_user_cannot_delete_post', __( 'Sorry, you are not allowed to delete this post.' ), array( 'status' => 401 ) );
 		}
 
+		$mock_request = new WP_REST_Request( 'GET', rest_url( 'wp/v2/' . $this->get_post_type_base( $this->post_type ) . '/' . $post->ID ) );
+		$mock_request->set_param( 'context', 'edit' );
+		$response = $this->prepare_item_for_response( $post, $mock_request );
+
 		// If we're forcing, then delete permanently
 		if ( $force ) {
 			$result = wp_delete_post( $id, true );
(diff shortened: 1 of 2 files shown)


danielbachhuber:
@WP-API/amigos #reviewmerge

rachelbaker:
I would like @rmccue to weigh in here and in: #1255, #1254, and #1253

rachelbaker:
From the HTTP spec for [DELETE](http://www.w3.org/Protocols/rfc2616/rfc2616-sec9.html#sec9.7/):

> A successful response SHOULD be 200 (OK) if the response includes an entity describing the status, 202 (Accepted) if the action has not yet been enacted, or 204 (No Content) if the action has been enacted but the response does not include an entity.

I don’t know what `an entity describing the status` means

lib/endpoints/class-wp-rest-posts-controller.php
```diff
@@ -327,6 +327,10 @@ public function delete_item( $request ) {
 			return new WP_Error( 'rest_user_cannot_delete_post', __( 'Sorry, you are not allowed to delete this post.' ), array( 'status' => 401 ) );
 		}
 
+		$mock_request = new WP_REST_Request( 'GET', rest_url( 'wp/v2/' . $this->get_post_type_base( $this->post_type ) . '/' . $post->ID ) );
+		$mock_request->set_param( 'context', 'edit' );
```
rmccue: `$mock_request` isn't really the right name, since this is a real request and "mock" has a specific meaning. `$get_request`?

lib/endpoints/class-wp-rest-posts-controller.php
```diff
@@ -341,16 +345,12 @@ public function delete_item( $request ) {
 			$result = wp_trash_post( $id );
 		}
 
-		if ( ! $result ) {
+		if ( $result ) {
```
rmccue: We should flip this, and eliminate the unneeded else. In general, the pattern should be like:
```php
if ( ! $x ) {
    return new WP_Error;
}

if ( ! $y ) {
    return new WP_Error;
}

return 'success';
```

rmccue:
This generally looks OK to me. One thing we might want to consider is wrapping the post in another object (e.g. `{ data: {...}, trashed: true, gone: false }`) to make it clear that the request did succeed (since we're meant to be describing the status), buuuuut that might be overkill.

rmccue:
Also, should we consider this from #627?
```php
if ( $new_status === 'trash' ) {
	// Send a 202 if it's only trashed to indicate that it will be
	// actually deleted later
	$response->set_status( 202 );
```

i.e. 200 = permanently deleted, 202 = trashed

Not sure we need the distinction now with the split from double-DELETE to separate trash/perma-delete, but worth thinking about.

danielbachhuber:
> Not sure we need the distinction now with the split from double-DELETE to separate trash/perma-delete, but worth thinking about.

See #1260

lib/endpoints/class-wp-rest-posts-controller.php
danielbachhuber: I've always thought it to be a bad pattern to check for a negative.

@joehoyle @rachelbaker ?

PR #1259, opened 2015-05-19
Use a constant every time we want an impossibly high number

danielbachhuber:
If we ever get to the impossibly high number and our tests fail
crypically, this will be easier to fix in one fell swoop


danielbachhuber:
@WP-API/amigos #reviewmerge

rachelbaker:
Merged #1259

Issue #1222, opened 2015-05-08
Return schema and endpoint args in body of OPTIONS request


danielbachhuber:
Internally the schema will need to be registered through `register_rest_route()`.

PR #1262, opened 2015-05-19
Return user schema and request args in body of OPTIONs request

danielbachhuber:
See #1222

### DIFF
--- a/lib/endpoints/class-wp-rest-users-controller.php
+++ b/lib/endpoints/class-wp-rest-users-controller.php
@@ -17,6 +17,7 @@ class WP_REST_Users_Controller extends WP_REST_Controller {
 				'callback'        => array( $this, 'get_items' ),
 				'permission_callback' => array( $this, 'get_items_permissions_check' ),
 				'args'            => $query_params,
+				'schema'          => $this->get_item_schema(),
 			),
 			array(
 				'methods'         => WP_REST_Server::CREATABLE,
@@ -27,6 +28,7 @@ class WP_REST_Users_Controller extends WP_REST_Controller {
 						'required' => true,
 					),
 				) ),
+				'schema'          => $this->get_item_schema(),
 			),
 		) );
 		register_rest_route( 'wp/v2', '/users/(?P<id>[\d]+)', array(
@@ -39,6 +41,7 @@ class WP_REST_Users_Controller extends WP_REST_Controller {
 						'default'      => 'embed',
 					),
 				),
+				'schema'          => $this->get_item_schema(),
 			),
 			array(
 				'methods'         => WP_REST_Server::EDITABLE,
@@ -47,6 +50,7 @@ class WP_REST_Users_Controller extends WP_REST_Controller {
 				'args'            => array_merge( $this->get_endpoint_args_for_item_schema( false ), array(
 					'password'    => array()
 				) ),
+				'schema'          => $this->get_item_schema(),
 			),
 			array(
 				'methods' => WP_REST_Server::DELETABLE,
@@ -55,6 +59,7 @@ class WP_REST_Users_Controller extends WP_REST_Controller {
 				'args' => array(
 					'reassign' => array(),
 				),
+				'schema'          => $this->get_item_schema(),
 			),
 		) );
 
@@ -63,13 +68,10 @@ class WP_REST_Users_Controller extends WP_REST_Controller {
 			'callback'        => array( $this, 'get_current_item' ),
 			'args'            => array(
 				'context'          => array(),
-			)
+			),
+			'schema'          => $this->get_item_schema(),
 		));
 
-		register_rest_route( 'wp/v2', '/users/schema', array(
-			'methods'         => WP_REST_Server::READABLE,
-			'callback'        => array( $this, 'get_item_schema' ),
-		) );
 	}
 
 	/**
--- a/lib/endpoints/class-wp-rest-users-controller.php
+++ b/lib/endpoints/class-wp-rest-users-controller.php
@@ -17,7 +17,6 @@ class WP_REST_Users_Controller extends WP_REST_Controller {
 				'callback'        => array( $this, 'get_items' ),
 				'permission_callback' => array( $this, 'get_items_permissions_check' ),
 				'args'            => $query_params,
-				'schema'          => $this->get_item_schema(),
 			),
 			array(
 				'methods'         => WP_REST_Server::CREATABLE,
@@ -28,8 +27,8 @@ class WP_REST_Users_Controller extends WP_REST_Controller {
 						'required' => true,
 					),
 				) ),
-				'schema'          => $this->get_item_schema(),
 			),
+			'schema'          => $this->get_item_schema(),
 		) );
 		register_rest_route( 'wp/v2', '/users/(?P<id>[\d]+)', array(
 			array(
@@ -41,7 +40,6 @@ class WP_REST_Users_Controller extends WP_REST_Controller {
 						'default'      => 'embed',
 					),
 				),
-				'schema'          => $this->get_item_schema(),
 			),
 			array(
 				'methods'         => WP_REST_Server::EDITABLE,
@@ -50,7 +48,6 @@ class WP_REST_Users_Controller extends WP_REST_Controller {
 				'args'            => array_merge( $this->get_endpoint_args_for_item_schema( false ), array(
 					'password'    => array(),
 				) ),
-				'schema'          => $this->get_item_schema(),
 			),
 			array(
 				'methods' => WP_REST_Server::DELETABLE,
@@ -59,8 +56,8 @@ class WP_REST_Users_Controller extends WP_REST_Controller {
 				'args' => array(
 					'reassign' => array(),
 				),
-				'schema'          => $this->get_item_schema(),
 			),
+			'schema'          => $this->get_item_schema(),
 		) );
 
 		register_rest_route( 'wp/v2', '/users/me', array(
--- a/lib/endpoints/class-wp-rest-users-controller.php
+++ b/lib/endpoints/class-wp-rest-users-controller.php
@@ -323,7 +323,7 @@ class WP_REST_Users_Controller extends WP_REST_Controller {
 		if ( ! $result ) {
 			return new WP_Error( 'rest_cannot_delete', __( 'The user cannot be deleted.' ), array( 'status' => 500 ) );
 		}
-		
+
 		return $orig_user;
 	}
 
--- a/plugin.php
+++ b/plugin.php
@@ -599,7 +599,6 @@ function rest_handle_options_request( $response, $handler, $request ) {
 			if ( empty( $body['schema'] ) && ! empty( $endpoint['schema'] ) ) {
 				$body['schema'] = $endpoint['schema'];
 			}
-
 		}
 		break;
 	}
--- a/plugin.php
+++ b/plugin.php
@@ -563,6 +563,10 @@ function rest_handle_options_request( $response, $handler, $request ) {
 	$response = new WP_REST_Response();
 
 	$accept = array();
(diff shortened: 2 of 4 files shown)


lib/endpoints/class-wp-rest-users-controller.php
```diff
@@ -55,6 +59,7 @@ public function register_routes() {
 				'args' => array(
 					'reassign' => array(),
 				),
+				'schema'          => $this->get_item_schema(),
```
danielbachhuber: This repeat declaration of `schema` isn't great. Better ideas?

plugin.php
```diff
@@ -573,12 +577,25 @@ function rest_handle_options_request( $response, $handler, $request ) {
 
 		foreach ( $endpoints as $endpoint ) {
 			$accept = array_merge( $accept, $endpoint['methods'] );
+
+			$methods = array_keys( $endpoint['methods'] );
+			$method = array_shift( $methods );
```
danielbachhuber: Why are endpoints registered in this format?

```
'methods' =>
    array (
      'DELETE' => true,
    ),
```

When would a method be `false`, or an endpoint registered supporting two methods?

plugin.php
```diff
@@ -573,12 +577,25 @@ function rest_handle_options_request( $response, $handler, $request ) {
...
+
+			if ( empty( $body['request_args'][ $method ] ) ) {
+				$body['request_args'][ $method ] = $endpoint['args'];
+			}
+
+			if ( empty( $body['schema'] ) && ! empty( $endpoint['schema'] ) ) {
```
danielbachhuber: Hacky thing to default to the first time the schema was registered for a resource.

tests/test-rest-users-controller.php
```diff
@@ -674,12 +674,17 @@ public function test_delete_user_invalid_reassign_id() {
 		$this->assertErrorResponse( 'rest_user_invalid_reassign', $response, 400 );
 	}
 
-	public function test_get_item_schema() {
-		$request = new WP_REST_Request( 'GET', '/wp/v2/users/schema' );
+	public function test_options_request() {
+
+		$request = new WP_REST_Request( 'OPTIONS', '/wp/v2/users' );
```
danielbachhuber: If you did `OPTIONS /wp/v2/users/1`, would you expect to get the same?

danielbachhuber:
@joehoyle Would love your input on this. The abstraction seems a bit messy at the moment.

plugin.php
rmccue: Mainly this was so it's a constant-time lookup internally in `dispatch`. Endpoints often support more than one method as well, all EDITABLE supports POST, PUT, and PATCH.

danielbachhuber:
Response looks a bit like this right now:

```
{
  "request_args": {
    "GET": {
      "page": {
        "description": "Current page of the collection.",
        "type": "integer",
        "default": 1,
        "sanitize_callback": "absint"
      },
      "per_page": {
        "description": "Maximum number of items to be returned in result set.",
        "type": "integer",
        "default": 10,
        "sanitize_callback": "absint"
      },
      "search": {
        "description": "Limit results to those matching a string.",
        "type": "string",
        "sanitize_callback": "sanitize_text_field"
      },
      "context": {
        "default": "view",
        "description": "Change the response format based on request context.",
        "enum": [
          "view",
          "edit"
        ],
        "sanitize_callback": "sanitize_key",
        "type": "string"
      },
      "order": {
        "default": "asc",
        "description": "Order sort attribute ascending or descending.",
        "enum": [
          "asc",
          "desc"
        ],
        "sanitize_callback": "sanitize_key",
        "type": "string"
      },
      "orderby": {
        "default": "name",
        "description": "Sort collection by object attribute.",
        "enum": [
          "id",
          "name",
          "registered_date"
        ],
        "sanitize_callback": "sanitize_key",
        "type": "string"
      }
    },
    "POST": {
      "capabilities": [],
      "description": [],
      "email": {
        "required": true
      },
      "first_name": [],
      "last_name": [],
      "name": [],
      "nickname": [],
      "roles": [],
      "slug": [],
      "username": {
        "required": true
      },
      "password": {
        "required": true
      }
    }
  },
  "schema": {
    "$schema": "http://json-schema.org/draft-04/schema#",
    "title": "user",
    "type": "object",
    "properties": {
      "avatar_url": {
        "description": "Avatar URL for the object.",
        "type": "string",
        "format": "uri",
        "context": [
          "embed",
          "view",
          "edit"
        ],
        "readonly": true
      },
      "capabilities": {
        "description": "All capabilities assigned to the user.",
        "type": "object",
        "context": [
          "view",
          "edit"
        ]
      },
      "description": {
        "description": "Description of the object.",
        "type": "string",
        "context": [
          "embed",
          "view",
          "edit"
        ]
      },
      "email": {
        "description": "The email address for the object.",
        "type": "string",
        "format": "email",
        "context": [
          "view",
          "edit"
        ],
        "required": true
      },
      "extra_capabilities": {
        "description": "Any extra capabilities assigned to the user.",
        "type": "object",
        "context": [
          "edit"
        ],
        "readonly": true
      },
      "first_name": {
        "description": "First name for the object.",
        "type": "string",
        "context": [
          "embed",
          "view",
          "edit"
        ]
      },
      "id": {
        "description": "Unique identifier for the object.",
        "type": "integer",
        "context": [
          "embed",
          "view",
          "edit"
        ],
        "readonly": true
      },
      "last_name": {
        "description": "Last name for the object.",
        "type": "string",
        "context": [
          "embed",
          "view",
          "edit"
        ]
      },
      "link": {
        "description": "Author URL to the object.",
        "type": "string",
        "format": "uri",
        "context": [
          "embed",
          "view",
          "edit"
        ],
        "readonly": true
      },
      "name": {
        "description": "Display name for the object.",
        "type": "string",
        "context": [
          "embed",
          "view",
          "edit"
        ]
      },
      "nickname": {
        "description": "The nickname for the object.",
        "type": "string",
        "context": [
          "embed",
          "view",
          "edit"
        ]
      },
      "registered_date": {
        "description": "Registration date for the user.",
        "type": "date-time",
        "context": [
          "view",
          "edit"
        ],
        "readonly": true
      },
      "roles": {
        "description": "Roles assigned to the user.",
        "type": "array",
        "context": [
          "view",
          "edit"
        ]
      },
      "slug": {
        "description": "An alphanumeric identifier for the object unique to its type.",
        "type": "string",
        "context": [
          "embed",
          "view",
          "edit"
        ]
      },
      "url": {
        "description": "URL of the object.",
        "type": "string",
        "format": "uri",
        "context": [
          "embed",
          "view",
          "edit"
        ],
        "readonly": true
      },
      "username": {
        "description": "Login name for the user.",
        "type": "string",
        "context": [
          "edit"
        ],
        "required": true
      }
    }
  }
}
```

danielbachhuber:
@joehoyle Could you weigh in on this? Thanks

lib/endpoints/class-wp-rest-users-controller.php
joehoyle: Can't we register the schema on the route rather than each endpoint?

plugin.php
```diff
@@ -584,12 +588,24 @@ function rest_handle_options_request( $response, $handler, $request ) {
...
+				$body['schema'] = $endpoint['schema'];
+			}
 		}
 		break;
 	}
 	$accept = array_keys( $accept );
 
 	$response->header( 'Accept', implode( ', ', $accept ) );
+	$response->set_data( (object) $body );
```
joehoyle: Why is this cast to an object? Pretty much everything else return an array / `ArrayAccess`

lib/endpoints/class-wp-rest-users-controller.php
danielbachhuber: How?

plugin.php
danielbachhuber: Because I want to make sure it's always an object, even when empty. Otherwise it will be an array when empty.

lib/endpoints/class-wp-rest-users-controller.php
joehoyle: Like this:

```PHP
register_rest_route( 'wp/v2', '/posts', array(
   array( ... 
), false, array( 'schema' => $schema ) );

File: plugin.php (hunk 5 above)
Comment (joehoyle) [inline plugin.php]: How would it be empty, per https://github.com/WP-API/WP-API/pull/1262/files#diff-55d3bf5e833c76ee6e3c2f4be15fa1cdR577 ?

Comment (danielbachhuber) [inline plugin.php]: Oh. Derp. I guess it was empty at one point, and then I changed my code.

File: plugin.php [hunk 6]
```diff
@@ -584,12 +588,24 @@ function rest_handle_options_request( $response, $handler, $request ) {
 
 		foreach ( $endpoints as $endpoint ) {
 			$accept = array_merge( $accept, $endpoint['methods'] );
+
+			$methods = array_keys( $endpoint['methods'] );
+			$method = array_shift( $methods );
+
+			if ( empty( $body['request_args'][ $method ] ) ) {
+				$body['request_args'][ $method ] = $endpoint['args'];
```
danielbachhuber: Should I use `create`, `update`, `delete`, `get` here, instead of HTTP methods?

joehoyle:
Implementation is ok I think - though don't like the repeated schema so much - as I think there should only be one schema per route.

By using `GET` / `POST` what do we do when someone uses `WP_REST_Server::UPDATABLE` as that technically supports `PUT` and `POST` I believe.

plugin.php
joehoyle: Just mentioned it in my comment too! I think yes, a map with `WP_REST_Server::CREATABLE` and all those constants would be best I think.

tests/test-rest-users-controller.php
danielbachhuber: @joehoyle thoughts on ^^

joehoyle: Weird, though I added a comment. Yes, though not quite the same, I'd expect to get the `get` / `update` / `delete` request arguments, as opposed to `get` / `create`.

danielbachhuber: Thanks. I think I have enough to wrap this up now.

lib/endpoints/class-wp-rest-users-controller.php
rmccue: Let's not add more parameters to `register_rest_route`. Can we just add it to the existing route stuff?

```php
register_rest_route( 'wp/v2', '/comments', array(
	array(
		'methods'   => WP_REST_Server::READABLE,
		...
	),
	array(
		'methods'  => WP_REST_Server::CREATABLE,
		...
	),
	'schema' => ...
) );
```

rmccue: (Similar to meta queries with `relation` mixed with the indexed arrays)

rmccue:
So, question: if you don't already know how to form the URL, how do you get the schema for it? For example, `posts/{id}`; how do I get the schema for that? Kind of annoying.

(Working with a few others on porting the dotcom console and noticed this one.)

danielbachhuber:
[Some discussion in Slack](https://wordpress.slack.com/archives/core-restapi/p1432596205000130). 

Consensus is to chuck everything into the index, as well as return it for `OPTIONS` on the endpoint.

rmccue:
Splitting the index part out to #1305 so we can ship this sucker.

@WP-API/amigos #reviewmerge

rachelbaker:
Punting to b3 to get passing tests.

rmccue:
> Punting to b3 to get passing tests.

(Also, because as pointed out, this really only affects user endpoints, and we should probably tackle them all.)

rachelbaker:
@danielbachhuber,

I believe this has been taken care of in #1415

danielbachhuber:
This has since been addressed.

PR #1267, opened 2015-05-20
Add check when creating a user to verify the provided role is valid.

### LINKED ISSUE #1252: Add check for role when creating a user
In #1249 we also added a check to make sure the user role was valid when being updated, we should add the same check when creating a user.

rachelbaker:
Fixes #1252

### DIFF
--- a/tests/test-rest-users-controller.php
+++ b/tests/test-rest-users-controller.php
@@ -1,11 +1,11 @@
 <?php
-
 /**
  * Unit tests covering WP_REST_Users_Controller functionality.
  *
  * @package WordPress
  * @subpackage JSON API
  */
+
 class WP_Test_REST_Users_Controller extends WP_Test_REST_Controller_Testcase {
 	/**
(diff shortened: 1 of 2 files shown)


rachelbaker:
@WP-API/amigos #reviewmerge

danielbachhuber:
Related #1128. It would be nice if these roles were just an enum

tests/test-rest-users-controller.php
```diff
@@ -356,6 +356,25 @@ public function test_create_user_invalid_id() {
...
 
+	public function test_create_user_invalid_role() {
+		$this->allow_user_to_manage_multisite();
+		wp_set_current_user( $this->user );
+
+		$params = array(
+			'username' => 'maggiesimpson',
+			'password' => 'i_shot_mrburns',
+			'email'    => 'packingheat@example',
+			'role'     => 'baby',
```
rmccue: :+1: :+1: :+1:

PR #1269, opened 2015-05-21
Don't nullify `parent`

### LINKED ISSUE #1193: Nullify values when the value is 0 in the database
We're already doing this for `parent`, but we aren't doing it for some other fields:

image

We should be consistent about this.

danielbachhuber:
We don't want it to be unique

Fixes #1193


rachelbaker:
Merged #1269

PR #1270, opened 2015-05-21
Load the API much earlier

### LINKED ISSUE #764: Run on an earlier action (work without WP_USE_THEMES and avoid main query)
When `WP_USE_THEMES` is `false` no output is rendered for `/wp-json'`

Why is that? 

We're building an Angular app on top of WordPress with some plugins and this json api. The theme would be completely empty as there is no use for it.

rmccue:
Rather than running the main query, let's run just before it.

Fixes #764.

### DIFF
--- a/plugin.php
+++ b/plugin.php
@@ -334,7 +334,7 @@ function rest_api_loaded() {
 	// We're done.
 	die();
 }
-add_action( 'template_redirect', 'rest_api_loaded', -100 );
+add_action( 'parse_request', 'rest_api_loaded' );
 
 /**
  * Register routes and flush the rewrite rules on activation.


rachelbaker:
Let's see if this breaks anything.

kadamwhite:
For my own edification, what's the advantage we'll glean from this change?

joehoyle:
@kadamwhite faster API requests, and reduced load on your database server.

rmccue:
@kadamwhite Without this, the main query will always be run (essentially `SELECT * FROM wp_posts LIMIT 10`), even though the data isn't actually used. Even with object caching enabled on the site, you'll still hit the database when it could have been cached.

This fixes that, at the expense of moving before some actions/filters might have run, and breaking backwards compatibility.

Issue #1273, opened 2015-05-21
Redirecting /users/me to /users/123 breaks OAuth signature creation


AlexC:
By redirecting requests to /users/me to /users/123 (where 123 is the user ID), you're making it impossible to access this protected resource using OAuth as the URL used to create the signature is no longer the URL that is client requested.

Pull request to follow

rmccue:
See comment on #1275.

PR #1275, opened 2015-05-21
Fix issue #1273, do not redirect /users/me to /users/123

AlexC:
This breaks OAuth signatures as the URL has now changed from underneath
the client, so the provided URL no longer matches.

### DIFF
--- a/lib/class-wp-json-users.php
+++ b/lib/class-wp-json-users.php
@@ -97,22 +97,7 @@ class WP_JSON_Users {
 			return new WP_Error( 'json_not_logged_in', __( 'You are not currently logged in.' ), array( 'status' => 401 ) );
 		}
 
-		$response = $this->get_user( $current_user_id, $context );
-
-		if ( is_wp_error( $response ) ) {
-			return $response;
-		}
-
-		if ( ! ( $response instanceof WP_JSON_ResponseInterface ) ) {
-			$response = new WP_JSON_Response( $response );
-		}
-
-		$data = $response->get_data();
-
-		$response->header( 'Location', $data['meta']['links']['self'] );
-		$response->set_status( 302 );
-
-		return $response;
+		return $this->get_user( $current_user_id, $context );
 	}
 
 	/**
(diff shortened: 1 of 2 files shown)


rmccue:
This is intentional, as it's intended to be a pointer to the real user resource. The data is returned in the response body anyway.

For the most part this is fine, except for in-browser contexts where you can't stop the browser automatically redirecting you. For this reason, you can "envelope" the request, which sends a 200 response, but wraps the code and message in the body instead. Pass `&_envelope` for that one.

AlexC:
@rmccue Thanks, though the string "_envelope" does not exist anywhere in the 1.2.2 API, and nothing in the documentation to suggest its existence. Either way, this endpoint will not work with OAuth due to the reason mentioned above - and this isn't exactly RESTful.

ZackBoe:
@AlexC I'm not sure if this will help in your particular case, but using WP REST API 2.0-beta5 and the [WP OAuth Server](https://wordpress.org/plugins/oauth2-provider/) plugin I'm able to add ?_envelope to the /me endpoint and get the user object within the body

`http://local.wordpress.dev/wp-json/wp/v2/users/me?_envelope` returns  
`{
  "body": {
    "avatar_urls": {
      "24": "http://1.gravatar.com/avatar/7667c21f97b546a816e2c071d70e871b?s=24&d=mm&r=g",
      "48": "http://1.gravatar.com/avatar/7667c21f97b546a816e2c071d70e871b?s=48&d=mm&r=g",
      "96": "http://1.gravatar.com/avatar/7667c21f97b546a816e2c071d70e871b?s=96&d=mm&r=g"
    },
    "description": "",
    "id": 2,
    "link": "http://local.wordpress.dev/author/zack/",
    "name": "zack boehm",
    "url": "",
    "_links": {
      "self": [
        {
          "href": "http://local.wordpress.dev/wp-json/wp/v2/users/2"
        }
      ],
      "collection": [
        {
          "href": "http://local.wordpress.dev/wp-json/wp/v2/users"
        }
      ]
    }
  },
  "status": 302,
  "headers": {
    "Location": "http://local.wordpress.dev/wp-json/wp/v2/users/2",
    "Allow": "GET"
  }
}`

PR #1279, opened 2015-05-21
Show `show_in_rest` for exposing taxonomies, not `public`

joehoyle:
Came out of #1216

### DIFF
--- a/lib/endpoints/class-wp-rest-taxonomies-controller.php
+++ b/lib/endpoints/class-wp-rest-taxonomies-controller.php
@@ -74,7 +74,7 @@ class WP_REST_Taxonomies_Controller extends WP_REST_Controller {
 
 		$tax_obj = get_taxonomy( $request['taxonomy'] );
 
-		if ( $tax_obj && false === $tax_obj->public ) {
+		if ( $tax_obj && empty( $tax_obj->show_in_rest ) ) {
 			return false;
 		}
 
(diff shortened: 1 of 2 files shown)


joehoyle:
@WP-API/amigos #reviewmerge

picklepete:
Hi @danielbachhuber - I just had a read through #1895 and I'm curious how this change, adding `show_in_rest`, will affect existing plugins which register custom taxonomies. Is your/WP-API's position going to be that these plugins should release an update to include `show_in_rest`?

I only ask because we @camayak are actively building a WP-API integration (we currently rely entirely on XML-RPC for publishing) and noticed a small-ish roadblock with your old plugin, Co-Authors Plus. At the moment with XML-RPC we can simply fetch the "author" taxonomy and its terms without any trouble, but using the latest WP-API release [this line of code](https://github.com/WP-API/WP-API/blob/f344010c0908b663a11b58f2d0e31c7fb7abe790/lib/endpoints/class-wp-rest-taxonomies-controller.php#L75) is rejecting the taxonomy altogether.

This is obviously a quick fix, which could maybe get included in [this PR](https://github.com/Automattic/Co-Authors-Plus/pull/334), but it's just a shame that WordPress site owners will need to upgrade all plugins which `register_taxonomy` before using the REST API's taxonomies endpoint.

danielbachhuber:
> Is your/WP-API's position going to be that these plugins should release an update to include show_in_rest?

Yes, because they need to ensure their data is appropriately modeled before it's exposed in the REST API.

> noticed a small-ish roadblock with your old plugin, Co-Authors Plus.

Co-Authors Plus is introducing its own endpoints, to ensure its data is appropriately modeled https://github.com/Automattic/Co-Authors-Plus/pull/334

Issue #1223, opened 2015-05-08
Require arguments to be registered in order to be included in WP_REST_Request


danielbachhuber:
Currently we include all supplied arguments. However, this poses the opportunity for data to leak through without sanitization. We should be extra safe and only include registered arguments.

quasel:
+1 on this seems a good idea but add it in a way that we can still access not defined args if we want !

joehoyle:
Added PR https://github.com/WP-API/WP-API/pull/1281

danielbachhuber:
We discussed this quite in depth on Slack today: https://wordpress.slack.com/archives/core-restapi/p1433807913001353

danielbachhuber:
Although requiring arguments to be registered isn't a default behavior of `register_rest_route()`, I'd like to consider introducing a REST API strict mode ([Trac ticket](https://core.trac.wordpress.org/ticket/35000)) and enabling it for all of our core endpoints.

rmccue:
We're going to spin this up as a feature plugin; we have an [early beta](https://gist.github.com/rmccue/184a2329a922f6b4350d) to be turned into a proper plugin.

PR #1281, opened 2015-05-21
Only allow arguments that have been registered to be set on the Request object

joehoyle:
See #1223 - this is an experiment, don't merge yet :)

### DIFF
--- a/lib/endpoints/class-wp-rest-comments-controller.php
+++ b/lib/endpoints/class-wp-rest-comments-controller.php
@@ -855,6 +855,12 @@ class WP_REST_Comments_Controller extends WP_REST_Controller {
 			'sanitize_callback' => 'absint',
 			'type'              => 'integer',
 		);
+		$query_params['context']   = array(
+			'default'           => 'view',
+			'description'       => 'Set the context for the requet. "view" or "edit".',
+			'sanitize_callback' => 'sanitize_text_field',
+			'type'              => 'string',
+		);
 		return $query_params;
 	}
 
--- a/lib/endpoints/class-wp-rest-users-controller.php
+++ b/lib/endpoints/class-wp-rest-users-controller.php
@@ -657,6 +657,11 @@ class WP_REST_Users_Controller extends WP_REST_Controller {
 					'type'        => 'array',
 					'context'     => array( 'view', 'edit' ),
 				),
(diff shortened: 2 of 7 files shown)


joehoyle:
@WP-API/amigos feedback appreciated here :)

lib/endpoints/class-wp-rest-comments-controller.php
```diff
@@ -855,6 +855,12 @@ public function get_collection_params() {
 			'sanitize_callback' => 'absint',
 			'type'              => 'integer',
 		);
+		$query_params['context']   = array(
+			'default'           => 'view',
+			'description'       => 'Set the context for the requet. "view" or "edit".',
```
danielbachhuber: Spelling

danielbachhuber: Also, shouldn't the options be present as `enum`?

danielbachhuber:
I am :100: to the idea.

danielbachhuber:
To @rmccue for #reviewmerge

rmccue:
@joehoyle and I talked this one over, and there's some conceptual things here that we both want to continue thinking over. Will compile coherent thoughts on it, but in the meantime, we decided to punt this one to b3.

rmccue:
We're going to tone this one down a bit. Instead of requiring all arguments to be registered, we're going to require all registered arguments to have either a validation callback or a sanitisation callback (or both!). This stops people from accidentally misspelling `validate_callback`/`sanitize_callback`, but won't stop them from accessing other unregistered arguments.

This isn't a full solution that solves every problem, but we need to balance extensibility and usability here. We need people to use the request object as much as possible, and anything that encourages them to directly use `$_GET`/`$_POST` is going to hurt internal reuse.

Removed Review tag, and reassigning back to @joehoyle to switch it up.

danielbachhuber:
> We're going to tone this one down a bit. Instead of requiring all arguments to be registered, we're going to require all registered arguments to have either a validation callback or a sanitisation callback (or both!). This stops people from accidentally misspelling validate_callback/sanitize_callback, but won't stop them from accessing other unregistered arguments.

Given the infrastructure has been committed to core, this still needs to be addressed: https://core.trac.wordpress.org/ticket/34725

PR #1284, opened 2015-05-24
Sort out link relations and embedded data

### LINKED ISSUE #1139: Audit link relations
Links were always intended to follow the [IANA link relations registry](http://www.iana.org/assignments/link-relations/link-relations.xhtml#link-relations-1), but we accidentally broke this in v2 by using taxonomy slugs directly as relations.

Let's fix these up, and also rethink our approach. Rather than using custom link relations, we can restructure to match the way that Atom/RSS/etc feed discovery works by including a `type` parameter.

For example, rather than `Link: <.../meta>; rel=<http://wp-api.org/2.0/meta>`, we can do `Link: <.../meta>; rel=related; type=<http://wp-api.org/2.0/meta>`

Old-style in JSON:
```json
{
    "_links": {
        "http://wp-api.org/2.0/meta": [ {
            "href": ".../meta",
        } ]
    }
}
```

New-style in JSON:
```json
{
    "_links": {
        "related": [ {
            "href": ".../meta",
            "type": "http://wp-api.org/2.0/meta",
        } ]
    }
}
```

This should make indexing into it easier (`_links.related[0].href`), and allows us to be more flexible too. For example, something like Pods could also expose `Link: <.../pods/42/fields>; rel=related; type=<http://pods.io/1.0/fields/>` and have it grouped alongside.

(Props @rachelbaker for inspiring this.)

### LINKED ISSUE #1287: Comment schema needs embed context
The comment schema doesn't include any response fields in the `embed` context right now, which means when embedding the related data for a single post the response data is empty.

Example:

```
"_embedded": {
	...
	"replies": [
		[ 
		{ 
			"_links": {
				"self": [
					{
					"href": "http://vagrant.local/wp-json/wp/v2/comments/3"
					}
				],
				"collection": [
					{
					"href": "http://vagrant.local/wp-json/wp/v2/comments"
					}
				],
				"author": [
					{
					"embeddable": true,
					"href": "http://vagrant.local/wp-json/wp/v2/users/1"
					}
				],
				"post": [
					{
					"embeddable": true,
					"href": "http://vagrant.local/wp-json/wp/v2/posts/5"
					}
				]
			}
		}
	] ],
	...
}
```

### LINKED ISSUE #1289: Terms schema needs embed context
Similar to #1287, when embedding related data in a single post response the term data is missing.

rmccue:
This is a start to fix #1139.

Right now, this still uses the custom relation types, but next step is going to be trying switching that over.

I've also gone through and added a lot more fields to the embed context to make these links actually useful. Now you can go and get everything to render a single post nice and easily ([example post with embedding](https://gist.github.com/rmccue/1eba34f4ec6300e5513a)).

Also fixes #1289, fixes #1287

### DIFF
--- a/lib/endpoints/class-wp-rest-posts-controller.php
+++ b/lib/endpoints/class-wp-rest-posts-controller.php
@@ -1123,7 +1123,7 @@ class WP_REST_Posts_Controller extends WP_REST_Controller {
 		if ( ! in_array( $post->post_type, array( 'attachment', 'nav_menu_item', 'revision' ) ) ) {
 			$attachments_url = rest_url( 'wp/v2/media' );
 			$attachments_url = add_query_arg( 'post_parent', $post->ID, $attachments_url );
-			$links['http://wp-api.org/v2/attachment'] = array(
+			$links['http://v2.wp-api.org/attachment'] = array(
 				'href'       => $attachments_url,
 				'embeddable' => true,
 			);
@@ -1131,7 +1131,7 @@ class WP_REST_Posts_Controller extends WP_REST_Controller {
 
 		$taxonomies = get_object_taxonomies( $post->post_type );
 		if ( ! empty( $taxonomies ) ) {
-			$links['http://wp-api.org/v2/term'] = array();
+			$links['http://v2.wp-api.org/term'] = array();
 
 			foreach ( $taxonomies as $tax ) {
 				$taxonomy_obj = get_taxonomy( $tax );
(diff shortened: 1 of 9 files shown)


rmccue:
(Also, this leaves us with one non-IANA-registered link relation, `in-reply-to`, but that's over on the [Microformats wiki as a proposal](http://microformats.org/wiki/rel-in-reply-to). Despite being a wiki, the Microformats wiki is used by a few things as the canonical source, including the HTML(5) spec.)

rachelbaker:
Will also fix #1289

rmccue:
@WP-API/amigos #reviewmerge

lib/endpoints/class-wp-rest-posts-controller.php
```diff
@@ -1134,6 +1131,8 @@ protected function prepare_links( $post ) {
 
 		$taxonomies = get_object_taxonomies( $post->post_type );
 		if ( ! empty( $taxonomies ) ) {
+			$links['http://wp-api.org/v2/term'] = array();
```
rachelbaker: @rmccue Will http://wp-api.org/v2/term resolve somewhere?

PR #1285, opened 2015-05-24
Experimental - Auto route registration

joehoyle:
Given that we have established a pretty clear pattern for our controllers, as an experiment I wanted to see if we could auto register the routes based off what is implemented in each controller. It worked out pretty well, at least with the current abstraction.

The idea for this came when I was writing documentation for adding routes. By implementing `WP_REST_Controller` it's also required that one know the syntax and underlaying API of `register_rest_route` which would be nice to not _require_. Of course, child classes can always overwrite `register_routes` if required (which for example `WP_REST_Users_Controller` does.


lib/endpoints/class-wp-rest-controller.php
```diff
@@ -3,111 +3,113 @@
 
 abstract class WP_REST_Controller {
 
+	protected $route_base = '';
+	protected $id_regex    = '[\d]+';
+	protected $namespace   = '';
 	/**
 	 * Register the routes for the objects of the controller.
 	 */
 	public function register_routes() {
```
danielbachhuber: Should we break each part of this into separate methods, so you can subclass individual methods to disable and don't need to copy/pasta the internals for `register_routes()`?

lib/endpoints/class-wp-rest-controller.php
```diff
@@ -3,111 +3,113 @@
...
-	/**
-	 * Update one item from the collection
-	 *
-	 * @param WP_REST_Request $request Full data about the request.
-	 * @return WP_Error|WP_REST_Request
-	 */
-	public function update_item( $request ) {
-		return new WP_Error( 'invalid-method', __( 'Method not implemented. Must be over-ridden in subclass.' ), array( 'status' => 405 ) );
-	}
+			if ( method_exists( $this, 'get_items_permissions_check' ) ) {
```
danielbachhuber: How do we handle the magical — if this is misspelled, we expose a bunch of data?

We could force override by having the default registry return false? e.g. you always have to successfully override.

danielbachhuber:
I really like this idea. Very DRY.

danielbachhuber:
Punt on this for now: https://wordpress.slack.com/archives/core-restapi/p1449531502001344

We could come back to it once we feel more comfortable about the abstraction.

PR #1291, opened 2015-05-25
Add embed context to Comments response

rachelbaker:
Includes:

id
parent
author_name
author_url
content[rendered]
date
link
post


rmccue:
Handled in #1284 (see #1287 for discussion).

Issue #555, opened 2014-10-29
Consistent collection query arguments across the board


rachelbaker:
**Currently:**

- All over the place
- Arguments typically correspond to the arguments used in core Query classes (post_status, etc.)

**Proposed:**

- Post:
   - author (int)
   - search (string)
   - status (string)
   - before
   - after
   - modified_before
   - modified_after
   - page (int)
   - per_page (int)
- User:
   - search (string)
   - page (int)
   - per_page (int)
   - order (asc|desc)
   - orderby (?)
- Comment:
   - post (int)
- Term:
   - search (string)
   - page (int)
   - per_page (int)
- Look into arguments behind design of WP_Date_Query and ensure we’re not reinventing the wheel

danielbachhuber:
Related #924

danielbachhuber:
This is being handled in #924

PR #1302, opened 2015-05-26
Declare post collection params so they can be exposed

danielbachhuber:
See #924, #555


lib/endpoints/class-wp-rest-posts-controller.php
```diff
@@ -35,11 +35,12 @@ public function register_routes() {
 			}
```
danielbachhuber: @WP-API/amigos What should I do with legacy `get_allowed_query_vars()` ?

danielbachhuber:
@WP-API/amigos I'm not 100% sure what we decided to do with `WP_Query`-style arguments (e.g. also supporting `posts_per_page` in addition to `page`).

Given we don't support `WP_User_Query`-style arguments et al, and trying to magically support them would be a pretty massive hack, I think we should drop them for now.

rachelbaker:
> I'm not 100% sure what we decided to do with WP_Query-style arguments 

I don't know if we discussed this, but it sounds like we should.

lib/endpoints/class-wp-rest-posts-controller.php
rachelbaker: The API has always supported the public query_vars of WP_Query, included a few others, and provided a filter to add support for additional vars.  There are some public query vars that are pretty pointless (example `withcomments`) but trying to normalize all these params sounds like a headache to me.

rachelbaker:
Looks like this needs some more discussion.  @rmccue @joehoyle thoughts?

joehoyle:
When we discussed this at LoopConf, I was under the impression we settled on the following:

1. Create new query params, for example `page`, `per_page`, `slug` etc.
2. Put all current WP_Query style WordPress query names under a `filter` or `wp_query` argument. So these would be equivalent: 

`GET /wp/v2/posts?per_page=10`

`GET /wp/v2/posts?wp_query[posts_per_page]=10`

That way, we can support WordPress devs thinking the a `WP_Query` mindset, and give them the ability to construct more complex queries (though ofcourse of the public query vars that we currently use). Then have a clean top level args that will be consistent across all collections.

danielbachhuber:
> `GET /wp/v2/posts?wp_query[posts_per_page]=10`

I dig this — seems like a reasonable compromise.

rmccue:
If we're going to do it, keep the `filter` name so it's consistent with v1. That way, v2 = v1 + additional better named querying, rather than completely different.

danielbachhuber:
I dig `filter`

rmccue:
Marking as Needs Refresh, since `filter` needs to be added in.

danielbachhuber:
Closing in favor of #1793

PR #1304, opened 2015-05-28
Delete/trash consistency

### LINKED ISSUE #1271: DELETE consistency when trash isn't available for resource
Across our endpoints, we should be consistent with how trashing works. (#789)

in #789, I commented:
> Ideally, we want consistent behaviour here, so that brings up the question of "what do we do for untrashable items?" - that is, core objects that don't support being trashed. I think it's possible that down the line core will support these being trashed, so we should plan with that eventuality in mind, even if it never happens.

Along these lines, I think we should be consistent with trashing. For every resource that supports either type of DELETE, `DELETE <resource>` should always trash, and `DELETE <resource>?force=true` should always permanently delete. When trash isn't available for a given resource, the bare `DELETE` should fail to do anything; ditto if permanently delete isn't available.

This gives us room to expand if core adds trash. Take terms for example. If bare `DELETE` fails, clients will code to permanently delete. If core adds trash for terms e.g., the bare `DELETE` would change from a "not supported" response to working. Clients could then update to start using this endpoint if they want to use the trash, and still fall back on sites that don't have it by checking the status code. (Eager clients could code for this from the start so they begin supporting it immediately when it's available.)

@WP-API/amigos Thoughts? I think this is a natural extension to ensure we're consistent, which along with #1260 should make our DELETEs super-awesome-sauce.

rmccue:
Fixes #1271

### DIFF
--- a/lib/endpoints/class-wp-rest-comments-controller.php
+++ b/lib/endpoints/class-wp-rest-comments-controller.php
@@ -284,11 +284,29 @@ class WP_REST_Comments_Controller extends WP_REST_Controller {
 			return new WP_Error( 'rest_comment_invalid_id', __( 'Invalid comment ID.' ), array( 'status' => 404 ) );
 		}
 
+		/**
+		 * Filter whether the comment type supports trashing.
+		 *
+		 * @param boolean $supports_trash Does the comment type support trashing?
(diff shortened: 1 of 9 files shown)


rachelbaker:
Merged #1304

PR #1306, opened 2015-05-28
Support for featured image

vdemay:
Add support for featured image as described in create_post method

### DIFF
--- a/lib/class-wp-json-posts.php
+++ b/lib/class-wp-json-posts.php
@@ -973,6 +973,22 @@ class WP_JSON_Posts {
 				unstick_post( $post_ID );
 			}
 		}
+		
+		//featured image
+		if ( isset( $data['post_thumbnail'] ) ) {
+			$featured_image = (int) $data['post_thumbnail'];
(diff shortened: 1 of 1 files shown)


danielbachhuber:
Thanks for the pull request. We aren't taking enhancements against `master` at this time, however.

vdemay:
Thanks for the quick answer.

I'm Vincent Demay, lead engineer at Scoop.it.
We are currently working on new way to synchronise our content with WP (we encounter a lot of issues with XML-RPC communication). 

WP-API is for us the best alternative but managing the featured image is crucial for our end users. Moreover the Scoop.it/WP synchronisation is core for our business, I looked at the V2 of the api but it seems to not yet be ready for production. 

Obviously I've also tried to set _thumbnail_id meta but I get an error message "_thumbnail_id is marked as a protected field." as expected. For now we have meet with a wall. Is there a way I didn't get to set/edit the featured image via WP_API V1(without packaging our own version)?

Many thanks for your help and your quick answer.

Best,

PR #1316, opened 2015-06-01
Fix parsing of local dates in TimeStampedMixin

staylor:
In `TimeStampedMixin`, when parsing dates into UTC timestamps, `date`
and `modified` need to have their timezone offset factored into their
value. Without modifying their timestamp with UTC offset, subsequent
calls to `Backbone.Model.save()` will store the values wrong (with UTC
offset added twice).

Example: Loading a post published at 4pm EDT and calling `.save()` on
the model will change the published time to 12pm EDT.


rmccue:
@staylor Looks like you branched this off master, but are requesting merge into develop? The two have diverged, and there's stuff in master I don't think we want in develop.

rachelbaker:
Closed.  Still needs refresh.

PR #1318, opened 2015-06-01
If `phpcs` failed, hard fail

danielbachhuber:
This will get WordPress Coding Standards to fail our builds again.


danielbachhuber:
@WP-API/amigos #reviewmerge

rachelbaker:
Merged #1318 :100:

PR #1324, opened 2015-06-03
Inline docs for `global $wp_rest_server`

Shelob9:
I added inline docs to usages of `global $wp_rest_server` so its explicitly clear what the variable contains and so phpStorm can autocomplete methods of the class called using the global variable.


rachelbaker:
Thank you @Shelob9

PR #1328, opened 2015-06-07
Improve @return inline docs for rest_ensure_response()

Shelob9:
Existing docs use "mixed" as the data type for return. That does not allow autocompletion doc reference for `rest_ensure_response()` in phpStorm, which made me sad.

### DIFF
--- a/plugin.php
+++ b/plugin.php
@@ -489,7 +489,7 @@ function rest_ensure_request( $request ) {
  * immediately check for this value.
  *
  * @param WP_Error|WP_HTTP_ResponseInterface|mixed $response Response to check.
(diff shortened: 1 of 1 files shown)


Shelob9:
@rachelbaker I refreshed the patch with more details on what is returned.

rmccue:
The line needs to be re-wrapped to 80 characters to match what we have there, or put all on one line. I'm easy either way.

Shelob9:
@rmccue Should be good now.

rmccue:
Small spelling mistake here with "rerurns"

Happy to merge immediately once fixed. :)

rmccue:
Weirdly, the Scrutinizer build failed, not sure why on that.

@Shelob9 Thanks for the PR :)

PR #1332, opened 2015-06-07
Permit post terms to be queried by term order

### LINKED ISSUE #1214: Respect `term_order` for taxonomies

danielbachhuber:
But still default to ordering by `name`

Fixes #1214


rachelbaker:
Merged #1332

Issue #1218, opened 2015-05-08
Make it easier to cross call controllers by creating WP_REST_Requests


joehoyle:
Currently cross calling `get_item` etc is laborious and you need to know the route etc. Typically, a Request object does't need all that information, as the callback just wants to the arguments. For example, rather than internally creating a `WP_REST_Request` for `/taxonomy/post_tag/terms/123`, it would be nicer to just do `new WP_REST_Request( array( 'taxonomy' => 'post_tag', 'id' => 123 ) )` or something.

@rmccue thoughts on this?

danielbachhuber:
Why not a utility function?

danielbachhuber:
Another example https://github.com/WP-API/WP-API/pull/1253/files#diff-7900746e7eb16a3b53b0cb9d4070e010R315

danielbachhuber:
Argument favor of a utility function is that the request should route through the server so we're not tightly coupled on the current controller (e.g. someone has subclassed the controller).

From https://github.com/WP-API/WP-API/pull/1253/files#r30641180

danielbachhuber:
We also do a fair amount of

```
return $this->get_item( array(
	'id'      => $post_id,
	'context' => 'edit',
));
```

These should be proper request objects.

danielbachhuber:
Landed #1333, now I need to do some updates across the board

rmccue:
Let's close this one out as done in #1333; any other stuff that needs doing can have separate tickets.

danielbachhuber:
The issue isn't done. I still need to make updates across the board. I'll move this to b4.

danielbachhuber:
@joehoyle created a utility function for this in https://gist.github.com/joehoyle/e7321570525af6daeae2

danielbachhuber:
@joehoyle look what's already been committed to core — `rest_do_request()` https://developer.wordpress.org/reference/functions/rest_do_request/

Introduced in #1333, and we're using it in `WP_REST_Posts_Controller::delete_item()`, So....

danielbachhuber:
Created a core ticket to improve upon `rest_do_request()` https://core.trac.wordpress.org/ticket/35329

danielbachhuber:
Punting on the rest of this for now. While I like the pattern implied in #1991, there's concern about the performance impact of doing more work for dubious benefit.

Slack chat https://wordpress.slack.com/archives/core-restapi/p1452557078004064

PR #1333, opened 2015-06-07
Introduce `rest_do_request()` for making REST requests internally

danielbachhuber:
Routing requests through the server in a RESTful manner permits us to
de-couple controllers from one another, permitting subclassing, etc. It
also ensures all of the appropriate permission and validation checks are
run on each internal request.

See #1218

### DIFF
--- a/lib/endpoints/class-wp-rest-posts-controller.php
+++ b/lib/endpoints/class-wp-rest-posts-controller.php
@@ -340,7 +340,9 @@ class WP_REST_Posts_Controller extends WP_REST_Controller {
 			return new WP_Error( 'rest_user_cannot_delete_post', __( 'Sorry, you are not allowed to delete this post.' ), array( 'status' => 401 ) );
 		}
 
-		$response = rest_do_request( 'GET', '/wp/v2/' . $this->get_post_type_base( $this->post_type ) . '/' . $post->ID, array( 'context' => 'edit' ) );
+		$request = new WP_REST_Request( 'GET', '/wp/v2/' . $this->get_post_type_base( $this->post_type ) . '/' . $post->ID );
+		$request->set_param( 'context', 'edit' );
+		$response = rest_do_request( $request );
 
 		// If we're forcing, then delete permanently
 		if ( $force ) {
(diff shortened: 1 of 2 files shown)


danielbachhuber:
@WP-API/amigos How do you feel about this?

lib/endpoints/class-wp-rest-posts-controller.php
```diff
@@ -340,9 +340,7 @@ public function delete_item( $request ) {
 			return new WP_Error( 'rest_user_cannot_delete_post', __( 'Sorry, you are not allowed to delete this post.' ), array( 'status' => 401 ) );
 		}
 
-		$get_request = new WP_REST_Request( 'GET', rest_url( 'wp/v2/' . $this->get_post_type_base( $this->post_type ) . '/' . $post->ID ) );
-		$get_request->set_param( 'context', 'edit' );
-		$response = $this->prepare_item_for_response( $post, $get_request );
+		$response = rest_do_request( 'GET', '/wp/v2/' . $this->get_post_type_base( $this->post_type ) . '/' . $post->ID, array( 'context' => 'edit' ) );
```
danielbachhuber: It would be nice if I could pass any arbitrary path, including one prepared with `rest_url()`

plugin.php
```diff
@@ -464,6 +464,23 @@ function rest_url( $path = '', $scheme = 'json' ) {
...
 /**
+ * Make a REST request
+ *
+ * @param string $method
+ * @param string $route
+ * @param array $params Parameters to pass to the request.
+ * @return WP_REST_Response
+ */
+function rest_do_request( $method, $route, $params = array() ) {
+	global $wp_rest_server;
```
danielbachhuber: We should likely check that `$wp_rest_server` is valid and initialized.

joehoyle:
Per my Slack feedback, I'd like to see it return a `WP_Error` on an error response, so all my error checking code would be the same as other error checking code. For example: 

```PHP
$request = rest_do_requests( '/wp/v1/posts` );

if ( is_wp_error( $request ) ) {
    return $request;
}
```

PR #1345, opened 2015-06-12
Ability to specify `arg_options` in the schema to override the arguments

joehoyle:


tests/test-rest-controller.php
```diff
@@ -73,4 +73,12 @@ function test_validate_schema_format_date_time() {
 		);
 	}
 
+	function test_get_endpoint_args_for_item_schema_arg_options() {
+
+		$controller = new WP_REST_Test_Controller();
+		$args = $controller->get_endpoint_args_for_item_schema();
+
+		$this->assertEquals( false, $args['someargoptions']['required'] );
```
rachelbaker: @joehoyle can this just use `assertFalse()`?

joehoyle: I could - given that I was comparing the equality in a few fields - it kind of meant sense to keep it consistent with that.

joehoyle: Fixed in 5ac10a6

rachelbaker:
Merged #1345

PR #1348, opened 2015-06-13
Remove 201 status response for missed endpoints

rachelbaker:
Previous changes in #1142 only addressed Posts.  This PR fixes the HTTP response code for update responses in Comments, Attachments, and Users.  I reviewed all endpoint controllers and these were the only remaining places this issue still existed.


rachelbaker:
@WP-API/amigos #reviewmerge

PR #1353, opened 2015-06-16
Experimental - Auto route registration

joehoyle:
Given that we have established a pretty clear pattern for our controllers, as an experiment I wanted to see if we could auto register the routes based off what is implemented in each controller. It worked out pretty well, at least with the current abstraction.

The idea for this came when I was writing documentation for adding routes. By implementing `WP_REST_Controller` it's also required that one know the syntax and underlaying API of `register_rest_route` which would be nice to not _require_. Of course, child classes can always overwrite `register_routes` if required (which for example `WP_REST_Users_Controller` does.


joehoyle:
Vim had an old message :( re-opening

PR #1354, opened 2015-06-16
Check taxonomies exist before trying to set properties

joehoyle:
In the case where a develop has unregistered the post_tag / category taxonomy,
it currently produces PHP Notices.

### DIFF
--- a/plugin.php
+++ b/plugin.php
@@ -133,13 +133,17 @@ add_action( 'init', '_add_extra_api_post_type_arguments', 11 );
 function _add_extra_api_taxonomy_arguments() {
 	global $wp_taxonomies;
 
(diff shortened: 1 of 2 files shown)


joehoyle:
@WP-API/amigos #reviewmerge

danielbachhuber:
> In the case where a develop has unregistered the post_tag / category taxonomy,

I'll refrain from commenting about this ;)

danielbachhuber:
@WP-API/amigos anyone know why codecov fails?

rmccue:
@danielbachhuber Because there's no specific testing for this code, so it's not marked as covered. (`codecov/patch` is the coverage for just the changed lines)

joehoyle:
@WP-API/amigos #reviewmerge plz

PR #1355, opened 2015-06-17
If not namespace is provided, don't create double slashes

joehoyle:
In cases where I'm writing a full bespoke/whitelabelled API rather than adding to WordPress's routes, or them even being available, I want to not use namespaces at all.

This is probably controversial - but point being I want my api at `myserviers.com/api` for better or worse. (In this specific case I'm migrating from another API system, so I need the URLs to match the old ones as it's a drop in replacement).

In this case, it's pretty easy for us to support this by just checking if a namespace was not-empty and don't add the double slashes if so.

### DIFF
--- a/plugin.php
+++ b/plugin.php
@@ -76,6 +76,11 @@ function register_rest_route( $namespace, $route, $args = array(), $override = f
 	if ( $namespace ) {
 		$full_route = '/' . trim( $namespace, '/' ) . '/' . trim( $route, '/' );
 	} else {
+		// Non-namespaced routes are not allowed, with the exception of the main
+		// and namespace indexes. If you really need to register a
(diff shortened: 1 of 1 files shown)


rmccue:
I'd probably rather do a `_doing_it_wrong`, butttt I dunno.

rachelbaker:
I feel like this allows registered routes to skirt around namespacing their routes too easily, which is something we want to avoid.  I agree with @rmccue that we should probably display a debug message.

rmccue:
Thought: `_doing_it_wrong` in `register_rest_route`, but allow it inside `WP_REST_Server::register_route`.

joehoyle:
Given the incorrect usage capture here, I'm ok with the tests at 60% given calling it will cause a notice anyway and fail the tests!

Issue #1330, opened 2015-06-07
Deprecate rest_get_avatar_url() in favor of WP 4.2 get_avatar_url()


danielbachhuber:
From https://github.com/WP-API/WP-API/pull/1327#discussion_r31869878

It will mean bumping our minimum supported version to WP 4.2.

rachelbaker:
Done in #1361

Issue #1290, opened 2015-05-25
Add author_avatar_url to the Comment item response


rachelbaker:
Related: https://github.com/WP-API/WP-API/issues/1287#

danielbachhuber:
@rachelbaker Do we need to offer an array of sizes when we return avatar URLs? What should the default size be?

rachelbaker:
Tagging for discussion because I don't have a strong opinion on @danielbachhuber's question above.

>  Do we need to offer an array of sizes when we return avatar URLs? What should the default size be?

rachelbaker:
Punted to Beta 4

rachelbaker:
Done in #1361

PR #1361, opened 2015-06-19
Return an array of avatar urls in different sizes for Comments and Users

rachelbaker:
- Add rest_get_avatars and deprecate rest_get_avatar
- Update Comment controller to use rest_get_avatars 
- Update Users controller to use rest_get_avatars

### DIFF
--- a/lib/endpoints/class-wp-rest-comments-controller.php
+++ b/lib/endpoints/class-wp-rest-comments-controller.php
@@ -683,6 +683,16 @@ class WP_REST_Comments_Controller extends WP_REST_Controller {
 	 * @return array
 	 */
 	public function get_item_schema() {
+		$avatar_properties = array();
+
+		$avatar_sizes = rest_get_avatar_sizes();
+		foreach ( $avatar_sizes as $size ) {
+			$avatar_properties[ $size ] = array(
+				'description' => 'Avatar URL with image size of ' . $size . ' pixels.',
+				'type'        => 'uri',
+				'context'     => array( 'embed', 'view', 'edit' ),
+			);
+		}
 
 		$schema = array(
 			'$schema'              => 'http://json-schema.org/draft-04/schema#',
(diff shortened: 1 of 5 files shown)


rachelbaker:
Fixes #1330 and #1290

We now return an array of avatar urls that are keyed on the image size.

Example:

```
  "avatar_urls": {
    "24": "http://0.gravatar.com/avatar/06e92fdf4a9a63441dff65945114b47f?s=24&d=mm&r=g",
    "48": "http://0.gravatar.com/avatar/06e92fdf4a9a63441dff65945114b47f?s=48&d=mm&r=g",
    "96": "http://0.gravatar.com/avatar/06e92fdf4a9a63441dff65945114b47f?s=96&d=mm&r=g"
  },
```

lib/endpoints/class-wp-rest-comments-controller.php
```diff
@@ -700,9 +700,9 @@ public function get_item_schema() {
 					'type'         => 'integer',
 					'context'      => array( 'view', 'edit', 'embed' ),
 					),
-				'author_avatar_url' => array(
-					'description'   => 'Avatar URL for the object author.',
-					'type'          => 'string',
+				'author_avatar_urls' => array(
+					'description'   => 'Avatar URLs for the object author.',
+					'type'          => 'array',
 					'format'        => 'uri',
```
rachelbaker: @danielbachhuber can I still use the `uri` format with the `array` type?

rachelbaker:
@WP-API/amigos #reviewmerge

lib/endpoints/class-wp-rest-comments-controller.php
danielbachhuber: You're actually returning an object now, so it would be an object with specific properties

rachelbaker: http://media.giphy.com/media/8WdsK61D9YOOc/giphy.gif

rachelbaker:
Hold off on a merge, I need to fix the schema.

rachelbaker:
@WP-API/amigos #reviewmerge

rmccue:
:+1:

PR #1372, opened 2015-06-28
Add a note mentioning WordPress Trunk should be used.

joehoyle:
As we are using features / patches recently committed to WordPress Core, we
should let people know they should be using trunk.


joehoyle:
@WP-API/amigos #reviewmerge

PR #1393, opened 2015-07-08
Adds cast to integer in checking object metadata

saulopaiva:
When deleting or updating post meta, the verification of meta
belonging to object was failing because of "string - integer"
comparison.


rmccue:
`$id` already has an integer cast a few lines above, so not sure how this could actually help anything?

saulopaiva:
In `update_meta()` method that does not really help, but the `delete_meta()` method does not perform this cast and failure.

It was added in two to standardize with respect to `get_meta()`, where this cast is made.

rmccue:
Ah, `delete_meta` makes sense, I missed that in the GitHub UI.

Can you remove the `update_meta` bit please? :)

rachelbaker:
Merged #1393

PR #1397, opened 2015-07-12
Remove permissions checks from get users so we can show public users

joehoyle:
Using the new WordPress `WP_User_Query` `has_published_posts` flag we can show only public users that would otherwise be exposed via authors on public posts on the `/wp/v2/users` endpoint.

See #839

### DIFF
--- a/lib/endpoints/class-wp-rest-users-controller.php
+++ b/lib/endpoints/class-wp-rest-users-controller.php
@@ -15,7 +15,6 @@ class WP_REST_Users_Controller extends WP_REST_Controller {
 			array(
 				'methods'         => WP_REST_Server::READABLE,
 				'callback'        => array( $this, 'get_items' ),
-				'permission_callback' => array( $this, 'get_items_permissions_check' ),
 				'args'            => $query_params,
 			),
 			array(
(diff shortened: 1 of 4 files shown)


joehoyle:
This is failing because we run the unit tests against the latest stable release of WordPress, not trunk - an this PR requires the recent patch that was committed to core.

rmccue:
Fixing the tests up here.

rmccue:
@joehoyle Testing fixed up here. You happy with the PR and ready to merge it?

joehoyle:
@rmccue thanks for that, yup I'm ready for a merge

lib/endpoints/class-wp-rest-users-controller.php
```diff
@@ -92,6 +91,10 @@ public function get_items( $request ) {
 		$prepared_args['orderby'] = $orderby_possibles[ $request['orderby'] ];
 		$prepared_args['search'] = $request['search'];
 
+		if ( ! current_user_can( 'list_users' ) ) {
+			$prepared_args['has_published_posts'] = true;
```
rmccue: Should we only include posts marked as `show_in_rest`? Internally, WP grabs types with `public => true` here.

joehoyle: The thought was this would only reflect users who were already exposed via the RSS feeds, which could be any public post type, though I guess that's not a guarantee either. `show_in_rest` might be safer, or literally `post_type=post`

rmccue: Ah, good point on that. I'm fine with all then. :)

PR #1398, opened 2015-07-12
Fix adding alt text to attachments on create

joehoyle:
This functionality was only available in `update` for one reason or another.
Preusming this is an oversite.


joehoyle:
@WP-API/amigos #reviewmerge

PR #1399, opened 2015-07-12
Sanitize pass on all controllers args

joehoyle:
This is a sweep for all the args to make sure we are snitizing where needed. The basic principle is sanitize evertyhing, however the shema does a bunch of stuff for us so we don't need to sepficy it in every case, such as:

- the schema is marked up as an `enum` as we have chcking to make sure values are only these
- the `format` is set in the schema whereby we have validation, for example, `date-time`
- the type is `integer`

Also, for `content` and `title` in the `wp/v2/posts` endpoint does not have any, as `wp_insert_post` handles that depending on whatever permissions the user has, so rather than replicating this, we just pass it through.

### DIFF
--- a/lib/endpoints/class-wp-rest-attachments-controller.php
+++ b/lib/endpoints/class-wp-rest-attachments-controller.php
@@ -103,7 +103,7 @@ class WP_REST_Attachments_Controller extends WP_REST_Posts_Controller {
 		$data = $response->get_data();
 
 		if ( isset( $request['alt_text'] ) ) {
-			update_post_meta( $data['id'], '_wp_attachment_image_alt', sanitize_text_field( $request['alt_text'] ) );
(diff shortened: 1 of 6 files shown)


joehoyle:
Doh, need to merge `develop`

rachelbaker:
@joehoyle Any reason you left out the comments controller?

rachelbaker:
@joehoyle other than my comment about the comments controller being left out the only other open item I can see is we should also be setting the default values where possible.  We can merge this PR, and open issues for both of those items.  Just wanted to run it past you to see what you think.

joehoyle:
@rachelbaker ah I didn't do comments because I wanted to check if there was a reason we hadn't switch to using `get_endpoint_args_from_item_schema`. I'm presuming not so I'll make that transition. 

I think another pass / ticker for defaults would be good.

rachelbaker:
Merged #1399

Issue #1280, opened 2015-05-21
User should have a role, or roles - not both


joehoyle:
Currently on the response of a user, we return array of roles they have. However, on update / create, the client passed the `role` argument. We should either decide that users only have on role as far as the API concerned, or they have multiple, and you can specify `roles` when updating the resource.

rmccue:
This is a good point. I think the key problem is that WP treats it like you only have one role a lot of the time, but obviously supports multiple. The 80% use case is going to be a single role because of this, but we need to support multiple. For the flexibility, we might not be able to optimise for single. :disappointed:

rachelbaker:
Punted to Beta 4

dd32:
Exposing `roles` on read seems correct, because a user can certainly have multiple roles.

However, when it comes to setting it, I feel there it may be a case where you want `role=custom_role mode=append` and also `role=administrator mode=replace` To extend the multiple-role idea though once it's wide-spread, you might want to do `roles=[ administrator, custom_role ] mode=replace` or `roles=[ custom_role, custom_role2 ] mode=append`.

It's worth noting, that WordPress also support per-user capabilities (outside of roles - and I believe as both a "give this user this cap" and "this user cannot perform this cap") which is even more rarely used. It could be worth saying that if you want to use the advanced capabilities/roles of the capabilities system, you should rely upon a plugin which exposes those details.

joehoyle:
We're going to support multiple roles by just changing the key to `roles` which is an array. This makes it more consistent and works for all use cases, rather than trying to support the "single role" use case.

PR #1400, opened 2015-07-12
Add `role` to schema for users

joehoyle:
This way we get to supply the clients with an `enum` of roles, also the field
for updating / creating a user is `role`. The maining question is whether to
remove support for multiple roles or not. Currently it's half of both.

Adding this PR to get the discussion moving!


rmccue:
> The maining question is whether to remove support for multiple roles or not. Currently it's half of both.

#1280 for discussion about this issue.

Looks good to me. :+1:

Issue #816, opened 2015-01-28
Options Endpoints


joehoyle:
Given that the API will be used for more than just WordPress data objects, I thought it would be good to discuss how we would implement Options Endpoints and then if all goes well, have a go an implementing it. This is something we'd need in say, supporting the WP-Admin via the API. Option ID seems like something we'd want to remove too.

### Abstraction

I'm not currently sure how much abstraction we would want to provide for options. I might not be realistic to get a schema driven interface for options (but it would be cool), as the underlying core api is so loose and lenient. It would seem likely that we would want to support all the core options in some more meaning full way however. For example, take an option that has a date stored in format `x`, we'd presumably want standard date formats across the API and would therefore perhaps transforms these things. We'd only be able to do that on inbuilt options as we'd need to define it.

### Types

An option value can be of arbitrary type, which means we can't do much to document and enforce types. The type of a single option can change, say from a string to an int. It would be nice to support all the inbuilt options in with validation etc, but this would require a pretty hefty array of hardcoded schema for options that doesn't currently exist in core. For example, do we allow updating `siteurl` to `true`? 

Autoloading is something that would be nice to hide away, but not sure if there is valid use cases for exposing it.

### Permissions

It seems all options are managed under the `manage_options` capability - though I'm not sure if there are meta caps on that, doesn't seem so. However, core allows modification of the required capability on a per-subpage basis via `option_page_capability_{$option_page}`. 

### Routes

I considered including the option group name or something in the route, but I think that's probably not a good idea now. It seems that just one at `/wp-json/options/$option_name` would be enough.

```
GET     /wp-json/options/siteurl
DELETE  /wp-json/options/siteurl
POST    /wp-json/options/siteurl
```

### Object Schema

Based of information we can get, I think we'd need more information in an object that just the value it's self as the object.

```
{
    name: "siteurl",
    value: mixed,
    autoload: bool
}
```

Alternatively, we could just have the value as the object it's self, though that exposing issues with arrays etc being the direct response.

### Schemas

The options may not initially seem like they are schema based, however a bunch of the stuff proposed in #759 would be pretty cool - especially around defining enums etc. With options that are arrays/objects etc, we could use that schema to markup the full object in the option value too. `wp-admin/options.php` really sucks because it has no inference of the data types etc of the options. It would be pretty cool if we could get all the goodness proposed in #759 for options (if said path is chosen).

[...]

PR #1401, opened 2015-07-12
Add options route.

Shelob9:
This is designed to satisfy #816 

This is as complete of a first pass as I could make, but there are a few things I'm unsure of:
- Response formatting. Need a consistent response, especially in create/update so consumers are provided with the ID of the option.
- Should there be an endpoint to discover option ID, by name? Seems tricky for me as an enduser to work by ID?
- Initial whitelist of options? I just allowed three basic ones so I could test.
- How to add to whitelist. I made a filter for it, but maybe something like `register_api_field()` is called for.
- Right now the update/create endpoints do not sanitize value. I don't think we can count on register setting for that. Maybe, when options are whitelisted a sanitization callback should be provided?
- I wrote two protected methods in the class to get option name by option ID, and get option ID by option name. I did this as I couldn't find a function in core to do so. Did I miss one? Also, should these be exposed publicly or put in functions out of the class? Would make them easier to test.
- Tests... Yah I wrote some of those, no idea whether they are sufficient, sensible, etc.

### DIFF
--- a/lib/endpoints/class-wp-rest-options-controller.php
+++ b/lib/endpoints/class-wp-rest-options-controller.php
@@ -1,0 +1,394 @@
+<?php
+
+class WP_REST_Options_Controller extends WP_REST_Controller {
+	protected $base = 'options';
+
+	/**
+	 * @var array
+	 */
+	protected $allowed;
+
+
+	public function __construct( ) {
+
+		$this->set_allowed();
+	}
+
+	/**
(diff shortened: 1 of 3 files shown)


danielbachhuber:
Thanks for the PR, @Shelob9. Closing this for now in favor of #816, and we'll come back to it later should we decide to go down this route.

PR #1411, opened 2015-07-21
Correct the error message returned when user does not have permission…

danielpunkass:
… to edit a post.


rachelbaker:
Merged #1411 

Thank you @danielpunkass

PR #1412, opened 2015-07-21
Add permission check for retrieving the /posts endpoint when context=…

danielpunkass:
…edit

By extension of the fact that getting any individual post yields a forbidden context error when the context=edit and the user is not authorized, the user should also not be permitted to list any post items when unauthorized.

### DIFF
--- a/lib/endpoints/class-wp-rest-posts-controller.php
+++ b/lib/endpoints/class-wp-rest-posts-controller.php
@@ -39,6 +39,7 @@ class WP_REST_Posts_Controller extends WP_REST_Controller {
 			array(
 				'methods'         => WP_REST_Server::READABLE,
 				'callback'        => array( $this, 'get_items' ),
+				'permission_callback' => array( $this, 'get_items_permissions_check' ),				
 				'args'            => $posts_args,
 			),
 			array(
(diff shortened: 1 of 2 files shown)


rachelbaker:
@danielpunkass Great catch that we were missing this permission check.  I don't know if `edit_posts` is the correct permission check here, though.  

@WP-API/amigos Any suggestions for the permission check we should be using here?  I was thinking something like `current_user_can( $this->post_type->cap->create_posts );`

danielbachhuber:
> Any suggestions for the permission check we should be using here? I was thinking something like current_user_can( $this->post_type->cap->create_posts );

Why not `current_user_can( $this->post_type->cap->edit_posts );` ?

rachelbaker:
> Why not current_user_can( $this->post_type->cap->edit_posts ); ?

WFM

danielpunkass:
Ah, am I reading this correctly that it would tie the permissions to e.g. whether the user has the ability to edit pages, or a custom post_type? I guess $this->post_type is set as a consequence of the specific endpoint e.g. /posts -> "post", etc? Sounds good to me, I'll update the pull request later tonight probably.

rachelbaker:
Merged #1412

PR #1415, opened 2015-07-22
Send schema in OPTIONS requests and index

### LINKED ISSUE #1222: Return schema and endpoint args in body of OPTIONS request

### LINKED ISSUE #1305: Add schema data to the index(es)
Split from #1262. The plan is to make the args/schema also available from the index under a certain context (`help` or something) to facilitate things like API consoles.

rmccue:
Fixes #1222, fixes #1305.

This adds the schema to the body of the OPTIONS request in the following format:
```json
{
    "schema": {
        "$schema": "http://json-schema.org/draft-04/schema#",
        "title": "post",
        ...
    }
}
```

Wrapped it here, because we should return argument data at some point as well.

It's also available in both the main index and namespace indexes by setting the `context` to `help`. I picked this pretty arbitrarily, so we can pick something else if anyone has strong feelings:
```json
{
    "routes": {
        "/wp/v2/posts/{id}": {
            "namespace": "wp/v2",
            "schema": {
                "$schema": "http://json-schema.org/draft-04/schema#",
                "title": "post,
                ...
            }
        }
    }
}
```


danielbachhuber:
> This adds the schema to the body of the OPTIONS request in the following format:

What about collection params?

rmccue:
> What about collection params?

Not sure on what to do on that, so still reading the spec to see if they deal with that. Technically, the data we're returning doesn't match the schema, since it's a list of it, and will actually be further different when we change the collection format.

rmccue:
Added the schema to the collection OPTIONS responses as well; although the returned data doesn't exactly match the schema (since it's wrapped in a list), the convenience of having the schema available without needing an ID equals it out.

@WP-API/amigos #reviewmerge

joehoyle:
Discussed briefly with @rmccue about the fact now an `OPTIONS` request was used to determine Allowed Methods, which it still does - however using that method now will also give the client a (potentially wanted) larger payload.

PR #1420, opened 2015-07-24
Move /posts WP_Query vars to `filter` param

joehoyle:
Currently, `/wp/v2/posts` allows WP_Query params and a few of our own
"per_page" etc. This is confusing and inconsistant. I think we discussed a
while back that we'd put all the WP_Query (internal) named collection params
under a `filter` param (inspired by v1) and implement standardized collection
params for posts that's inline with the params for `terms`, `users` and
`comments`.

Feedback on this before I go further I think :)


rmccue:
:+1: on the concept. Right now, having the raw WP_Query vars is a bit confusing.

danielbachhuber:
I think I started but never completed this in #1302. It might be worthwhile to just finish that branch

rachelbaker:
Why only do this for `/posts/`?  Shouldn't we do this for `/users/`, `/comments/`, and '/taxonomies/' as well?

lib/endpoints/class-wp-rest-posts-controller.php
```diff
@@ -89,12 +84,17 @@ public function register_routes() {
...
 	 */
 	public function get_items( $request ) {
-		$args = (array) $request->get_params();
-		$args['post_type'] = $this->post_type;
-		$args['paged'] = $args['page'];
-		$args['posts_per_page'] = $args['per_page'];
+		$args                   = array();
+		$args['post_type']      = $this->post_type;
+		$args['paged']          = $request['page'];
+		$args['posts_per_page'] = $request['per_page'];
```
rachelbaker: I always thought `paged` and `posts_per_page` being outside of the `filter` argument was weird.  If we are going to move the WP_Query args to the filter array we should move them all!

joehoyle: @rachelbaker I did move them all :) this is mapping `page` and `per_page` to WP_Query vars.

rmccue:
> Why only do this for /posts/?

As @joehoyle noted, "Feedback on this before I go further I think :)", so probably just proving it out with posts first.

lib/endpoints/class-wp-rest-posts-controller.php
danielbachhuber: `filter` is only for WP_Query arguments. `page` and `per_page` are our new, consistent with the rest of the controllers, arguments.

lib/endpoints/class-wp-rest-posts-controller.php
```diff
@@ -89,12 +84,17 @@ public function register_routes() {
 	 * @return WP_Error|WP_REST_Response
 	 */
 	public function get_items( $request ) {
-		$args = (array) $request->get_params();
-		$args['post_type'] = $this->post_type;
-		$args['paged'] = $args['page'];
-		$args['posts_per_page'] = $args['per_page'];
+		$args                   = array();
+		$args['post_type']      = $this->post_type;
```
rachelbaker: @joehoyle we should probably unset `post_type` here

rmccue:
> Why only do this for /posts/?

One problem with not doing it for the others is that we don't know which variables can be public, since there's no analog to `WP::$public_query_vars` that I can see for the others.

I'm going to merge this for now, since the others already use our custom mapping. If we want to add the ability to get at the underlying `WP_*_Query` args, we can handle that later for each. This PR simply brings `/posts` in line with our custom query args, while maintaining `filter` for full access.

joehoyle:
Looks good to me

PR #1421, opened 2015-07-24
Add the ability to assign more than one term at once to a post

joehoyle:
This is currently difficult to do, as you have to send one request for each
associtation. Instead, this PR makes it possible to POST to
`/wp/v2/posts/123/terms/category` with `{ ids: [1,2,4] }`.

We could also add a `names` param to assign multiple terms via Name, which is
an often requested feature we don't currently support.

### DIFF
--- a/lib/endpoints/class-wp-rest-posts-terms-controller.php
+++ b/lib/endpoints/class-wp-rest-posts-terms-controller.php
@@ -26,6 +26,19 @@ class WP_REST_Posts_Terms_Controller extends WP_REST_Controller {
 				'permission_callback' => array( $this, 'get_items_permissions_check' ),
 				'args'                => $query_params,
 			),
+			array(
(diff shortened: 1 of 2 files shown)


TimothyBJacobs:
Why use an `append` keyword instead of using the differences between the HTTP Verbs `PUT` and `PATCH`?

rmccue:
> Why use an `append` keyword instead of using the differences between the HTTP Verbs `PUT` and `PATCH`?

Supporting PATCH is a pretty major undertaking, since PATCH isn't just a partial update. We'd need to support something like JSON Patch fully to do it. (I don't think that's a bad idea, but it does mean a fair amount of work.)

See also http://williamdurand.fr/2014/02/14/please-do-not-patch-like-an-idiot/

Re: the main issue, on my list to look at and review. :)

danielbachhuber:
Problem solved in #2063

Issue #1418, opened 2015-07-23
Move Comment Controller args santization and defaults into schema


rachelbaker:
This is a continuation of #1399.

PR #1422, opened 2015-07-24
Power the Comments endpoint args from the schema, like everything else

joehoyle:
The previous args had defaults of empty strings for `wp_allow_comment`, 
however I see this as an incorrect use of the abstraction. Args should not
have defaults just to please an underlaying function that expects those
values.

This also needed to move into `create_item` as the same default can not be
used for the `update_item` endpoint args.


rachelbaker:
Intended to fix #1418

rachelbaker:
@joehoyle Thank you!  :tada:

PR #1426, opened 2015-07-24
Ensure the REST API URL has a trailing slash

danielpunkass:


tests/test-rest-plugin.php
```diff
@@ -250,4 +250,22 @@ public function test_add_extra_api_taxonomy_arguments() {
...
+	/**
+	 * The get_rest_url function should return a URL consistently terminated with a "/",
+	 * whether the blog is configured with pretty permalink support or not.
+	 */
+	public function test_rest_url_generation() {
+		$original_structure = get_option( 'permalink_structure' );
+
+		// In pretty permalinks case, we expect a path of wp-json/ with no query
+		update_option( 'permalink_structure', '/%year%/%monthnum%/%day%/%postname%/' );
+		$this->assertEquals( "http://example.org/wp-json/", get_rest_url() );
```
rmccue: This string should be single-quoted (causing tests to fail with double).

tests/test-rest-plugin.php
```diff
@@ -250,4 +250,22 @@ public function test_add_extra_api_taxonomy_arguments() {
...
+
+		// In non-pretty case, we get a query string to invoke the rest router
+		update_option( 'permalink_structure', '' );
+		$this->assertEquals( "http://example.org/?rest_route=/", get_rest_url() );
```
rmccue: Ditto here.

tests/test-rest-plugin.php
```diff
@@ -250,4 +250,22 @@ public function test_add_extra_api_taxonomy_arguments() {
...
+
+		update_option( 'permalink_structure', $original_structure );
```
rmccue: I think the option should be restored by PHPUnit's built-in setup/teardown functions, so shouldn't need this.

rmccue:
Oops, neglected to merge this previously. Thanks a bunch! :banana:

Issue #1395, opened 2015-07-11
Taxonomy term IDs not the same as what Wordpress actually wants?


jdolan:
I'm seeing some very strange behavior when creating and then associating terms via the REST API:

 * The API will create the new term as requested and return a well-formed response including the new term ID.
 * Attempting to associate the new term using the returned ID to an existing post will succeed as well (HTTP 201).
 * However, when viewing the post in WP-Admin, some of the term associations are either missing or entirely wrong.

What's really strange is that I can import a term via the API and then go search for it in WP-Admin as well. The ID that I got back in my REST response is NOT the term ID that Wordpress is using to render the Edit form for the term. That seems a red flag to me, but I'm not sure how or why it's happening.

Attached are screenshots of importing the term, associating the term, searching for the term against the REST API -- these are all in agreement with each other.

![screen shot 2015-07-11 at 1 50 26 pm](https://cloud.githubusercontent.com/assets/643118/8634836/5c4406b4-27d7-11e5-8569-b9565d331183.png)
![screen shot 2015-07-11 at 2 05 33 pm](https://cloud.githubusercontent.com/assets/643118/8634838/697a5b80-27d7-11e5-9775-fd90a95aa364.png)
![screen shot 2015-07-11 at 2 19 15 pm](https://cloud.githubusercontent.com/assets/643118/8634851/fc4c0ff8-27d7-11e5-88a8-08f046b292e3.png)
![screen shot 2015-07-11 at 1 53 00 pm](https://cloud.githubusercontent.com/assets/643118/8634840/784d96f4-27d7-11e5-9c90-78514d77fd95.png)

The last two screenshots are from the WP-Admin, and they clearly show that "Women's Tennis" is not associated to the Post, and the term ID is also different (27564 != 27580).

![screen shot 2015-07-11 at 1 56 15 pm](https://cloud.githubusercontent.com/assets/643118/8634841/7cb7723c-27d7-11e5-8371-cb4e8fb09216.png)
![screen shot 2015-07-11 at 1 56 52 pm](https://cloud.githubusercontent.com/assets/643118/8634842/7fa41388-27d7-11e5-8bad-7dfa14c07241.png)

Any idea what's going wrong here?

danielbachhuber:
WP-API exposes the `term_taxonomy_id`, not the `term_id`. WordPress core still uses the latter in URLs.

jdolan:
Okay, that's fine. But why are the associations not persisting correctly?

[...]

PR #1430, opened 2015-07-25
Resolves two issues with term creation and term assocaition

jdolan:
* When the client attempts to create a term that already exists, we should return to them the `term_taxonomy_id` as part of the error data, NOT the `term_id`. The `term_id` is useless and arguably dangerous to the client, since they may attempt to use it elsewhere (ahem ;)).

 * When associating terms, we must resolve the `term_id` from the client-provided `term_taxonomy_id`. Wordpress core uses the `term_id` to make the association.

### DIFF
--- a/lib/endpoints/class-wp-rest-posts-terms-controller.php
+++ b/lib/endpoints/class-wp-rest-posts-terms-controller.php
@@ -126,8 +126,8 @@ class WP_REST_Posts_Terms_Controller extends WP_REST_Controller {
 		if ( is_wp_error( $is_request_valid ) ) {
 			return $is_request_valid;
 		}
-		
-		$term = get_term_by('term_taxonomy_id', $term_id, $this->taxonomy);
+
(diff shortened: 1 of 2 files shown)


rmccue:
Thanks for the pull request! Good catch on the issue!

The tests are [currently failing here](https://travis-ci.org/WP-API/WP-API/jobs/72593787#L205) due to your code not matching the WP coding standards. Can you fix these up please? :)

jdolan:
All set. Would you mind re-opening #1395 until this is resolved so that other folks experiencing the same problem might find the solution?

rmccue:
This is missing a bit of test coverage on the new lines, but I'm not sure there's a way to test this with the latest versions of WP, since IIRC they don't let you create a term where `term_taxonomy_id != term_id`.

With that considered, merging. Thanks for the PR!

jdolan:
:+1:

PR #1431, opened 2015-07-26
Improve inline docs for register_api_field

Shelob9:
Added @global for $wp_rest_additional_fields
Added array params for $args
Corrected capitilization of WP_Error in @return


rachelbaker:
Committed in 0bc8cf74875e5ab422083f7f376841c9bc0829a4

PR #1432, opened 2015-07-26
@return inline docs for rest_mysql_to_rfc3339()

Shelob9:


Shelob9:
@rachelbaker Updated, good catch.

rachelbaker:
Merged #1432

PR #1433, opened 2015-07-26
Add missing inline docs for WP_REST_Server class constants

Shelob9:


lib/infrastructure/class-wp-rest-server.php
```diff
@@ -15,22 +15,85 @@
  * @package WordPress
  */
 class WP_REST_Server {
+		/**
```
rachelbaker: @Shelob9 looks like the indentation here is off?

lib/infrastructure/class-wp-rest-server.php
```diff
@@ -15,22 +15,85 @@
...
+	 * GET transport method
```
rachelbaker: @Shelob9 mind adding a period at the end of these lines?

Shelob9: Sure because you are right standards-wise, even though most of them are not complete sentences and therefore do not deserve a period.

rmccue: FWIW, totally agree on the trailing period, but @DrewAPicture disagrees. ;)

PR #1435, opened 2015-07-26
Don't expose non-public user information in the collection

rmccue:
Just now in #1397, we added the ability to list users with published posts at `/wp/v2/users`. Problem is, right now, it's exposing *all* user data (`context=view`), but users should only be able to access `context=embed`

Merging now to avoid security issues, but we can fix up the handling later.

cc @joehoyle


rmccue:
Oh, also, forgot to mention! Major props to @Shelob9 for noticing this one and notifying me immediately. :sparkles:

joehoyle:
Doh! Apologies for that

Issue #1436, opened 2015-07-27
Updating a user works, but throws error


Shelob9:
It appears that authenticated PUT and POST requests to the `wp/v2/user/<id>` endpoint update the user, but create an error.

Here is a GIF of me updating a user's email (I'm using basic auth). I get a 404 error "No route was found matching the URL and request method" but when check in the admin, the user's email has changed.

![edit-error](https://cloud.githubusercontent.com/assets/1994311/8897365/178e7aba-33e2-11e5-80f2-a344db968ad3.gif)

rachelbaker:
@Shelob9 can you provide the HTTP response headers and response code you receive when getting the `rest_no_route` error?  For both POST (new) and PUT (update) requests, if possible.

danielbachhuber:
I can't reproduce with `PUT http://wordpress-develop.dev/wp-json/wp/v2/users/1 name==foo`

PR #1441, opened 2015-07-28
Don't set the Location header in update responses

rachelbaker:
This should have been removed as part of #1348.  There is no need to set the Location header in update requests, because the response location isn't changing.

I found this while looking into #1436, and noticed that my response HTTP code when editing a user was `302` instead of the expected `200`.


rachelbaker:
@WP-API/amigos #reviewmerge

PR #1442, opened 2015-07-29
Fix double slashes in HAL link urls

rachelbaker:
HAL links and embedding are broken on sites with pretty permalinks as a result of #1426


rachelbaker:
@rmccue This fixes the `_embed` issue if pretty permalinks are enabled.  
BUT if permalinks are set to the default structure, query parameters don't work.  As an example: http://example.com/wp-json/wp/v2/posts?posts_per_page=4 returns a 301 with a Location Header of http://example.com/wp-json/wp/v2/posts/?posts_per_page=4  

Any suggestions?

Issue #1381, opened 2015-07-02
Taxonomies not discoverable due to "rest_base" prefix


jdolan:
Some taxonomies are treated specially in the API, such that they have a `rest_base` endpoint prefix that is never published via the `taxonomies` endpoint. Example: Tags.

The `taxonomies` response for Tags looks like this:

    {
    "name": "Tags",
    "slug": "post_tag",
    "description": "",
    "labels": {
        "name": "Tags",
        "singular_name": "Tag",
        "search_items": "Search Tags",
        "popular_items": "Popular Tags",
        "all_items": "All Tags",
        "parent_item": null,
        "parent_item_colon": null,
        "edit_item": "Edit Tag",
        "view_item": "View Tag",
        "update_item": "Update Tag",
        "add_new_item": "Add New Tag",
        "new_item_name": "New Tag Name",
        "separate_items_with_commas": "Separate tags with commas",
        "add_or_remove_items": "Add or remove tags",
        "choose_from_most_used": "Choose from the most used tags",
        "not_found": "No tags found.",
        "menu_name": "Tags",
        "name_admin_bar": "post_tag"
    },
    "types": [
        "post"
    ],
    "show_cloud": true,
    "hierarchical": false
    }

An intelligent guess might use the `slug` as the Tags endpoint prefix, but alas, Tags use a hard-coded `rest_base` property ("tag") instead. Category also has a `rest_base`, but through sheer luck it is the taxonomy's slug ("category").

This means there's literally no way to logically map the Tags taxonomy to the Tags endpoints. Yes, the routes are published at the API index, but that's simply a flat listing of "here's a bunch of URLs that probably have some data." There's absolutely no logical association between the Taxonomy and its endpoint URLs.

The `rest_base` property should either be included in the Taxonomy response, or it should be dropped in favor of simply using the slug -- since that's precisely what slugs are for. I'm not sure what `rest_base` is accomplishing besides breaking the alignment of data.

jdolan:
In addition, the post-terms endpoints _DO_ use the taxonomy `slug`:

`POST` -> `/wp/v2/posts/${post_id}/terms/post_tag/${term_id}`

This is just more motivation to do away with `rest_base` if you ask me.

rachelbaker:
@jdolan Thank you for the feedback on the `rest_base` oddities.

@danielbachhuber Any objection to including the `rest_base` in the Taxonomy response?  I know how you dislike the `post_tag` slug :)

[...]

PR #1447, opened 2015-07-31
Fix post taxonomy terms link urls

### LINKED ISSUE #1383: Getting terms by post id doesn't work
When calling, `wp/v2/posts/1` (for instance), the following links are returned in the JSON response:

    "_links": {
        "http://v2.wp-api.org/term": [
            {
                "embeddable": true, 
                "href": "http://.../wp/v2/terms/category?post=1", 
                "taxonomy": "category"
            }, 
           // ...
        ]
    }

But when calling `wp/v2/terms/category?post=1`, all the available categories are returned instead of just the ones of post of id `1` in this example.

rachelbaker:
Fixes #1383

This was missed in #1216, where the routes for the Taxonomies and Terms for a Post moved.


rachelbaker:
@WP-API/amigos #reviewmerge

lib/endpoints/class-wp-rest-posts-controller.php
```diff
@@ -1155,13 +1155,11 @@ protected function prepare_links( $post ) {
 				}
 
 				if ( 'post_tag' === $tax ) {
```
joehoyle: We should remove this special case, I can pick up if you want @rachelbaker

rachelbaker: @joehoyle thank you, please do.

rachelbaker:
Merged #1447

lib/endpoints/class-wp-rest-posts-controller.php
```diff
@@ -1154,13 +1154,8 @@ protected function prepare_links( $post ) {
 					continue;
 				}
 
-				if ( 'post_tag' === $tax ) {
```
jdolan: Is there going to be some alignment on "tag" (`rest_base`) vs "post_tag" (`slug`)? I get that "post_tag" is "ugly," but who really cares? The API should be data-driven, and not make special cases for personal preferences. The more the API special-cases "tag", the less consistent it becomes. Please, I beg you, use `$tax->slug` everywhere.

joehoyle: @jdolan this issue is about using the rest base, so your concern is not specifically here. However, as far as the API exposes data, you'll only ever see `tag`, not a mix - so I don't know why it's a big issue. Tags are not just for posts, so the name is confusing, that's my 2c anyway. This change just fixes one small inconsistency where we didn't use `rest_base` (where we always should be doing).

jdolan: It's an issue because `rest_base` is never exposed to the client. Nowhere in the JSON response for the taxonomy object does it include `rest_base`. It does, however, include the `slug` -- and `slug` is precisely what should be used to map to URIs. That's _what slug is for_. See #1381.

joehoyle: @jdolan #1381 still stands, that's not about what this fix was for. This is about using rest_base to build the url. If #1381 means we set `rest_base` to be `post_tag` then fine. This code is right either way.

jdolan: I realize that. I just see more special casing to accommodate the `rest_base` mistake seeping into the code and it makes me cringe. This code doesn't even use `rest_base` from a constant or some other variable -- it just replicates it. If I'm a jerk for pointing this out, then I apologize.

joehoyle: @jdolan `rest_base` is not about `post_tag` we have to support it for custom taxonomies too. If anything, this commit supports you argument, it is _getting rid_ of special casing the case.

PR #1449, opened 2015-07-31
Update Travis CI

ntwb:
- Adds PHP 5.2, 5.3, 5.4, 5.5, 5.6, 7.0, and HHVM Single and Multisite tests parallel to WordPress Core test matrix
- WordPress `/trunk` is now used, sourced from `develop.git.wordpress.org` and removing previous latest/nightly zip method  
- Dedicated PHPCS and PHLint test job
- NodeJS and Composer packages are cached between builds 
- Git shallow clones for WP-API and WordPress Git checkouts
- Improved compatibility for users who have `WP_DEVELOP_DIR`, `WP_TESTS_DIR`, or `WP_ROOT_DIR` constants already defined in their local environment
- Uses the latest versions of Composer, NPM, and Grunt regardless of Travis CI build configuration.


rachelbaker:
@ntwb We used to only run the multisite tests on PHP 5.2.  I see you are now running them on each version.  With the speed of the tests, this isn't a big concern for me.  @rmccue any issues here?

ntwb:
Indeed, per the PR note:
> Adds PHP 5.2, 5.3, 5.4, 5.5, 5.6, 7.0, and HHVM Single and Multisite tests parallel to WordPress Core test matrix

When this gets merged into WordPress we'll know beforehand that it passes the same test matrix :smirk:

ntwb:
:+1:

Issue #1196, opened 2015-05-06
Return collections as an object and include pagination, links


danielbachhuber:
Rather than returning a collection of objects as an array, we need to return the collection response as an object with the array of data under `data` attribute or similar. This will let us add `_links` for the top-level collection, as well as move our existing pagination headers into the response body.

Related #1136, #865

rmccue:
Bumping this one forward to beta 3, since I want to test it a bit more first.

joehoyle:
We collectively decided to keep the response as an array rather than wrapper object.

danielbachhuber:
Discussion: https://wordpress.slack.com/archives/core-restapi/p1449532826001485

danielbachhuber [4:00 PM] ok
​[4:00] collections as an object is on @rmccue :simple_smile:
​[4:00] I wipe my hands of that
joehoyle [4:00 PM] I thought he decided not to do it at some point
rmccue [4:01 PM] @joehoyle: yeahhhh...
joehoyle [4:01 PM] FWIW the .com api is a wrapper object
​[4:01] but, I still don't like it
danielbachhuber [4:01 PM] I think collections is Ryan's december project
joehoyle [4:01 PM] :stuck_out_tongue:
rmccue [4:01 PM] At this point, I'm going to go with a no on changing it
danielbachhuber [4:01 PM] oh, and just ignore the header problem?
​[4:01] or limit the size?
rmccue [4:02 PM] @danielbachhuber: I'm on holidays from next week, and already spent my one day this week :stuck_out_tongue:(edited)
joehoyle [4:02 PM] wha'ts the header problem?
​[4:02] X-WP-Total pages is small
rmccue [4:02 PM] The header problem was because we were sending one-link-per-item
joehoyle [4:02 PM] right, -1 on that, obvs.
rmccue [4:02 PM] i.e. you hit `/posts` and it includes links to all the items
danielbachhuber [4:02 PM] https://github.com/WP-API/WP-API/issues/475
rmccue [4:02 PM] Just the next/prev links should be fine
github BOT [4:02 PM] [WP-API/WP-API] Issue closed: #1045 500 error on /wp-json/posts by danielbachhuber
rmccue [4:03 PM] We're already sending the autodiscovery Link header on the rest of the site
joehoyle [4:03 PM] Ok, Close?
danielbachhuber [4:03 PM] I'm fine with continuing to use an array

Issue #865, opened 2015-02-09
Should we embed links from collection items into the collection response?


rmccue:
Related to #859.

Let's work out the best way to bring links up from individual collection items, and allow embedding them into the page.

rmccue:
tl;dr: yes.

PR #1459, opened 2015-08-05
Embed links inside items in a collection

rmccue:
See #865

This is a very simple approach. Essentially: if you return a list from your endpoint, apply `embed_links` to the items inside that rather than the top-level.

If we switch collections to objects, we can handle this differently, but that's a separate issue at #1196.


rachelbaker:
@rmccue well the PHPCS task is still working :P https://travis-ci.org/WP-API/WP-API/jobs/74231936

rachelbaker:
Merged #1459

PR #1462, opened 2015-08-06
Use wp_get_attachment_image_src() on the image sizes in attachments

joehoyle:
Currently it builds this with str_replace which can be error prone, and 
doesn't file the expected `image_downsize` filter than a lot of media related
plugins use to alter the image.


rmccue:
I guess we'll find out if it was using `str_replace` for a reason ¯\_(ツ)_/¯

PR #1464, opened 2015-08-07
Make the embed context a default, not forced

rmccue:
We should allow linking with the `view` context explicitly set. This allows you to build custom endpoints that link out to other collections and include the full data when embedding.

A good use for this is having your Blog page link to the full posts collection to allow loading in the title/permalink as well as the posts to display on the page.


rmccue:
@WP-API/amigos #reviewmerge

PR #1466, opened 2015-08-07
post terms route should respect `rest_base` in taxonomy

joehoyle:


joehoyle:
@WP-API/amigos #reviewmerge

PR #1467, opened 2015-08-07
Fix trailing slashes on non-pretty permalinks

rmccue:
See #1426. This one shouldn't accidentally break embedding; tested locally and it's all working.

Props @danielpunkass for the bug report and initial patch, plus the unit tests which I've borrowed here.


rmccue:
@rachelbaker Assigning to you to double-check. :)

rachelbaker:
Merged #1467

PR #1470, opened 2015-08-10
Use correct attribute when updating attachment post

### LINKED ISSUE #1437: Incorrect "post parent" assignment in attachments controller's prepare_item_for_database.
The `WP_REST_Attachments_Controller` function `prepare_item_for_database` incorrectly plucks out the `post_parent` key when it should be `post`. This prevents attachments from having a "parent" set via the API.

The faulty line [can be found here](https://github.com/WP-API/WP-API/blob/4b35ebbf33f7512977e16aed4f25065b23dbd2cf/lib/endpoints/class-wp-rest-attachments-controller.php#L140).

Changing the line to the following will allow an attachment to have a `post_parent`:

```php
$prepared_attachment->post_parent = (int) $request['post'];
```

rmccue:
Props @picklepete, fixes #1437.

### DIFF
--- a/lib/endpoints/class-wp-rest-attachments-controller.php
+++ b/lib/endpoints/class-wp-rest-attachments-controller.php
@@ -136,7 +136,7 @@ class WP_REST_Attachments_Controller extends WP_REST_Posts_Controller {
 		}
 
 		if ( isset( $request['post'] ) ) {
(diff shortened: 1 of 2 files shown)


rmccue:
@WP-API/amigos #reviewmerge

rachelbaker:
Merged #1470

PR #1471, opened 2015-08-10
filter offset enable

tigusigalpa:


rmccue:
You should be able to handle this already through the `page` parameter (that is `?page=2`). Is there a reason you need offset as well?

tigusigalpa:
Hi, Ryan! I think we need the one because if we need to get all posts with some offset or posts_per_page the page param is very poor for it

jcutrell:
I'd like to do a :+1: on this - there are many cases where offset is needed and page/posts_per_page doesn't quite cut it. Is there a particular reason this parameter needs to be hidden behind authentication?

danielbachhuber:
Closing in favor of #924. We should decide whether we want to support `offset` as a global parameter for collections. You can add support in your local instance of WordPress by filtering `query_vars`

PR #1480, opened 2015-08-16
Update the search example to again use `filter`.

kwight:
Also adds a link to instructions for using Subversion's `trunk` branch for development.


rmccue:
Thanks!

PR #1482, opened 2015-08-17
Adding rest api controller for plugins

dankrause:
My first attempt at a rest API controller for plugins. I'm interested in feedback, so I can get this up to your standards for acceptance.

### DIFF
--- a/lib/endpoints/class-wp-rest-plugins-controller.php
+++ b/lib/endpoints/class-wp-rest-plugins-controller.php
@@ -1,0 +1,277 @@
+<?php
+
+require_once( ABSPATH . '/wp-admin/includes/plugin.php' );
+
+class WP_REST_Plugins_Controller extends WP_REST_Controller {
+
+	/**
+	 * Register the routes for the objects of the controller.
+	 */
+	public function register_routes() {
+		register_rest_route( 'wp/v2', '/plugins', array(
+			array(
+				'methods' => WP_REST_Server::READABLE,
+				'callback' => array( $this, 'get_items' ),
+				'permission_callback' => array( $this, 'get_items_permissions_check' ),
+			),
+			'schema' => array( $this, 'get_public_item_schema' ),
+		) );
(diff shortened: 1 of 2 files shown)


rmccue:
Hi @dankrause, and thanks for the PR!

Typically before we jump into creating new endpoints (especially new resources), we design the data we're looking for first. The code should always come afterwards.

During this design process, we look at what objects we're looking at, what properties they have, and what interactions you can do with them. Once that's done, the next question is how to map that to REST semantics. This is where we take everything we want and work out how to do it with just `GET`, `PUT`, `POST`, and `DELETE`. (For example with plugins, there's no `ACTIVATE` verb, so we'd need to work out how to deal with that.)

Code-wise, this seems like a decent start, but without the solid design behind it, it's hard to say. (The design is also the hardest part, and why we don't create new endpoints in a particularly swift manner.)

We'd be happy to accept a patch to add plugin endpoints, as long as there's a) documented design decisions and process behind it; and b) you're committed to maintaining the endpoint in the plugin. (The API is already quite big as-is, and introducing new endpoints increases that burden, so you'd need to be willing to help us out with that, at least at the start.)

Hopefully this doesn't discourage you; we definitely appreciate the effort you've put in so far! Thanks a bunch!

dankrause:
I'm definitely interested in talking about design, but at the moment, I have no idea how this particular group of developers works. Any time I get involved in a project, I prefer to show up with some code - it tends to give a little weight to the discussions that it generates. That said, what should I do to start the discussion on the design of plugin data in this API?

rmccue:
Happy to work through the process with you. Best time to catch us is during our Monday (23:00 UTC) meeting, or during our Wednesday (23:00 UTC) office hours. However, feel free to ping any one of us (myself, @rachelbaker, @joehoyle, or @danielbachhuber) in the Slack room if you see us online and want to chat. :)

danielbachhuber:
Closing this in favor of #1523 and #1669

Issue #1303, opened 2015-05-27
Missing [file] on line 355 in attachments controller


ViktorPontinen:
Hello,

Not sure if this is intended. Was having issues uploading in media endpoint and found that the code on line 355 was looking like this:
```
$file = wp_handle_upload( $files, $overrides );
``` 

Changed it to:
```
$file = wp_handle_upload( $files['file'], $overrides );
```

Which fixed uploading issues for me(using form-data).

danielbachhuber:
Hm, it works for me locally in testing and e24e1f1299f38849f21c8f9099a1aad87437a8ad was TDD to fix the exact opposite of your change.

Can you share your sample request?

ViktorPontinen:
Using version 2.0-beta1, perhaps it has been fixed after that.
The request:
```
POST /wp-json/wp/v2/media HTTP/1.1
Host: example.com
X-Admin-Public: xxxxxxx
X-Admin-Hash: xxxxxxxx
Cache-Control: no-cache
Postman-Token: xxxxxxx
Content-Type: multipart/form-data; boundary=----WebKitFormBoundary7MA4YWxkTrZu0gW

----WebKitFormBoundary7MA4YWxkTrZu0gW
Content-Disposition: form-data; name="file"; filename="avatar.gif"
Content-Type: image/gif


----WebKitFormBoundary7MA4YWxkTrZu0gW
```

Which gave the following error:
```
File is empty. Please upload something more substantial. This error could also be caused by uploads being disabled in your php.ini or by post_max_size being defined as smaller than upload_max_filesize in php.ini.
```
The php.ini was correctly configured since the attachment upload worked without problems on the  1.2.2 version.

Also have another issue on the same note, I have to send both "post" and "post_parent" fields, with the same value for both, for the attachment to get attached to a parent post, none of them works separately.

rmccue:
This looks like a real bug to me, I think we have an issue with our test cases if it's not being picked up.

kosso:
Uploading a media file was not working for me. After a lot of debugging, and following where the file was being tested, I found that as @ViktorPontinen also suggested, wp_handle_upload needs to have $files['file'], and not just $file.

`$file = wp_handle_upload( $files['file'], $overrides );`

picklepete:
Just experienced this too, as @kosso said, the change to...

```php
$file = wp_handle_upload( $files['file'], $overrides );
```

[...]

PR #1492, opened 2015-08-22
Fix file upload for attachments.

thiago-negri:
Attachments controller is passing the entire $_FILES array to wp_handle_update, when it should only pass a single entry.

This commit fixes it by using the 'file' entry in $_FILES array, as stated in the documentation.

See http://wp-api.org/#media_create-an-attachment


rachelbaker:
Fixes #1303

rachelbaker:
@thiago-negri Cheers!  :confetti_ball:

Issue #1260, opened 2015-05-19
Disambiguate between trashed and permanently deleted Posts and Comments


danielbachhuber:
In `develop`, a client making a `DELETE` request to Posts or Comments has no way of distinguishing between whether the request trashed the resource or permanently deleted it. We should provide this detail.

From #1224

Also, 2x `DELETE wp/v2/comments/<id>` will permanently delete a comment, but 2x `DELETE wp/v2/posts/<id>` will return some `rest_already_deleted` error (2d5cf2ca2987353bcaac17b25e4650250119d7f5, #959). We should make sure these are consistent.

rmccue:
#1271 should help out with this for now, but we want to also wrap the data in an object with a `trashed`/`deleted` key. This **intentionally** won't match the data itself, since we're not returning the resource as-it-currently-is, but rather as-it-was-before-the-call-executed.

Let's bump to b3 though, as we might want to be semi-consistent with #1196 (e.g. use `data` in both)?

rachelbaker:
Punted to Beta 4

danielbachhuber:
After a little bit of discussion, we're going to return the original resource instead of wrapping it in a `data` object.

danielbachhuber:
:ship: :ship:

PR #1499, opened 2015-08-24
Return the trashed status of delete posts/comments

pento:
When a post or a comment is deleted, return a flag to say whether it's been trashed or properly deleted.

See #1260.

### DIFF
--- a/lib/endpoints/class-wp-rest-comments-controller.php
+++ b/lib/endpoints/class-wp-rest-comments-controller.php
@@ -271,6 +271,7 @@ class WP_REST_Comments_Controller extends WP_REST_Controller {
 
 		$get_request = new WP_REST_Request( 'GET', rest_url( '/wp/v2/comments/' . $id ) );
 		$get_request->set_param( 'context', 'edit' );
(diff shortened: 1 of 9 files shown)


pento:
Okay, I give up on Scrutinizer. Let's commit it.

pento:
Ready for review @WP-API/amigos.

lib/endpoints/class-wp-rest-comments-controller.php
```diff
@@ -282,8 +282,18 @@ public function delete_item( $request ) {
 			}
 
 			$result = wp_trash_comment( $comment->comment_ID );
+			$status = 'trashed';
 		}
 
+		$response = $this->prepare_item_for_response( $comment, $get_request );
```
rmccue: If we do this _after_ the comment has been deleted, I think it's possible some preparation might fail here?

pento: Good point, moved back to its original position.

rachelbaker:
Merged #1499

PR #1501, opened 2015-08-25
Allow removing links froma response object

joehoyle:
Currently there's no way a filter can easily remote link relations via the reponse object. This allows that!


rachelbaker:
Merged #1501

PR #1508, opened 2015-08-27
Fix MD5 check on file uploads.

thiago-negri:
Upload from file checking for header 'CONTENT_MD5', but all headers are canonicalized by WP_REST_Request::canonicalize_header_name(), so this header is never present.

Upload from data correctly checks for 'content_md5' header.

Also, upload from file tries to handle the header as a string instead of as an array.

Surprisingly, the test case for 'bad MD5 header' is incorrectly checking for a 201 return, instead of a 412.

This commits fixes the MD5 check on upload from file by copying the way it is handled by upload from data.


rachelbaker:
@thiago-negri, thank you!

PR #1511, opened 2015-08-28
Add help properties to endpoint arguments.

thiago-negri:
Context 'help' already exists to show the schema of each endpoint.

This commits adds the possibility of adding helpful descriptions for humans to the arguments of an endpoint.

This is useful for those learning the API.

Example is given for some arguments of `class-wp-rest-posts-controller.php`.


danielbachhuber:
`description` has been added to the infrastructure committed to core. We'll make sure all of the core arguments have sufficient and correct documentation.

PR #1512, opened 2015-08-29
Add missing 'force' arg to posts-terms DELETE.

thiago-negri:
Function `delete_item()` requires de "force" arg, but the API description does not point it out.


rachelbaker:
Moved to #1705

PR #1516, opened 2015-08-29
Fix typo in message

thiago-negri:


rachelbaker:
Merged #1516 

Thank you again @thiago-negri

PR #1527, opened 2015-09-04
Added content_raw to the returned post fields

kubante:


rmccue:
This is intentionally not available to unauthenticated users. Private data can be exposed in pre-filtered content (attributes for shortcodes, for example).

If you're looking for `content_raw` from v1, you can now use `content.raw` in v2, with `context=edit` and authenticated.

kubante:
I c. Thanks for your explanation rmccue.. Greetings

Issue #1518, opened 2015-08-30
Builds are failing thanks to WPCS


rmccue:
Looks like from WordPress-Coding-Standards/WordPress-Coding-Standards#417 the builds have started failing. You can see this in unrelated pull requests like #1511.

We should do two things:

1. Fix the violations, obviously
2. Stick to a fixed version of WPCS, so we can control updates instead.

Having builds fail for unrelated reasons is pretty shit. We just need a fixed version instead of `dev-develop` in our composer.json.

thiago-negri:
I tried to fix the warnings in #1508.
Should I remove that commit (3a7873ed) from the PR?

Issue #833, opened 2015-02-03
Unable to access post meta in a CORS request with basic auth


federicovezzoli:
I'm developing a phonegap (angular) application and I'm using wordpress as a backend.

I can do basic query, without authentication and everything works fine.
Then, I'm trying to retrieve some post meta through basic authentication and the CORS setttings are locking me out.

My GET call turns into an OPTION call and it fails miserably.
I've tried to specify the origin (which is nonsense for an App but just to try) and I still get a 400.
I've tried everything I could find but nothing seems to work, the basic auth just don't work from the browser.

If I try the same GET authenticated call in POSTMAN it works super fine, no problem at all.
Where is the difference?

Any hint?

danielbachhuber:
Can you share some example requests and responses so I have a better sense of what you're talking about?

federicovezzoli:
of course.

Request in the ctrl:
```
SediMeta.get({id: value.ID}, function(l,getResponseHeaders){
     console.log(l);
});
```

Factory:
```
//Sedi Meta
.factory('SediMeta',['$resource', '$base64',
	function($resource, $base64){
		return $resource('http://fonda56.com/content/wp-json/posts/:id/meta/', {}, {
			get : {
				method: 'GET',
				//params: {},
				isArray: true,
				/* Note the headers property */
				headers: {
					Authorization : 'Basic YWRtaW46cHJpbWF0ZTIwNQ=='
				},
				cache: false
			}
		});
	}
]);
```
Response
![schermata 2015-02-05 alle 08 37 46](https://cloud.githubusercontent.com/assets/404154/6056525/51d73aba-ad12-11e4-965c-e773e355bbaf.png)

the same exact request works fine in POSTMAN.
In jQuery it does the exact same behaviour, I recon is some CORS issue but cannot work it out.
Let me know if you need anything.

danielbachhuber:
> http://fonda56.com/content/wp-json/posts/:id/meta/

`:id` needs to be replaced with an actual ID.

federicovezzoli:
As you can see in the screenshot, the request url is correct:

```
http://fonda56.com/content/wp-json/posts/23/meta
```

if you don't know angular I can post the jquery request, it result in the same issue.

thanks

rmccue:
> I can do basic query, without authentication and everything works fine.
> Then, I'm trying to retrieve some post meta through basic authentication and the CORS setttings are locking me out.

[...]

Issue #144, opened 2014-04-18
Allow Cross-domain requests


austincondiff:
I am unable to make a request from a separate domain. Is this supported by this plugin? I am developing an AngularJS app locally and I am having trouble requesting data from my live Wordpress site.

krewmarco:
BTW, if using application passwords and  angular, you set $httpProvier.defaults.withCredentials = true, you might find your user is being set to 0 if you did not send a nonce.   That's because youre telling the API youre using cookie-auth which requires a nonce (Check rest-api.php rest_cookie_check_errors() ).  Remove withCredentials and it does what you expect.  Save someone some hours...

krewmarco:
Since 4.4.0, I believe the right way to enable CORS is
add_filter('http_origin', function() { return "http://your domain";});
see wp-includes/rest_api.php rest_send_cors_headers() .  Seems to work fine.

createbuildship:
where do you add add_filters?

kadamwhite:
@frontr-uk in your theme or plugin code.

PR #1529, opened 2015-09-05
Allow Cross Domain Basic auth

shprink:
and add paginations to exposed headers list so it can be retrieved by clients.

Related to this old PR I made: https://github.com/WP-API/WP-API/pull/802

### DIFF
--- a/lib/infrastructure/class-wp-rest-server.php
+++ b/lib/infrastructure/class-wp-rest-server.php
@@ -250,6 +250,7 @@ class WP_REST_Server {
 		// Mitigate possible JSONP Flash attacks
 		// http://miki.it/blog/2014/7/8/abusing-jsonp-with-rosetta-flash/
 		$this->send_header( 'X-Content-Type-Options', 'nosniff' );
+		$this->send_header( 'Access-Control-Allow-Headers', 'X-WP-Total, X-WP-TotalPages, Authorization' );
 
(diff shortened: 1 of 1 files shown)


itsananderson:
:+1: Without this, it's impossible to do cross-origin AJAX requests for authenticated resources.

One thing I'm curious about is, should we add these headers in https://github.com/WP-API/WP-API/blob/818eec1291d059406289eab6e9bc4888c4473f68/plugin.php#L597

That's where the other `Access-Control-` headers are. Alternatively, we could move those other headers into this function.

Also, making those headers filterable would be great for custom plugin developers, who might want to expose or accept additional headers. I have an example of what that might look like in my "wp-api-cors" plugin. https://github.com/itsananderson/wp-api-cors/blob/09d0e5e13343b492fb0ab060b59bde836505220d/wp-api-cors.php#L51

scottopolis:
+1

rachelbaker:
Related #833 and #144

rmccue:
:+1: Thanks for this! The build failed because of #1518, so merging this despite that.

rickharrison:
How do you customize the `Access-Control-Allow-Headers` header? I am trying to work with a wordpress API and no methods will go through because `Content-Type` is not allowed.

kadamwhite:
@rickharrison that's a good question, moved to its own issue: #2844

PR #1537, opened 2015-09-07
Specify a version for WPCS in composer.json

### LINKED ISSUE #1518: Builds are failing thanks to WPCS

rachelbaker:
Fixes #1518 and our builds

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -18,7 +18,7 @@
         "composer/installers": "~1.0"
     },
     "require-dev": {
-        "wp-coding-standards/wpcs": "0.7.1"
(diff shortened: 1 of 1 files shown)


rachelbaker:
Merged #1537 to fix our Travis builds

PR #1538, opened 2015-09-07
Fix force param logic when deleting Post Terms

### LINKED ISSUE #1522: Delete Post Terms, strange error on force parameter
Hello,

when deleting a category from a post I get the following error:

```json
[{
    "code": "rest_trash_not_supported",
    "message": "Terms do not support trashing.",
    "data": {
      "status": 501
    }}]
```
as soon as I set the "force" parameter to false the error goes away and the category is deleted from the post as expected...

My expectation was that the "force" parameter was optional and anyway defaulted to "false". Let me know if I missed something.

rachelbaker:
Default the `force` param to false, and only display error about not supporting true if it is explicitly set. 

Fixes #1522 and #1512


rachelbaker:
@WP-API/amigos #reviewmerge

rachelbaker:
Ignore, this goes against #1304

PR #1541, opened 2015-09-08
Add `post_count` field in users response

AhmadElMelegy:
Add number of posts the user [author] has published in the json response of `prepare_item_for_response` in class `wp-rest-users-controller`.


rmccue:
Thanks for the pull request! `count_user_posts` is actually a reasonably expensive and uncached call, so we don't want to add this to the user response all the time. I think it could potentially be useful with certain contexts, so it's likely worth opening an issue to discuss this. (Generally, we prefer discussing issues before getting to the code/PR stage, just so we can hammer out all the details. :) )

Issue #1549, opened 2015-09-14
Audit and improve hook docs


adamsilverstein:
Looking forward to the core merge, we should make sure our hook docs follow core standards and will be easily parseable. Scanning thru the existing code, some hooks are documented while others are not. 

Here is an inventory of project files where I found hooks and will be going thru to update/add docs:

- [x] plugin.php
- [x] extras.php
- [x] lib/endpoints/class-wp-rest-attachments-controller.php
- [x] lib/endpoints/class-wp-rest-comments-controller.php
- [x] lib/endpoints/class-wp-rest-meta-controller.php
- [x] lib/endpoints/class-wp-rest-posts-controller.php
- [x] lib/endpoints/class-wp-rest-taxonomies-controller.php
- [x] lib/endpoints/class-wp-rest-terms-controller.php
- [x] lib/endpoints/class-wp-rest-users-controller.php
- [x] lib/infrastructure/class-wp-rest-request.php
- [x] lib/infrastructure/class-wp-rest-server.php

adamsilverstein:
Making good progress on the hook docs. During the hook audit, I noticed the absence of several hooks in some places. For example, actions are not fired when terms or taxonomies are added or updated (posts have such actions). This appears to be an oversight, so I will go back and add the missing hooks to try to make pattern as consistent as possible and have hooks everywhere you would expect them as a developer.

I also noticed that no actions are fired on delete anywhere, which I think would be worth remedying. Feedback?

adamsilverstein:
Another note from the audit: We should add unit tests to verify the hooks are fired correctly.

rachelbaker:
@adamsilverstein looks like all 11 items have been checked-off, and other issues have been filed for the additional issues you discovered.  Can this issue be closed?

adamsilverstein:
yep, closing.

PR #1552, opened 2015-09-15
Hook docs- extras.php

adamsilverstein:

* Add a missing hook doc.

* xtras.php:185:	return apply_filters( 'rest_avatar_sizes', array( 24,
48, 96 ) );


rachelbaker:
Merged #1552 

@adamsilverstein :heart_eyes:

PR #1555, opened 2015-09-16
Hook Docs: Add missing hook docs in lib/infrastructure/class-wp-rest-request.php

adamsilverstein:


rachelbaker:
Merged #1555

PR #1556, opened 2015-09-16
Hook Docs for lib/endpoints/class-wp-rest-users-controller.php

adamsilverstein:
See https://github.com/WP-API/WP-API/issues/1549.

* Add some missing hook docs
* Spacing, punctuation cleanup as per standard
* Change second do_action from `rest_insert_user` to `rest_update_user` as
its in the update function
* added some todo notes, I don’t understand what the final false
variable is for, maybe its needed?

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -2,3 +2,5 @@
 composer.lock
 node_modules
 vendor
+
+.idea/
--- a/.gitignore
+++ b/.gitignore
@@ -2,5 +2,3 @@
 composer.lock
 node_modules
 vendor
-
-.idea/
--- a/lib/endpoints/class-wp-rest-users-controller.php
+++ b/lib/endpoints/class-wp-rest-users-controller.php
@@ -238,12 +238,11 @@ class WP_REST_Users_Controller extends WP_REST_Controller {
 		/**
 		 * Fires after a user is created via the REST API.
 		 *
-		 * @param object          $user    Data used to create the user (not a WP_User object).
-		 * @param WP_REST_Request $request Request object.
-		 * @param bool            $bool    A boolean that is false.
-		 * @todo remove final bool = false variable, unused?
+		 * @param object          $user      Data used to create the user (not a WP_User object).
+		 * @param WP_REST_Request $request   Request object.
+		 * @param bool            $creating  True when creating user, false when updating user.
 		 */
-		do_action( 'rest_insert_user', $user, $request, false );
+		do_action( 'rest_insert_user', $user, $request, true );
 
 		$response = $this->get_item( array(
 			'id'      => $user_id,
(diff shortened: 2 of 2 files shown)


lib/endpoints/class-wp-rest-users-controller.php
```diff
@@ -235,11 +236,12 @@ public function create_item( $request ) {
...
+		 * Fires after a user is created via the REST API.
 		 *
-		 * @param object $user Data used to create user (not a WP_User object)
+		 * @param object          $user    Data used to create the user (not a WP_User object).
 		 * @param WP_REST_Request $request Request object.
-		 * @param bool $bool A boolean that is false.
+		 * @param bool            $bool    A boolean that is false.
+		 * @todo remove final bool = false variable, unused?
 		 */
 		do_action( 'rest_insert_user', $user, $request, false );
```
rachelbaker: @adamsilverstein The third param here should be `true` on "Create" (it should be false on update).

lib/endpoints/class-wp-rest-users-controller.php
```diff
@@ -299,7 +301,15 @@ public function update_item( $request ) {
...
-		do_action( 'rest_insert_user', $user, $request, false );
+		/**
+		 * Fires after a user is updated via the REST API.
+		 *
+		 * @param object          $user    Data used to update the user (not a WP_User object).
+		 * @param WP_REST_Request $request Request object.
+		 * @param bool            $bool    A boolean that is false.
+		 * @todo remove final bool = false variable, unused?
+		 */
+		do_action( 'rest_update_user', $user, $request, false );
```
rachelbaker: @adamsilverstein to keep the same naming convention in the other endpoint controllers, it should be `rest_insert_user` but with the third param of `false`.

lib/endpoints/class-wp-rest-users-controller.php
adamsilverstein: Thanks for clarifying, I will add that info to the docblock

lib/endpoints/class-wp-rest-users-controller.php
adamsilverstein: I'm learning so much already :) Thanks, will fix the docblock.

adamsilverstein:
Note: I believe this also corrects an issue with the first call of the `rest_insert_user` filter inside `create_item`, I changed the last ("created") parameter to true as per @rachelbaker's comment on my previous commit (https://github.com/WP-API/WP-API/pull/1556#discussion_r39616194). 

Please confirm this is correct before merging :)

adamsilverstein:
Working on lib/endpoints/class-wp-rest-terms-controller.php now and wondering why I don't see similar do_action( 'rest_insert_term' ...) hooks firing. Is this intentional, or an oversight? If the latter I can go ahead and add them as I go :)

rachelbaker:
> Working on lib/endpoints/class-wp-rest-terms-controller.php now and wondering why I don't see similar do_action( 'rest_insert_term' ...) hooks firing. Is this intentional, or an oversight? If the latter I can go ahead and add them as I go :)

@adamsilverstein Just an oversight.

.gitignore
```diff
@@ -2,3 +2,5 @@
 composer.lock
 node_modules
 vendor
+
+.idea/
```
rachelbaker: @adamsilverstein can you undo this change to the .gitignore?

adamsilverstein: Ugh, should be in my local exclude sorry. kind of switching back to sublime anyway :)

adamsilverstein: Removed.

adamsilverstein:
@rachelbaker This should be good to go unless I missed some feedback, I'm addressing adding missing hooks in https://github.com/WP-API/WP-API/issues/1571

rmccue:
:sparkles:

Issue #1548, opened 2015-09-12
Approving a comment deletes content and author


scottopolis:
*Bug:* Changing comment status from hold to approved using a PUT request deletes the comment author and comment content. I am using the basic auth plugin and an $http request in AngularJS.

*How to replicate:* Add an unapproved comment to your site with the WP API v2 and basic auth plugins installed. Using Postman, send a PUT request with a basic auth header to a comment endpoint, such as mysite.com/wp-json/wp/v2/comments/54

In the form data, send along status=approved.  The comment will be approved, but the comment author and comment content are deleted.

scottopolis:
Sending a POST instead of PUT seems to retain the comment author, but the content is still deleted.

rachelbaker:
@scottopolis Thank you for reporting this.  Looks like the default values from the schema are causing this issue.  I started going down this rabbit hole and ran out of energy for the evening.  I will pick it back up tomorrow.

scottopolis:
I think I figured out the issue here, will send a PR shortly.

joehoyle:
@scottopolis I have a fix in #1557 already, just need to clean a couple things up before merging

scottopolis:
@joehoyle ok cool. Just FYI I'll report what I found here in case it's helpful:

Line 218 of class-wp-rest-comments-controller.php $prepared_args is not empty, so the next line `if ( empty( $prepared_args ) && isset( $request['status'] ) ) ` is false when it should be true.

Going down to `$this->prepare_item_for_database( $request )` I found that `$request['parent']` is 0 instead of empty, and `$request['author_name']` is "" instead of null.  That is why $prepared_args is not empty.

Hope that helps!

rachelbaker:
Fixed in #1557

PR #1557, opened 2015-09-16
Modify get_endpoint_args_for_item_schema to handle required/default

joehoyle:
Now get_endpoint_args_for_item_schema() will take a $method param which is used to determin if the endpoint args should include the `required` and `default` parameters.

We could have potentially handled this at the WP_REST_Server level, however the "incorrect" part here is generating the args from the schema. `get_endpoint_args_for_item_schema` is really shorthand for writing out the args for the endpoint - in the case we were writing those by hand, we wouldn't be passing `default` and `required` in the case of `EDITABLE` endpoint args.

### DIFF
--- a/lib/endpoints/class-wp-rest-controller.php
+++ b/lib/endpoints/class-wp-rest-controller.php
@@ -443,7 +443,7 @@ abstract class WP_REST_Controller {
 			return new WP_Error( 'rest_invalid_param', sprintf( __( '%s is not of type %s' ), $parameter, 'integer' ) );
 		}
 
-		if ( 'string' === $property['type']&& ! is_string( $value ) ) {
+		if ( 'string' === $property['type'] && ! is_string( $value ) ) {
 			return new WP_Error( 'rest_invalid_param', sprintf( __( '%s is not of type %s' ), $parameter, 'string' ) );
 		}
 
--- a/lib/endpoints/class-wp-rest-controller.php
+++ b/lib/endpoints/class-wp-rest-controller.php
@@ -365,7 +365,7 @@ abstract class WP_REST_Controller {
 	 *                           Where as create requests will.
 	 * @return array
 	 */
-	public function get_endpoint_args_for_item_schema( $add_required_flag = true ) {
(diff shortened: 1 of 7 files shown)


joehoyle:
See #1548

joehoyle:
@WP-API/amigos I need to fix up the tests, but does this make sense? I tried to explain the direction in the description above.

lib/endpoints/class-wp-rest-controller.php
```diff
@@ -365,7 +365,7 @@ protected function get_object_type() {
 	 *                           Where as create requests will.
 	 * @return array
 	 */
-	public function get_endpoint_args_for_item_schema( $add_required_flag = true ) {
+	public function get_endpoint_args_for_item_schema( $method = WP_REST_Server::CREATABLE ) {
```
rachelbaker: @joehoyle Can you update the Docblock as well?

rachelbaker:
@joehoyle I think this is a smart approach.  :+1:

lib/endpoints/class-wp-rest-controller.php
rmccue: Should we maybe rename this to just `get_endpoint_args`, and make `$method` required?

lib/endpoints/class-wp-rest-controller.php
```diff
@@ -383,16 +383,22 @@ public function get_endpoint_args_for_item_schema( $add_required_flag = true ) {
 				'sanitize_callback' => array( $this, 'sanitize_schema_property' ),
 			);
 
-			if ( isset( $params['default'] ) ) {
+			if ( WP_REST_Server::CREATABLE && isset( $params['default'] ) ) {
```
rmccue: This will always be true, you want `$method ==` ;)

rachelbaker:
@joehoyle @rmccue  I addressed Ryan's feedback (minus renaming the function) and am going to merge this to get the fix in for this pretty big :bug:

rachelbaker:
Merged #1557

PR #1560, opened 2015-09-17
Add hook docs for lib/endpoints/class-wp-rest-taxonomies-controller.php

adamsilverstein:
* Docs for `rest_prepare_taxonomy`


rmccue:
Thanks :)

PR #1563, opened 2015-09-18
Change featured image links

### LINKED ISSUE #1476: Featured image is not in _links on post response
We should link to the featured image, so it can be embedded, see http://demo.wp-api.org/wp-json/wp/v2/posts/111?_embed it's not included in the links.

@rmccue I know we ran into an issue with it embedding all attachments attached to the post, maybe that's the issue here.

rmccue:
Rather than linking to `wp/v2/media?post_parent=<id>`, we should be linking directly to the featured image, with a secondary link out to the former.

Fixes #1476.

### DIFF
--- a/lib/endpoints/class-wp-rest-posts-controller.php
+++ b/lib/endpoints/class-wp-rest-posts-controller.php
@@ -1159,7 +1159,6 @@ class WP_REST_Posts_Controller extends WP_REST_Controller {
 			$attachments_url = add_query_arg( 'post_parent', $post->ID, $attachments_url );
 			$links['http://v2.wp-api.org/attachment'][] = array(
 				'href'       => $attachments_url,
-				'embeddable' => true,
 			);
 		}
 
--- a/lib/endpoints/class-wp-rest-posts-controller.php
+++ b/lib/endpoints/class-wp-rest-posts-controller.php
@@ -1144,10 +1144,20 @@ class WP_REST_Posts_Controller extends WP_REST_Controller {
 			);
 		}
 
+		// If we have a featured image, add that
+		$links['http://v2.wp-api.org/attachment'] = array();
+		if ( $featured_image = get_post_thumbnail_id( $post->ID ) ) {
+			$image_url = rest_url( 'wp/v2/media/' . $featured_image );
+			$links['http://v2.wp-api.org/attachment'][] = array(
+				'href'       => $image_url,
+				'embeddable' => true,
+				'featured'   => true,
+			);
+		}
 		if ( ! in_array( $post->post_type, array( 'attachment', 'nav_menu_item', 'revision' ) ) ) {
 			$attachments_url = rest_url( 'wp/v2/media' );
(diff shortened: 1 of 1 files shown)


rmccue:
I was thinking we should maybe use a different relation for attached vs featured; maybe `http://v2.wp-api.org/featured`? In lieu of that, I've changed it to have `featured: true`

@WP-API/amigos Thoughts?

scottopolis:
Could we add the post_thumbnail src somewhere? It should be part of the embedded object so we can display a list of posts with thumbnails without going to the /media endpoint separately. For example:

    $links['http://v2.wp-api.org/attachment'] = array();
		if ( $featured_image = get_post_thumbnail_id( $post->ID ) ) {
			$image_url = rest_url( 'wp/v2/media/' . $featured_image );
			$links['http://v2.wp-api.org/attachment'][] = array(
				'href'       => $image_url,
				'embeddable' => true,
				'featured'   => true,
				'post_thumbnail'  => wp_get_attachment_image_src( $featured_image )[0]
			);
		}

rachelbaker:
@rmccue Any thoughts on @scottopolis's request?

> Could we add the post_thumbnail src somewhere?

rmccue:
@scottopolis Not sure what you're wanting here? The `source_url` is exposed; guessing you mean `media_details.sizes` should be exposed too?

rmccue:
@rachelbaker If we do make a change around the image URLs, it needs to be in the embedded data, not on the link itself, so shouldn't affect this PR. :)

scottopolis:
@rmccue @rachelbaker Yes, we need post thumbnail sizes, not just the source url in the _embed object.  This would be just the post thumb, not all attachments.

Example use case would be listing posts in a mobile app, so I can display the 150x150px post thumbnail for each post.

scottopolis:
Here is the pull request related to my comments https://github.com/WP-API/WP-API/pull/1582

rachelbaker:
Merged #1563

joehoyle:
I think we should have gone with a different relation here, I think having it part of the attachments relation is a bit cumbersome to work with for clients. It looks like that discussed wasn't picked up however.

danielbachhuber:
> I think having it part of the attachments relation is a bit cumbersome to work with for clients

Why so?

joehoyle:
Looking at the response in http://demo.wp-api.org/wp-json/wp/v2/posts/162?_embed, there's no way to hardcode the path to the featured image link for embedded item, you'd have to do a filter on the `http://v2.wp-api.org/attachment` relation to only get ones with `embedded => true`.

If you want to get the data from embedded, there's no way to programmatically know which embedded attachment is the featured image, short of cross referencing with the `_links` array, or blindly _embedded["http://v2.wp-api.org/attachment"][0]`.

I don't really understand the rational for the way we have it now, they are different relationships; it seems we are grouping the attachments with the featured image because they are the same resource type, which actually has little bearing on what the relationship is.

PR #1566, opened 2015-09-18
Hook docs: lib/endpoints/class-wp-rest-attachments-controller.php

adamsilverstein:
* Add hook docs for `rest_prepare_attachment`


rachelbaker:
Merged #1566

PR #1574, opened 2015-09-20
Amends PHPDoc and removes unneeded stuff in wp-api.js

eliorivero:
- Amends PHPDoc for $item parameter in prepare_item_for_response method of WP_REST_Revisions_Controller class.

- Removes unnecessary semicolon in wp-api.js

- Removes unused variable origParse in wp-api.js. Seems this was used to keep a copy of original parse method, but in wp-api.js parse is not replaced so this not necessary.


rachelbaker:
@eliorivero Hello, and thank you for submitting this Pull Request.  It looks like with one pull request you are making three very different changes, in two different files.  

Would you be able to split this on PR into two separate Pull Requests?  One for the Revisions Docblock, and the other for the fixes to the JS client?

PR #1576, opened 2015-09-20
wp-api.js fixes

eliorivero:
Removes unused variable origParse in wp-api.js. Seems this was used to keep a copy of original parse method, but in wp-api.js, parse is not replaced so this not necessary.

Removes unnecessary semicolon.


danielbachhuber:
Handled in #1709

PR #1581, opened 2015-09-22
fixes issue with terms controller request params 'parent' and 'per_page' being 0

Toddses:
For the `/terms/` endpoints, if the `parent` param is 0 (eg. retrieve all top-level terms), it fails to take into account that no taxonomy with an id of 0 exists.

For the `per_page`, it tries to figure out the `$max_pages` by dividing the total items by the `per_page` parameter. This of course fails because of dividing by 0.

### DIFF
--- a/lib/endpoints/class-wp-rest-terms-controller.php
+++ b/lib/endpoints/class-wp-rest-terms-controller.php
@@ -76,10 +76,15 @@ class WP_REST_Terms_Controller extends WP_REST_Controller {
 
 		$taxonomy_obj = get_taxonomy( $this->taxonomy );
 		if ( $taxonomy_obj->hierarchical && isset( $request['parent'] ) ) {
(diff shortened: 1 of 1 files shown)


Toddses:
this may be an issue with the other controllers as well, not just the terms controller.

rachelbaker:
@Toddses Hello, and thanks for submitting this PR.  `per_page` should never be `0`.  I think the best approach to fixing that bug, is setting a default value (see #1603) so we don't divide by zero if the `per_page` argument isn't set.

rachelbaker:
Merged in #1604

Toddses:
What about when we want to query for all terms? Passing 0 in the `numbers` arg to the `get_terms()` function is one way. The default for `numbers` is to return all terms. However, if the default value is 10 in the terms controller, and we can't pass 0, there's no way to query for all terms.

PR #1582, opened 2015-09-22
Add featured image sizes

scottopolis:
This adds the sizes, but I wasn't sure how to restrict this to only show when using ?_embed.  I'd be happy to edit as necessary.


scottopolis:
See comments on https://github.com/WP-API/WP-API/pull/1563

scottopolis:
@rmccue @rachelbaker Can I get your opinion this? Right now we have to make a separate HTTP request to get the post thumbnails when listing posts.  Trying to include src urls in the _embed object.

Thanks!

danielbachhuber:
We aren't going to change the `featured_image` attribute on the Post resource to return an object instead of an ID. However, we can make sure the linked object includes the `sizes` array: #1667

To get embedded representations of linked objects, you can use `?_embed` on any GET request.

PR #1584, opened 2015-09-23
Use permission check method instead of get_item

### LINKED ISSUE #1583: Avoid expensive Post->get_item call from Post_Terms
Right now, `Post_Terms::validate_request` calls `$this->posts_controller->get_item`, although it actually only needs to check permissions. This is relatively expensive, as it means that when you're grabbing a post with embeds, it needs to prepare the post ~4 times for one request.

rmccue:
Fixes #1583


rachelbaker:
@rmccue Looks like you have a failing unit test.

rmccue:
Indeed, I'll fix this up shortly. :)

rmccue:
Blocked on #1644.

danielbachhuber:
@rmccue Why aren't you doing a `HEAD` check (which I think is more semantic)? Using the permissions check seems like a bit of a hack.

rmccue:
> Why aren't you doing a `HEAD` check (which I think is more semantic)?

`HEAD` will still run the callback, which is the expensive part. We _might_ be able to get away with `OPTIONS` though...

rmccue:
Nope, `OPTIONS` doesn't do the permission checks, so no luck there I think. I think it's fine using the method directly, since we're already doing so for `get_post_type_base`

joehoyle:
I'm pretty sure `OPTIONS` does run permissions checks for the Accept header, though I may be wrong

danielbachhuber:
> HEAD will still run the callback, which is the expensive part. 

Maybe we should file an enhancement for more performant HEAD requests?

rmccue:
> I'm pretty sure OPTIONS does run permissions checks for the Accept header, though I may be wrong

Nope, it just uses the registered HTTP methods; it doesn't run the permission checks. Maybe it should though.

> Maybe we should file an enhancement for more performant HEAD requests?

I think it might be hard to separate the content generation out, as we'd still need to check all the errors that are deep down. I think changing OPTIONS would be the better option (heh) here if anything.

joehoyle:
> Nope, it just uses the registered HTTP methods; it doesn't run the permission checks. Maybe it should though.

Ahh, so any registered method that's passed through `dispatch` will get the `Accept` header from `rest_send_allow_header` which will run permissions checks. Right now we don't run that on `OPTIONS` which I think we should be doing.

rachelbaker:
Fixed in #1697

Issue #1502, opened 2015-08-25
get_rest_url() can return the wrong URL in Multisite


pento:
`get_rest_url()` uses `get_option( 'permalink_structure' )`, which will return the option for the current site. It then assumes that the blog given by `$blog_id` will then have the same permalink_structure.

We should be using `get_blog_option()` wrapped in an `is_multisite()` check, instead.

PR #1586, opened 2015-09-23
Added conditional check for multisite for `get_rest_url`

JohnDittmar:
Altered the conditional for 'permalink_structure' to cover subsites
of a multisite if applicable.


rachelbaker:
Related #1502

rachelbaker:
Fixes #1502

PR #1590, opened 2015-09-23
Core Docs standards

DrewAPicture:
https://make.wordpress.org/core/handbook/best-practices/inline-documentation-standards/php/


DrewAPicture:
@rmccue: I ... think I'm done here. Let me know if I missed any files. I did plugin.php and extras.php, and then the contents of lib/infrastructure.

rachelbaker:
@DrewAPicture Thank you!  :dancers:

PR #1591, opened 2015-09-23
Add the embed context to Users collection query params

rachelbaker:
The default context for a single user request is `embed`.  Whereas the default context for the users collection is `view`.  

This just adds the `embed` context as a specified option for users collection requests.


rachelbaker:
@WP-API/amigos #reviewmerge

rachelbaker:
Labeled as a bug, since `embed` is a valid and accepted context for the Users collection.

PR #1593, opened 2015-09-23
Change rest_send_allow_header() so it has a single return sentence.

eliorivero:


rachelbaker:
@eliorivero,

Can you give me some more details about your patch?

eliorivero:
@rachelbaker nothing meaningful regarding functioning, just readability. It's usually a bit more readable to have a single return at the end of a function when possible. This still would need a refresh. Will close it for now, let me know if you consider it meaningful and I'll add a new one.

rmccue:
FWIW, :-1: on this. Returning early is the preferred style, as it simplifies the code and means you don't end up with nested code 12 levels deep because of the conditionals.

eliorivero:
@rmccue roger that captain. I do prefer early return on long functions to avoid exactly what you mentioned, but this one is pretty short. Understoon, won't add a new one. Thanks Ryan for taking the time to check this. Cheers.

PR #1596, opened 2015-09-25
Remove unused `$supported` methods variable

jeremyfelt:
In 6388ce46530bfb2, logic was changed to rely on the handler data
directly. The capture of supported methods into a local variable
is no longer necessary.


rachelbaker:
Merged #1596.  Gracias, @jeremyfelt

PR #1597, opened 2015-09-25
Fix misspelling of since tag in `WP_Rest_Server` docs

jeremyfelt:
s/sincd/since/


rachelbaker:
Merged #1597.  Thank you @jeremyfelt

PR #1601, opened 2015-09-26
Correct methods in WP_REST_SERVER::EDITABLE description

rachelbaker:
`EDITABLE` is an alias for POST, PUT, and PATCH, not GET


rachelbaker:
@WP-API/amigos #reviewmerge

PR #1603, opened 2015-09-27
Add Terms Controller Collection Args Details

rachelbaker:
* Add missing collection args details to the Terms Controller
     * fill-in details for the `per_page`, `page`, `hide_empty`, and `search` args that are already handled in the get_items method
     * add a `sanitize_callback` to the `order` arg
     * add a `sanitize_callback` and add missing valid values to the enum array for the `orderby` arg
 * Don't hardcode the `hide_empty` collection arg to being false in `get_items` method

### DIFF
--- a/lib/endpoints/class-wp-rest-terms-controller.php
+++ b/lib/endpoints/class-wp-rest-terms-controller.php
@@ -615,7 +615,6 @@ class WP_REST_Terms_Controller extends WP_REST_Controller {
 			$query_params['parent'] = array(
 				'description'        => 'Limit result set to terms assigned to a specific parent term.',
 				'type'               => 'integer',
-				'sanitize_callback'  => 'absint',
 			);
 		}
 		return $query_params;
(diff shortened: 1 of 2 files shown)


rachelbaker:
@WP-API/amigos #reviewmerge

rachelbaker:
Related #1581

danielbachhuber:
If we're exposing `hide_empty` as an argument now, can we add a test for it?

Also, I think I didn't expose it originally because I wanted to have a conversation about its naming. Should we rename it? Querying posts has `with_comments`, so it could become `with_posts`

rachelbaker:
@danielbachhuber I added a test for `hide_empty`. I am just trying to fill out our collection args for the documentation, and don't feel strongly about renaming it.

danielbachhuber:
> I added a test for hide_empty. I am just trying to fill out our collection args for the documentation

Thanks, makes sense.

> don't feel strongly about renaming it.

@rmccue @joehoyle opinions?

rmccue:
Hmm, `hide_empty` is the more generic term there, so it'd make sense to make it consistent on that instead, but I'm wary of having to rename stuff particularly when it's in WP Query (which everyone knows).

Merging in the meantime, we can continue discussing the change though.

PR #1607, opened 2015-09-28
Switch to register_post_type_args filter.

### LINKED ISSUE #1588: Switch to register_post_type_args filter
@DrewAPicture pointed out that there's a new filter we can use, rather than using the global

jjeaton:
WordPress 4.4 includes a 'register_post_type_args' filter
which can be used instead of directly modifying the global.
Previously, the function was hooked to `init` with a
priority of 11, but it is now set to 10 on the new filter.

Added unit tests to confirm the arguments are added
to each type.

Fixes #1588.


rachelbaker:
@jjeaton Thank you for the contribution!

PR #1608, opened 2015-09-28
Patch PHPDoc of rest_send_allow_header() adding @return tag.

eliorivero:


danielbachhuber:
This function is now in WordPress core, so I've submitted a patch clarifying the return value https://core.trac.wordpress.org/ticket/34374

PR #1610, opened 2015-09-29
Adding Unit Tests for the `rest_url` and `rest_url_prefix` filters

ryanwelcher:
I'd like to contribute some Unit Tests to the project based on https://github.com/WP-API/WP-API/issues/1549#issuecomment-141571410. I am new to the project so any direction is greatly appreciated as to expected location of tests or formatting etc.


rachelbaker:
@ryanwelcher Now that the infrastructure is part of WP Core, these unit tests would need to be moved to a core patch.  Sorry for letting this PR get so stale.

Issue #1621, opened 2015-10-01
WP_REST_Request:get_params()/get_param() returns route...


Shelob9:
Not sure if this was intended, but it's annoying me. In my callback, where `$request` is `WP_REST_Request` object and I do this:

```
$params = $request->get_params();
$zero = $request->get_param( 0 );
```

I get a 0 index in `$params` and `$zero` has the route URL. I don't see why it <em>should</em> be a param, it's not registered on the endpoint, and its redundant to what I can get with the `get_url_params()` method.

joehoyle:
Hmm I'm not sure if this was ever intentional or not, perhaps @rmccue knows?

Shelob9:
@joehoyle I don't know about intentional. I would argue it's not useful information in that context. That information is available via other means that make more sense.

danielbachhuber:
Closing in favor of https://core.trac.wordpress.org/ticket/34647, as this needs to be addressed before WP 4.4 ships.

PR #1622, opened 2015-10-01
Don't return route URL

Shelob9:
#1621


lib/infrastructure/class-wp-rest-request.php
```diff
@@ -446,7 +446,8 @@ public function get_params() {
 		foreach ( $order as $type ) {
 			$params = array_merge( $params, (array) $this->params[ $type ] );
 		}
-
+
```
rmccue: This line has trailing whitespace, which is causing the standards check to fail :)

Shelob9: That's annoying. I fixed.

PR #1626, opened 2015-10-04
Add license in package.json, bump phplint package version

mattbanks:
Ran `grunt phplint` before and after upgrading package and output remains the same (no linting errors).

GPL-2.0 license added to avoid NPM warning when running NPM install.

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,7 +1,7 @@
 {
   "name": "wp-api",
   "version": "0.0.0",
-  "license" : "GPL-2.0",
+  "license" : "GPL-2.0+",
   "repository": {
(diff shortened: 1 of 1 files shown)


ntwb:
This should be the same license as core which is `GPL-2.0+` rather than `GPL-2.0`

See https://core.trac.wordpress.org/ticket/32718

mattbanks:
Ahh, I didn't see that you can have a + in there now. I'll update the PR!

danielbachhuber:
Tests are failing for an unrelated reason. PHP lint tests pass. Thanks for the PR!

Issue #1055, opened 2015-04-03
Remove password-protected posts


danielbachhuber:
Setting and unsetting a cookie within a RESTful API is quite ugly. We should probably internalize core's logic if core isn't going to. 475e041ff865ffe2f0102156d29f2d3e0cdfc3f1

Also, we should show password-protected posts with "Protected: `<title>`"

Previously #313

danielbachhuber:
Looks like Ryan started this in #939

danielbachhuber:
Some conversation from today's hangout:

* Core's cookie handling is per `COOKIE_HASH`, not post, so viewing a second password-protected post nullifies the first.
* We can't pass the password as a `GET parameter because it could expose the password in the logs.
* WordPress.com accepts the password as a `GET` parameter. This is likely an unacceptable risk for core.
* Ryan suggested authorization headers instead of parameters.
* Another open question: what fields are expected to be password-protected?
* We should err on the side of less data (e.g. not in the loop)

rmccue:
> We can't pass the password as a `GET parameter because it could expose the password in the logs.

One thing I forgot to mention here too: not just full server admins but also admins on a multisite network might have access to these.

mnelson4:
>Ryan suggested authorization headers instead of parameters.
Authorization headers seems like quite a natural RESTy way provide the password.

>what fields are expected to be password-protected?
Given how much data is shown on the archives page for password-protected posts, I think its almost better to consider what fields are expected to NOT be password-protected? In twenty-fourteen theme, I can only see the event's author, date, post_name, post_title, and categories (and kindof infer the post_type) in the archives page.

re #1100: should authorized users be able to access password-protected posts without providing the password? (eg a user who has the "read_private_posts" can access posts without providing a password in the WP admin, but on the posts archive page they DO need to provide a password). 
It seems that generally the API should display whatever the current user can access from either the frontend or backend, so if they can view the post without providing a password in the admin, the API should do the same. Agreed?

nschoenholtz:
I'm curious what the thinking behind the current handling is, because it seems a little bit backward to me -- it keeps password protected posts out of queries, but reveals all of the content if requested directly (vs in core, the posts would be included in the query but the content would be hidden).

I definitely support the authorization header approach -- would a pull request be welcome here, or is there a larger approach already in the works?

danielbachhuber:
More discussion: https://wordpress.slack.com/archives/core-restapi/p1449533485001559

[...]

PR #1630, opened 2015-10-06
add a filter to allow custom handling of post password behavior

nschoenholtz:
I've come around to agreeing with the default behavior for password protected posts (https://github.com/WP-API/WP-API/issues/1055), but do need to change that behavior for one project. Adding a filter here seems to be enough to let me introduce the rest of the custom handling.


danielbachhuber:
We won't add a filter until we come to some official determination for #1055. I'd recommend subclassing the Posts controller for now, and re-registering the endpoint.

Related https://github.com/WP-API/docs-v2/issues/61

Thanks for the pull request, though.

Issue #1614, opened 2015-09-30
Multisite activation process doesn't restore switch stack correctly


rmccue:
Waaaay back in #49, we added a special case for multisite activation. It uses a loop which switches to sites before flushing, then after the loop, restores the current site.

As it turns out, this has a bug. `restore_current_blog` only restores once up the stack, so this needs to be inside the loop instead. Oops :)

rmccue:
This is now in core, so it's up to core's upgrade processes to handle.

PR #1633, opened 2015-10-08
1614-multisite-activation-stack-fix

JohnDittmar:
Addresses: #1614

Quick fix to put `restore_current_blog()` inside of the subsite
loop so that the switch stack is restored on every blog.

http://wordpress.stackexchange.com/questions/89113/restore-current-blog-vs-switch-to-blog/123516#123516


rmccue:
Thanks for the pull request! The change here includes a change to a submodule, but there's no corresponding PR on the [api-core](https://github.com/WP-API/api-core) repo, so we can't merge this.

I _think_ this is also now not an issue, as the code has been merged into core, and it's up to core's upgrade processes to ensure this is all sorted. Thanks for the PR anyway!

Issue #1613, opened 2015-09-29
Searching terms by parent produces error if none found


jasonphillips:
When querying terms with the `parent` parameter set, I'm getting a fatal PHP error (_Unsupported operand types_) for cases in which it should be an empty array of no results. The line in question is third line here:

```php
$total_terms = wp_count_terms( $this->taxonomy, $prepared_args );
$response->header( 'X-WP-Total', (int) $total_terms );
$max_pages = ceil( $total_terms / $request['per_page'] );
```

In **class-wp-rest-terms-controller.php**:
https://github.com/WP-API/WP-API/blob/develop/lib/endpoints/class-wp-rest-terms-controller.php#L105

Since `max_pages` is set to the result of `wp_count_terms`, it can sometimes be an empty array instead of an integer, which then causes the division operation to fail. I've found that an explicit cast to **int** corrects the error.

```php
$max_pages = ceil( (int) $total_terms / $request['per_page'] );
```

joehoyle:
@jasonphillips thanks for the report! Fixed in PR #1641

PR #1641, opened 2015-10-10
Handle falsy value from wp_count_terms, fixing fatal

joehoyle:
See #1613


lib/endpoints/class-wp-rest-terms-controller.php
```diff
@@ -101,6 +101,12 @@ public function get_items( $request ) {
 		unset( $prepared_args['number'] );
 		unset( $prepared_args['offset'] );
 		$total_terms = wp_count_terms( $this->taxonomy, $prepared_args );
+
+		// wp_count_terms can return a falsy value when the term has no children
```
danielbachhuber: Should we submit a core ticket for this?

PR #1644, opened 2015-10-12
Fix tests with latest trunk

### LINKED ISSUE #1639: Update test suite
I noticed Travis is failing due to the recent changes in core and thought it would be good to update the test suite for future proofing the REST API, therefore we can track it with this issue.

rmccue:
Fixes #1639.


rmccue:
@WP-API/amigos Should be good to go; main substance is changing the taxonomy check. `get_term_by` started always returning the term in [r34997](https://core.trac.wordpress.org/changeset/34997) when you pass `term_taxonomy_id`, even if the taxonomy doesn't match, so we need to double-check.

PR #1656, opened 2015-10-17
Update get items to allow filter for users collection

timnashcouk:
Currently filtering users such as ?filter[per_page]=1 does not work, as $request->get_param is matching to ['get']['per_page'] not ['get']['filter']['per_page'] this patch matches the method used to allow filtering in posts. Long term a change to get_param might be a better solution.

### DIFF
--- a/lib/endpoints/class-wp-rest-users-controller.php
+++ b/lib/endpoints/class-wp-rest-users-controller.php
@@ -80,9 +80,9 @@ class WP_REST_Users_Controller extends WP_REST_Controller {
 	public function get_items( $request ) {
 
 		$prepared_args = array();
+
 		$prepared_args['order'] = $request['order'];
-		$prepared_args['number'] = $request['per_page'];
-		$prepared_args['offset'] = ( $request['page'] - 1 ) * $prepared_args['number'];
+		$prepared_args['per_page'] = $request['per_page'];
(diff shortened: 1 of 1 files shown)


timnashcouk:
Patch for #1657

joehoyle:
@timnashcouk this is by-design. `filter` is essentially a wrapper for WP_Query publis query vars, whereas there are also standardised query params across all collections (posts, terms, users, comments) such as `per_page` and `page`.

If you want to specify pagination in the `filter` param, one would use the WP_Query public query vars of `posts_per_page` and `paged`.

The intention longer term will be to have consistent collection query args across all the resources, e.g. `/posts/?per_page=10`

timnashcouk:
@joehoyle problem is the equivalent to posts_per_page and paged within users, is number and offset. Which are being diverted away so you end up with a very weird inconsistency. Which this patch only partially solves (I've just realised) as I actually haven't free'ed up number and offset.

Ideally the ability to override offset and number should be available through filter as it's a standard part of the User loop, regardless of any standardisation within the API.

Same issue occurs with search which is also being overridden currently so you end up with
/users/?search=*example.com&filter%5Bsearch_columns%5D%5B0%5D=user_email
rather then search also being a filter.

PR #1659, opened 2015-10-19
Fix CORS bug when preflight OPTIONS request sent + Mitigate the Abusing of JSONP with Rosetta Flash

Odyno:
## Fix CORS bug when preflight OPTIONS request sent

When app uses "application/json; charset=utf-8" content type when make POST request through CORS. When app use text/plain there are no preflight OPTIONS, but with "application/json; charset=utf-8" it occurs.
Bug is that preflight request finished, but POST not happends; Google Chrome outputs:
Request header field Content-Type is not allowed by Access-Control-Allow-Headers.

More Info:
- http://stackoverflow.com/questions/12630231/how-do-cors-and-access-control-allow-headers-work
- Here is described how and when preflight sends: https://developer.mozilla.org/en-US/docs/Web/HTTP/Access_control_CORS
## Mitigate the Abusing of JSONP with Rosetta Flash

More info  https://miki.it/blog/2014/7/8/abusing-jsonp-with-rosetta-flash/

### DIFF
--- a/plugin.php
+++ b/plugin.php
@@ -721,7 +721,8 @@ function json_send_cors_headers( $value ) {
 		header( 'Access-Control-Allow-Origin: ' . esc_url_raw( $origin ) );
 		header( 'Access-Control-Allow-Methods: POST, GET, OPTIONS, PUT, DELETE' );
 		header( 'Access-Control-Allow-Credentials: true' );
-		header( 'Access-Control-Allow-Headers: origin, content-type, accept, X-WP-Total, X-WP-TotalPages');
+		header( 'Access-Control-Allow-Headers: origin, content-type, accept');
+		header( 'X-Content-Type-Options: nosniff');
 	}
 
 	return $value;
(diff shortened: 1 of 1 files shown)


lesterchan:
Wow thanks for this! I ran into the exact same problem and the Chanegset works.

lesterchan:
If anyone is looking for the same thing, this is the code I used without the need to modify the core. The fix on 2.0 beta 7 doesn't solve the problem (at least for me).
```php
    add_filter( 'rest_pre_serve_request', 'custom_rest_pre_serve_request' );
    public function custom_rest_pre_serve_request( $value ) {
        $origin = get_http_origin();

        if ( $origin ) {
            header( 'Access-Control-Allow-Headers: origin, content-type, accept, authorization');
            header( 'X-Content-Type-Options: nosniff');
        }

        return $value;
    }
```

danielbachhuber:
This is fixed in core infrastructure.

PR #1661, opened 2015-10-20
Switch core submodule checkout to SSH instead of http

danielbachhuber:
The former is necessary for SSH-based authentication

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -1,3 +1,3 @@
 [submodule "core"]
 	path = core
-	url = https://github.com/WP-API/api-core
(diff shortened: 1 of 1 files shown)


danielbachhuber:
@WP-API/amigos #reviewmerge

danielbachhuber:
Hm. Do we need to generate a SSH key for checking out a public repo?

danielbachhuber:
I suppose we can :football: on this, as infrastructure changes should be going to Trac and not the `api-core` repo. I submitted originally because I wanted to be able to make PRs against the `api-core` repo using SSH auth.

PR #1663, opened 2015-10-20
Fix failing test

danielbachhuber:
Core's test suite began backing up and restoring public and private
query vars in https://core.trac.wordpress.org/changeset/35258/ This
means we need to register our custom query var each time we want to
reference it.


danielbachhuber:
Just to clarify, the test doesn't fail when you call it directly, because `rest_api_init` is called in the `init` hook. However, it fails when called as the second or later test, because `init` doesn't get called on subsequent tests.

PR #1666, opened 2015-10-20
Expose `slug` in `embed` context for Users

### LINKED ISSUE #1611: Expose user slug via embed
As it turns out, you kind of need the user slug for things like autocomplete. We already do similar for posts with exposing the slug in embed, and it's already available in the URL, so there's no real privacy or consistency concerns (apart from being _more_ consistent).

danielbachhuber:
`user_nicename` is a public attribute, used in user URLs, so this is
safe data to present

Fixes #1611


danielbachhuber:
@WP-API/amigos #reviewmerge

joehoyle:
Makes sense

PR #1667, opened 2015-10-20
Include `media_details` attribute for attachments in embed context

danielbachhuber:
For image attachments, `media_details` includes a `sizes` array of image
sizes, which is useful for templating.

Previously #1582


danielbachhuber:
@WP-API/amigos #reviewmerge

joehoyle:
I'd be +1 on this, but I'll let others weigh in here

joehoyle:
Going to roll with!

PR #1670, opened 2015-10-20
Make error message more helpful by specifying method to subclass

### LINKED ISSUE #1546: Make the "Method not implemented" message specific
It was a bit of a headache to extend the WP_REST_Controller class when I was seeing the "Method not implemented. Must be over-ridden in subclass." message. While developing, I had to add  __FUNCTION__ to the message so I could tell which functions I was missing.

danielbachhuber:
Fixes #1546


danielbachhuber:
@WP-API/amigos #reviewmerge

PR #1679, opened 2015-10-22
Remove page template from schema

joehoyle:
This function is no longer available via the API, see https://core.trac.wordpress.org/changeset/35353

Need to work out what to do on https://github.com/WP-API/WP-API/blob/develop/lib/endpoints/class-wp-rest-posts-controller.php#L892

### DIFF
--- a/lib/endpoints/class-wp-rest-posts-controller.php
+++ b/lib/endpoints/class-wp-rest-posts-controller.php
@@ -1529,7 +1529,6 @@ class WP_REST_Posts_Controller extends WP_REST_Controller {
 			$schema['properties']['template'] = array(
 				'description' => 'The theme file to use to display the object.',
 				'type'        => 'string',
-				'enum'        => array_values( get_page_templates() ),
 				'context'     => array( 'view', 'edit' ),
 			);
 		}


joehoyle:
@WP-API/amigos as noted above, what should we do on https://github.com/WP-API/WP-API/blob/develop/lib/endpoints/class-wp-rest-posts-controller.php#L892

danielbachhuber:
> as noted above, what should we do on

I think some form of loose validation that it's a proper file. However, being able to supply an arbitrary file path means arbitrary code execution, right?

rachelbaker:
@joehoyle @danielbachhuber @rmccue I think we should remove the ability to set/update the page template in the interest of shipping beta5 and fixing fatal errors with core's 4.4 Beta 1.

danielbachhuber:
We can also just load all of the admin files in the plugin, which is what all previous versions of the plugin have done, and solve the problem of conditionally loading admin APIs later.

joehoyle:
I think we should load the admin in the update callback for pages templates. Also need to find out why this doesn't fatal the unit tests



Joe Hoyle
Technical Lead
humanmade

e joe@humanmade.co.uk
t +44 (0) 7891 685613
skype joe-hoyle
w humanmade.co.uk

danielbachhuber:
:football: on this approach in favor of loading admin utils.

PR #1682, opened 2015-10-23
Include the admin functions for the API

joehoyle:
We should look at fixing this, but the core commit at 
https://core.trac.wordpress.org/changeset/35353 caused a fatal in several
cases.


danielbachhuber:
:cherry_blossom:

PR #1687, opened 2015-10-23
Prioritize WP-API v1.x rewrite rules over core REST API

danielbachhuber:
When sites using WP-API v1.x upgrade to WordPress 4.4, they'll expect the plugin to continue working.

Originally https://core.trac.wordpress.org/ticket/34399


mnelson4:
This worked for me. I suppose the only downside will be if folks expect both WP API 1.2 AND the WP API in core to work simultaneously. But that's not a huge deal

Issue #1675, opened 2015-10-22
The date field of `GET /comments` should contain either a 'Z' time offset signature.


southp:
## Environment
WordPress: 4.3.1
WP REST API: 2.0-beta 4
Machine: Chassis with VirtualBox 4.3.28 as its VM backend

## How To
Using the following snippet:

```
#!/bin/sh
http POST <url> author_name='Shell' \
    author_email='shell@xd.cc' \
    author_url='http://goog.le' \
    content='wp rest api rocks!' \
    date='2015-10-22T12:00:00Z' \
    post=<pid> \
    parent=0
```
The date field of the response will be `2015-10-22T12:00:00`, so does `GET /comments`.

According to the [ECMAScript 2015 spec](http://ecma-international.org/ecma-262/6.0/index.html#sec-date-time-string-format):

```
If the time zone offset is absent, the date-time is interpreted as a local time.
```

Thus, the returned date string is a local time string instead of the originally-submitted UTC time string for any browser implementing this spec.

danielbachhuber:
Thanks for the detailed report.

Dates are returned as ISO8601/RFC3339: https://github.com/WP-API/WP-API/blob/8f55847b87dfb09428ef86ffb9561199d8410c61/lib/endpoints/class-wp-rest-comments-controller.php#L473-L474

The timezone is omitted because:

> Explicitly strips timezones, as datetimes are not saved with any timezone information. Including any information on the offset could be misleading.

This decision was made in 6d0ad766ca525c7a3aee921b66554ddf21a5c023. I'll flag this for discussion so we can reverify the decision.

You can use the `date_gmt` value instead, but we'll need to expose it in the view context first. #1690

southp:
Hi @danielbachhuber ,
Thanks for bringing it further. As long as there is no timezone information saved in the first place, I think it is still the right call to not including such an information. Exposing `date_gmt` value fits my purpose very well, as a plus it also disambiguates the purpose of `date`.

Lastly, [the document](http://v2.wp-api.org/reference/comments/) states the `date` as:
```
The date the object was published.
```
I think it might be clearer if we put it as something like:
```
The date the object was published in the timezone as the blog settings.
```

danielbachhuber:
> I think it might be clearer if we put it as something like:

Sounds good — submitted in #1715

southp:
> Sounds good — submitted in #1715

Thanks for the help. It looks even better :+1:

rmccue:
[...]

PR #1690, opened 2015-10-25
Expose `date_gmt` for `view` context of Posts and Comments

danielbachhuber:
It can be useful for determining the absolute time for the resource.

`date_gmt` was set for `edit` context only in
29c768b93a51a14cdf4031493f03f798cc1f1730 and
a12c23a76a747a074083094e76dd169a2b335448

See #1675

### DIFF
--- a/lib/endpoints/class-wp-rest-comments-controller.php
+++ b/lib/endpoints/class-wp-rest-comments-controller.php
@@ -812,7 +812,7 @@ class WP_REST_Comments_Controller extends WP_REST_Controller {
 					'description'  => 'The date the object was published as GMT.',
 					'type'         => 'string',
 					'format'       => 'date-time',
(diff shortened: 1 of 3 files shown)


danielbachhuber:
@WP-API/amigos #reviewmerge

@rachelbaker Do you recall why you moved `comment_date_gmt` to `edit` context in 29c768b93a51a14cdf4031493f03f798cc1f1730 ?

rachelbaker:
@danielbachhuber the original philosophy I was trying to adhere to was is this property used for front-end public display?  I have never seen a theme that displayed the `post_date` and separately also `post_date_gmt`.  
That being said, I notice that we do expose `modified_gmt` for Posts....so we should be consistent one way or the other.  I am :+1: with including the `_gmt` dates in the view context.

Issue #1694, opened 2015-10-26
Remove v1 compatibility shim from v2


joehoyle:

danielbachhuber:
What's the context for this? I thought we told everyone their v1.x endpoints would work in v2.x?

joehoyle:
@danielbachhuber that compat shim will exist in v1.9.x

PR #1695, opened 2015-10-27
Remove compat shim, this will instead be in v1

joehoyle:
See #1694


danielbachhuber:
Should we start a `release-1-9` branch from `develop` before this lands?

joehoyle:
`release-1-9` will be from the version 1 branch, not develop I believe.

rachelbaker:
@danielbachhuber My understanding of the intended purpose of v1.9 is for v1.x endpoints of the API to be compatible with WP Core, not between v2.x and and v1.x.

PR #1697, opened 2015-10-27
Avoid expensive call to get_item() in post terms controller

joehoyle:
Only call the permissions check from terms controller

See #1583


rachelbaker:
The changes added in 6e2c731 are based on post-meta controller

rachelbaker:
@WP-API/amigos #reviewmerge

PR #1699, opened 2015-10-28
Correct the DocBlocks for Requires and Includes

ahmadawais:
This commit Corrects the DocBlocks for Requires and Includes in accordance with WP PHP documentation standards https://make.wordpress.org/core/handbook/best-practices/inline-documentation-standards/php/#3-requires-and-includes


danielbachhuber:
Thanks!

PR #1701, opened 2015-10-28
Better File Header Documentation

ahmadawais:
Better File Header Documentation
- Adds @since 4.4.0
- Separates short and long descriptions

### DIFF
--- a/extras.php
+++ b/extras.php
@@ -4,7 +4,6 @@
  *
  * Contains extra functions from plugin.php go.
  *
- * @since 4.4.0
  * @package WordPress
  * @subpackage JSON API
  */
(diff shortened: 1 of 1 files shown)


extras.php
```diff
@@ -1,12 +1,12 @@
 <?php
-
 /**
- * Extra File where a lot of the extra functions from plugin.php go.
+ * Extra File
+ *
+ * Contains extra functions from plugin.php go.
  *
+ * @since 4.4.0
```
danielbachhuber: This is actually incorrect, because the file isn't included in WordPress core, and the plugin isn't at 4.4.0

ahmadawais: I added this, since the other function in this file had the same @since tag. Let me remove that.

ahmadawais:
@danielbachhuber Removed the `since` tag.

PR #1702, opened 2015-10-28
Add Periods to end comments, @params and @returns

ahmadawais:
In accordance with WP PHP documentation standards Add Periods to end comments, @params and @returns

### DIFF
--- a/lib/endpoints/class-wp-rest-controller.php
+++ b/lib/endpoints/class-wp-rest-controller.php
@@ -14,7 +14,7 @@ abstract class WP_REST_Controller {
 	 * Get a collection of items.
 	 *
 	 * @param WP_REST_Request $request Full data about the request.
-	 * @return WP_Error|WP_REST_Response.
+	 * @return WP_Error|WP_REST_Response
 	 */
 	public function get_items( $request ) {
 		return new WP_Error( 'invalid-method', sprintf( __( "Method '%s' not implemented. Must be over-ridden in subclass." ), __METHOD__ ), array( 'status' => 405 ) );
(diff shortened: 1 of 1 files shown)


lib/endpoints/class-wp-rest-controller.php
```diff
@@ -11,121 +11,121 @@ public function register_routes() {
 	}
 
 	/**
-	 * Get a collection of items
+	 * Get a collection of items.
 	 *
 	 * @param WP_REST_Request $request Full data about the request.
-	 * @return WP_Error|WP_REST_Response
+	 * @return WP_Error|WP_REST_Response.
```
danielbachhuber: I haven't ever seen periods at the end of the type, only when there's a comment too. Can you clarify where else in core follows this pattern?

ahmadawais: Hey, @danielbachhuber !

Periods at the end of the type are the part of standard WordPress PHP documentation in the Core handbook. You can check this information here https://make.wordpress.org/core/handbook/best-practices/inline-documentation-standards/php/#1-functions-class-methods

danielbachhuber: I read:

> @return: Should contain all possible return types, and a description for each. Use a period at the end. Note: @return void should not be used outside of the default bundled themes.

This `@return` doc doesn't have a description, so I don't think it needs a period.

ahmadawais: @danielbachhuber  Though it does mention ` Use a period at the end.` 
What do you think about that? Let me know. I'll follow your lead.

danielbachhuber: Only add a period if there's actually a description, please. Better yet, add descriptions where they're missing, and then you can add a period to that :)

ahmadawais: @danielbachhuber Got it, re-submitting the corrected version.

ahmadawais:
@danielbachhuber Kindly, take a look now.

danielbachhuber:
:ship: :ship: :ship:

PR #1703, opened 2015-10-28
End comments in periods

ahmadawais:
End comments in periods


joehoyle:
A full stop after the params list seems odd, is that standard?

ahmadawais:
@joehoyle  I have just updated the PR. 
I had misread the some DocBlocks. Now, there is a period after `param` and `returns` only when there exists a description.

Take a look.

PR #1708, opened 2015-10-31
Change users schema order to be order of importance instead of alpha

rachelbaker:


rachelbaker:
@WP-API/amigos #reviewmerge

danielbachhuber:
> order of importance 

How did you determine this?

rmccue:
:+1:

@danielbachhuber At a guess, just a design decision.

rachelbaker:
@danielbachhuber how did you determine the order of the other endpoint schemas?

I moved `id` to the top and then tried to order the other properties by making educated guesses on likely use.

I can see there are some things I missed here:
 * `url` should probably move up
 * `capabilities` and `extra_capabilities` should be under `role` and `roles`

I will refresh this PR when I can.

danielbachhuber:
@WP-API/amigos #reviewmerge

I moved `url` to under `email`, where it appears in the User Profile in the dashboard. I already see `capabilities` and `extra_capabilities` to be under `role` and `roles`.

PR #1709, opened 2015-11-01
Update the wp-api.js client from the client-js repo

rachelbaker:


rachelbaker:
@WP-API/amigos #reviewmerge

rachelbaker:
Closes #1576

PR #1711, opened 2015-11-02
Change featured image relation to featuredmedia

### LINKED ISSUE #1476: Featured image is not in _links on post response

rmccue:
Fixes #1476 again.

### DIFF
--- a/lib/endpoints/class-wp-rest-posts-controller.php
+++ b/lib/endpoints/class-wp-rest-posts-controller.php
@@ -1210,19 +1210,17 @@ class WP_REST_Posts_Controller extends WP_REST_Controller {
 		}
 
 		// If we have a featured image, add that.
-		$links['http://v2.wp-api.org/attachment'] = array();
 		if ( $featured_image = get_post_thumbnail_id( $post->ID ) ) {
 			$image_url = rest_url( 'wp/v2/media/' . $featured_image );
-			$links['http://v2.wp-api.org/attachment'][] = array(
+			$links['http://v2.wp-api.org/featuredmedia'] = array(
 				'href'       => $image_url,
 				'embeddable' => true,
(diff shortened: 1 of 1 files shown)


lib/endpoints/class-wp-rest-posts-controller.php
```diff
@@ -1210,19 +1210,17 @@ protected function prepare_links( $post ) {
...
-			$links['http://v2.wp-api.org/attachment'][] = array(
+			$links['http://v2.wp-api.org/featuredmedia'] = array(
 				'href'       => $image_url,
 				'embeddable' => true,
-				'featured'   => true,
 			);
 		}
 		if ( ! in_array( $post->post_type, array( 'attachment', 'nav_menu_item', 'revision' ) ) ) {
 			$attachments_url = rest_url( 'wp/v2/media' );
 			$attachments_url = add_query_arg( 'post_parent', $post->ID, $attachments_url );
```
danielbachhuber: Shouldn't this be `parent` ?

rmccue: Potentially; I didn't change it here. Will look in the morning, good spot.

Cyberfolks:
For ionic 3
if you are using _embed
like 
/?rest_route=/wp/v2/posts&_embed
or
/wp-json/wp/v2/posts?_embed

then try
Method= array_name._embedded['wp:featuredmedia']['0'].source_url
Example=  post._embedded['wp:featuredmedia']['0'].source_url

Issue #1520, opened 2015-09-01
Issues with is_numeric validator


ChrisWiegman:
When testing with a parameter using the is_numeric function to validate I'm getting the following:

Warning: is_numeric() expects exactly 1 parameter, 3 given in [...]/plugins/rest-api/lib/infrastructure/class-wp-rest-request.php on line 684

The odd thing is this hadn't happened in previous testing and seems to be a new problem. Thoughts?

joehoyle:
I've experienced this issue with `floatval` too, even though it doesn't document any more than one param, passing too many arguments makes it return false. I'm not sure of the best way to solve this.

danielbachhuber:
@joehoyle @ChrisWiegman where is `is_numeric()` being called with three params?

joehoyle:
yeah I've run into this too, not sure if there is much we can do. The API will pass `$value, $param, $request` to the validate / sanitize callbacks, this causes issues with inbuilt functions like `is_numeric`. My work around has been to do: `validate_callback => function( $v ) { return is_numeric( $v ); }`

danielbachhuber:
I'd like to address this in the infrastructure https://core.trac.wordpress.org/ticket/34659

BE-Webdesign:
https://core.trac.wordpress.org/ticket/34659 With the diff looks good to me.

danielbachhuber:
Closing in favor of the core ticket. `is_numeric()` doesn't affect the plugin itself at the moment.

PR #1712, opened 2015-11-02
Avoid calling sanitize_title directly

### LINKED ISSUE #1585: rest_invalid_param if post slug is empty string
> Invalid parameter(s): slug (slug is not of type string)

This should use the automatically generated one instead.

rmccue:
Fixes #1585.

@WP-API/amigos Question: should we do this, or use `create_function( '$a', 'return sanitize_title($a);' )` ? I'm looking through, and it seems like there might be other places we need to do this too (`sanitize_user` e.g. takes a `$strict` parameter as second param).

### DIFF
--- a/lib/endpoints/class-wp-rest-comments-controller.php
+++ b/lib/endpoints/class-wp-rest-comments-controller.php
@@ -923,7 +923,7 @@ class WP_REST_Comments_Controller extends WP_REST_Controller {
 		$query_params['post_slug'] = array(
 			'default'           => null,
 			'description'       => 'Limit result set to comments associated with posts of a specific post slug.',
-			'sanitize_callback' => 'sanitize_title',
+			'sanitize_callback' => array( $this, 'sanitize_slug' ),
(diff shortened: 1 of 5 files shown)


rmccue:
Just did an audit, and `sanitize_user` is literally the only other one that has a second parameter.

danielbachhuber:
No to `create_function`.

`is_numeric()` has a similar problem #1520. Rather than create a public method for each, what if we had a utility function for calling another function with only one argument? Trying to think of how we can solve this in a more abstracted, less add-a-bunch-of-code manner.

joehoyle:
A wrapper is an interesting idea - I don't like the `create_function` example, and I don't like only passing the value to the `validate_callback`. Other than having a nasty whitelist of functions that we only pass a single value to, I don't see much alternative.

danielbachhuber:
Closing in favor of solving this in infrastructure: https://core.trac.wordpress.org/ticket/34659

PR #1714, opened 2015-11-03
Provide object type as callback argument

jtsternberg:
If you use `register_api_field` with an array of object types, then there does not seem to be a decent way to know what context these callbacks are being called from. This update passes the object type through to the callback function.


jtsternberg:
To illustrate the issue, here is a snippet of code wherein I am trying to register a cmb2 field update callback, and have to try and figure out the object type and object id:

danielbachhuber:
Seems reasonable — thanks!

PR #1715, opened 2015-11-03
Clarify what `date` and `modified` actually are

### LINKED ISSUE #1675: The date field of `GET /comments` should contain either a 'Z' time offset signature.

danielbachhuber:
They're the date, but in the site's timezone.

Fixes #1675

### DIFF
--- a/lib/endpoints/class-wp-rest-posts-controller.php
+++ b/lib/endpoints/class-wp-rest-posts-controller.php
@@ -1276,7 +1276,7 @@ class WP_REST_Posts_Controller extends WP_REST_Controller {
 			 */
 			'properties' => array(
 				'date'            => array(
-					'description' => 'The date the object was published.',
+					'description' => "The date the object was published, in the site's timezone.",
 					'type'        => 'string',
 					'format'      => 'date-time',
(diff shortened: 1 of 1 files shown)


danielbachhuber:
@WP-API/amigos #reviewmerge

PR #1727, opened 2015-11-10
Run tests against WP nightly and WP latest (4.3)

### LINKED ISSUE #1685: Run Travis against WordPress 4.3
Because we're expecting users to be writing v2.0-beta endpoints, but they're probably not running trunk in production, we need to make sure beta releases are compatible with the latest stable release of WordPress.

From #1683

danielbachhuber:
Fixes #1685

### DIFF
--- a/plugin.php
+++ b/plugin.php
@@ -84,7 +84,7 @@ include_once dirname( __FILE__ ) . '/lib/endpoints/class-wp-rest-posts-terms-con
  */
 include_once( dirname( __FILE__ ) . '/extras.php' );
 
-add_filter( 'register_post_type_args', '_add_extra_api_post_type_arguments', 10, 2 );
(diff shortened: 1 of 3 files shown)


danielbachhuber:
@WP-API/amigos Thoughts? We'll need to land https://github.com/WP-API/api-core/pull/6 for core compatibility, but that looks like the only hack we need.

plugin.php
```diff
@@ -84,7 +84,7 @@
  */
 include_once( dirname( __FILE__ ) . '/extras.php' );
 
-add_filter( 'register_post_type_args', '_add_extra_api_post_type_arguments', 10, 2 );
+add_filter( 'init', '_add_extra_api_post_type_arguments', 11 );
```
rmccue: I actually like this approach better anyway, since it means we're not running a bunch of times when we don't need to.

rmccue:
:+1:

Issue #1733, opened 2015-11-13
The filter parameter does not work


devotox:
It seems to not actually filter any posts and also it then causes the page and per_page parameters to no longer work when the filter parameter is present in the url

danielbachhuber:
Can you share the request you're making, the version of the plugin you're using, and the actual result you see?

devotox:
http://doctifyblog.com/wp-json/wp/v2/posts?page=1&per_page=1&filter=blah

I am on the latest beta and hitten the link will show you the actual result

danielbachhuber:
> I am on the latest beta

Do you have any PHP filters that are modifying the request, by chance?

`filter` works as expected for me locally:

```
local ➜  wp-api git:(develop) http wordpress-develop.dev/wp-json/wp/v2/posts | jq '.[].id'
100
99
98
97
96
95
94
93
92
91
local ➜  wp-api git:(develop) http wordpress-develop.dev/wp-json/wp/v2/posts per_page==1 | jq '.[].id'
100
local ➜  wp-api git:(develop) http wordpress-develop.dev/wp-json/wp/v2/posts per_page==1 filter\[posts_per_page\]==2 | jq '.[].id'
100
99
local ➜  wp-api git:(develop) http wordpress-develop.dev/wp-json/wp/v2/posts per_page==1 page==2 filter\[posts_per_page\]==2 | jq '.[].id'
98
97
```

`filter` will override `per_page` and `page` if `filter[posts_per_page]` or `filter[paged]` are provided https://github.com/WP-API/WP-API/blob/8e1cb40a9cc2b451f970044fc4e0ad76f16fffda/lib/endpoints/class-wp-rest-posts-controller.php#L93

devotox:
Could you give me an example of how you would pass the filter as part of
the request. Is it just a string as a query parameter?

danielbachhuber:
`GET http://wordpress-develop.dev/wp-json/wp/v2/posts?per_page=1&filter[posts_per_page]=2`

devotox:
Thank you but how would you actually search the text of each post and only
return matches based on the search parameter

danielbachhuber:
`GET http://wordpress-develop.dev/wp-json/wp/v2/posts?filter[s]=search+string`

devotox:
Yes if you see from the URL I sent you I tried exactly this with
filter=search string and I just got returned the first ten posts no matter
the filter passed in

danielbachhuber:
[...]

PR #1734, opened 2015-11-13
Only process `filter` when it's an array

danielbachhuber:
The value is still set if `&filter` is passed, but `array_merge()` will
error if `$request['filter']` isn't an array.

From #1733


danielbachhuber:
@WP-API/amigos #reviewmerge

Issue #1735, opened 2015-11-15
How to GET only top level terms (without parent)


equatorapps:
We are using WP-API v2 beta-5 and the following path: wp-json/wp/v2/terms/product_cat/ to get the woo commerce categories we have entered, and there are many. I have tried to use parent=0, but that does nothing. parent=xxx only seems to work for whole numbers. Is there another parameter or path to get just   terms without a parent? Thanks.

swissspidy:
According to the docs and [the code](https://github.com/WP-API/WP-API/blob/develop/lib/endpoints/class-wp-rest-terms-controller.php#L82) `parent=0` should work.

Perhaps the `0 === $request['parent']` check fails because it's not an int, so that would be a bug.

Toddses:
There's a bug in the terms controller beta-5 which prevents you from making a request for only top-level terms (ie. terms with parent 0).

A fix landed in beta-6. So if you update, the `wp-json/wp/v2/terms/product_cat?parent=0` endpoint should do the trick.

equatorapps:
Thanks guys! I have updated to beta6, but my results are the same as before. parent=0 returns all terms. parent=234 returns only terms with that parent id.

danielbachhuber:
> Perhaps the 0 === $request['parent'] check fails because it's not an int, so that would be a bug.

This is the problem. It was introduced in #1604. We'll fix this in 2.0 Beta 6.1

PR #1739, opened 2015-11-16
Properly handle parent=0 when 0 is a string

danielbachhuber:
See #1735


danielbachhuber:
@WP-API/amigos #reviewmerge

This merges into `develop`. I'm going to leave #1735 open until we release Beta 6.1, as we'll need to remember to cherry-pick the merge commit onto a release branch.

PR #1742, opened 2015-11-18
Load `api-core` as a Composer dependency if it exists

### LINKED ISSUE #1736: Files from "core" dir are not found when installed with Composer
This issue may have already been reported, but I couldn't find it.

I installed WP-API using Composer:

    {
        "require": {
            "wp-api/wp-api": "2.0-beta6"
        }
    }

And I now get this error on every Wordpress page:

> PHP Fatal error:  require_once(): Failed opening required '/var/www/my-website/wordpress/wp-content/plugins/json-rest-api/core/rest-api.php' (include_path='.:/usr/share/php:/usr/share/pear') in /var/www/my-website/wordpress/wp-content/plugins/json-rest-api/plugin.php on line 14

I was using WP API 2.0 beta 4 berore and i didn't have the issue, I think it appeared in beta 5 when the `core` folder was moved to another repository.

danielbachhuber:
Fixes #1736

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -15,7 +15,8 @@
         "forum": "https://wordpress.org/support/plugin/json-rest-api"
     },
     "require": {
-        "composer/installers": "~1.0"
+        "composer/installers": "~1.0",
+        "wp-api/api-core": "2.0-beta6"
     },
     "require-dev": {
         "wp-coding-standards/wpcs": "0.6.0"
--- a/plugin.php
+++ b/plugin.php
@@ -11,7 +11,11 @@
 
 // Do we need the compatibility repo?
 if ( ! defined( 'REST_API_VERSION' ) ) {
-	require_once dirname( __FILE__ ) . '/core/rest-api.php';
+	if ( file_exists( dirname( __FILE__ ) . '/core/rest-api.php' ) ) {
+		require_once dirname( __FILE__ ) . '/core/rest-api.php';
+	} else if ( file_exists( dirname( __FILE__ ) . '/vendor/wp-api/api-core/rest-api.php' ) ) {
+		require_once dirname( __FILE__ ) . '/vendor/wp-api/api-core/rest-api.php';
+	}
 }
 
 /**


danielbachhuber:
@michaelperrin @goblindegook Can you try this out to see if it meets your needs?

michaelperrin:
Thanks for the fix @danielbachhuber . That's the correct solution but it doesn't work for now as `wp-api/api-core` has not been published on Packagist, contrary to `wp-api/wp-api`.

I tested this creating a very simple `composer.json` file to test your branch:

    {
        "name": "hello/world",
        "require": {
            "wp-api/wp-api": "dev-1736-load-core"
        }
    }

And then running `composer install`, which gives this error:

> Your requirements could not be resolved to an installable set of packages.
> 
> Problem 1
>     - Installation request for wp-api/wp-api dev-1736-load-core -> satisfiable by wp-api/wp-api[dev-1736-load-core].
>     - wp-api/wp-api dev-1736-load-core requires wp-api/api-core 2.0-beta6 -> no matching package found.

rmccue:
Looks like we're incorrectly referencing a version for `api-core`, but [the project is up on Packagist](https://packagist.org/packages/wp-api/api-core)

danielbachhuber:
> Looks like we're maybe incorrectly referencing a version for api-core

I don't think so? I created a beta6 tag on api-core

rmccue:
Hmm, strange, maybe a caching issue of some sort.

danielbachhuber:
@michaelperrin Can you try clearing your local Composer cache? It's typically in `~/.composer/cache`

michaelperrin:
@danielbachhuber All right, I got it. It was not because of the Composer cache, but because of the minimum stability. So I updated my test `composer.json` like this for the tests:

    {
        "require": {
            "wp-api/wp-api": "dev-1736-load-core"
        },
        "minimum-stability": "dev"
    }

This way, the `wp-api/api-core` package gets installed. However, there is a problem as the resulting files are like this:

    |-vendor
    |---wp-api
    |-----api-core
    |------- The API Core files : the api-core folder should be a core subfolder of json-rest-api
    |-wp-content
    |---plugins
    |-----json-rest-api
    |------- The WP API files: everything is fine here.

I can make some tests with the `installer_paths` Composer options if needed.

danielbachhuber:
> This way, the wp-api/api-core package gets installed. However, there is a problem as the resulting files are like this:

I see. What configuration changes would you recommend for our composer.json?

danielbachhuber:
I think we're going to punt on Composer support directly through this repo for the time being.

Once WP 4.4 is released, we can drop the `core` submodule entirely, as we'll be using infrastructure directly in WordPress core.

PR #1746, opened 2015-11-18
Require admin functions for use of wp_handle_upload(), fixing fatal

joehoyle:


danielbachhuber:
Tests fail because of https://core.trac.wordpress.org/changeset/35671, which we'll need to sync into `api-core`

danielbachhuber:
Introduced in #1696

Issue #1321, opened 2015-06-02
Can register_api_field be better named?


kwight:
I understand  `register_api_field` is the best way to add any additional property to an endpoint, not just a meta field; for instance, I'm using it to [add post classes](https://github.com/kwight/wp-api-theming/blob/master/wp-api-theming.php#L26). Would a name like `register_api_property` be more appropriate, or is there a different way I should be adding non-meta-field properties?

joehoyle:
@kwight you are using it correctly, it's not just for meta field properties. Will raise this is a discussion point in our next meeting.

rachelbaker:
+1 to @joehoyle's comment above.  
The function is meant to add data to a given endpoint response object, so `property` may be a better fit based on the purpose of the function.  Naming things is hard.

rachelbaker:
@rmccue and I had a quick discussion about this to see if we could resolve for Beta 3.  We agreed that naming things is hard, and couldn't come to a quick consensus.  Pushing to Beta 4.

rmccue:
P.S. let's get `rest` into the name when we change it rather than `api`

jamesgol:
Was going to open up an issue on this because I noticed the naming discrepancy.  Is this going to be switched to register_rest_field()?

danielbachhuber:
Closing in favor of https://core.trac.wordpress.org/ticket/34730

danielbachhuber:
Actually, we're going to back it out of core, as it's conceptually tied to `WP_REST_Controller`

joehoyle:
decided on `regiister_rest_field`

danielbachhuber:
Let's rename to `register_rest_field()`, and include a backwards compat shim for `register_api_field()` for a few releases.

danielbachhuber:
I'm leaning back towards `register_rest_property()`... @WP-API/amigos thoughts?

danielbachhuber:
`register_rest_field()` it will be.

roikles:
Will `register_api_field` be used as an alias as it is still used in the documentation here: http://v2.wp-api.org/extending/modifying/

danielbachhuber:
> Will register_api_field be used as an alias?

Nope.

> It is still used in the documentation here

Sorry about that. I've updated https://github.com/WP-API/docs-v2/pull/78

PR #1748, opened 2015-11-18
Restore `register_api_field()` within the plugin

danielbachhuber:
It won't be included in WP 4.4 because it's conceptually tied to
`WP_REST_Controller`

See https://github.com/WP-API/WP-API/issues/1321#issuecomment-157835276


danielbachhuber:
Handled now in #1750

PR #1750, opened 2015-11-18
Sync with WordPress core as of r35691

danielbachhuber:


danielbachhuber:
Tests pass, except Travis seems stuck on something.

PR #1764, opened 2015-11-23
Don't permit requesting terms cross routes

### LINKED ISSUE #1763: Don't permit requesting terms cross routes
I have a category with `term_taxonomy_id=1`:

```
GET http://wordpress-develop.dev/wp-json/wp/v2/terms/category/1
```

I shouldn't be able to request it from the `tag` endpoint:

```
http://wordpress-develop.dev/wp-json/wp/v2/terms/tag/1
```

Reported in https://wordpress.org/support/topic/incorrect-route-wordpress-44-beta-4?replies=5

danielbachhuber:
Clients should only be able to request categories from the category route, and tags from the tag route

Fixes #1763

### DIFF
--- a/lib/endpoints/class-wp-rest-terms-controller.php
+++ b/lib/endpoints/class-wp-rest-terms-controller.php
@@ -138,7 +138,7 @@ class WP_REST_Terms_Controller extends WP_REST_Controller {
 	public function get_item( $request ) {
 
 		$term = get_term_by( 'term_taxonomy_id', (int) $request['id'], $this->taxonomy );
-		if ( ! $term ) {
+		if ( ! $term || $term->taxonomy !== $this->taxonomy ) {
 			return new WP_Error( 'rest_term_invalid', __( "Term doesn't exist." ), array( 'status' => 404 ) );
 		}
 		if ( is_wp_error( $term ) ) {
--- a/tests/test-rest-terms-controller.php
+++ b/tests/test-rest-terms-controller.php
@@ -323,6 +323,14 @@ class WP_Test_REST_Terms_Controller extends WP_Test_REST_Controller_Testcase {
 		$this->assertErrorResponse( 'rest_no_route', $response, 404 );
 	}
 
+	public function test_get_item_incorrect_taxonomy() {
+		register_taxonomy( 'robin', 'post' );
+		$term1 = $this->factory->term->create( array( 'name' => 'Cape', 'taxonomy' => 'robin' ) );
+		$request = new WP_REST_Request( 'GET', '/wp/v2/terms/category/' . $term1 );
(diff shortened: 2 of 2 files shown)


danielbachhuber:
@WP-API/amigos #reviewmerge

lib/endpoints/class-wp-rest-terms-controller.php
```diff
@@ -138,7 +138,7 @@ public function get_items( $request ) {
 	public function get_item( $request ) {
 
 		$term = get_term_by( 'term_taxonomy_id', (int) $request['id'], $this->taxonomy );
-		if ( ! $term ) {
+		if ( ! $term || $term->taxonomy !== $this->taxonomy ) {
```
joehoyle: How is this possible? My understanding is `get_term_by` will return null for a term_id that doesn't belong to that taxonomy, and no shared terms should now have stopped that being unexpected.

danielbachhuber: > How is this possible?

The test case fails as expected on nightly https://travis-ci.org/WP-API/WP-API/jobs/92832027



> My understanding is get_term_by will return null for a term_id that doesn't belong to that taxonomy, and no shared terms should now have stopped that being unexpected.

Nope:

```
} elseif ( 'term_taxonomy_id' == $field ) {
	$value = (int) $value;
	$_field = 'tt.term_taxonomy_id';

	// No `taxonomy` clause when searching by 'term_taxonomy_id'.
	$tax_clause = '';
} else {
	$term = get_term( (int) $value, $taxonomy, $output, $filter );
	if ( is_wp_error( $term ) || is_null( $term ) ) {
		$term = false;
	}
	return $term;
}

$term = $wpdb->get_row( $wpdb->prepare( "SELECT t.*, tt.* FROM $wpdb->terms AS t INNER JOIN $wpdb->term_taxonomy AS tt ON t.term_id = tt.term_id WHERE $_field = %s $tax_clause LIMIT 1", $value ) );
if ( ! $term )
	return false;

// In the case of 'term_taxonomy_id', override the provided `$taxonomy` with whatever we find in the db.
if ( 'term_taxonomy_id' === $field ) {
	$taxonomy = $term->taxonomy;
}
```

If `term_taxonomy_id` is provided, then `get_term_by()` ignores the provided taxonomy

rmccue: > How is this possible? My understanding is `get_term_by` will return null for a `term_id` that doesn't belong to that taxonomy, and no shared terms should now have stopped that being unexpected.

This changed in 4.4, see #1644 and [WP30620](https://core.trac.wordpress.org/ticket/30620).

PR #1770, opened 2015-11-26
Set the fields to 'id' fixing large memory usage

joehoyle:
The count query for the users should just pass the $fields => 'ID' param,
otherwise WordPress will init a WP_User object for every user on the site,
basically killing it.


danielbachhuber:
@joehoyle Did you mean to sneak the JS changes in?

PR #1772, opened 2015-11-27
Post updates should expect slashed data

joehoyle:
Right now, updating posts with quotes causes slashes to be saved to the db, as
wp_update_post does `wp_slash` on already slashed data if you pass it an
object.

I'm not sure if this is the best fix, but the unit test illustrates the issue.

### DIFF
--- a/tests/test-rest-posts-controller.php
+++ b/tests/test-rest-posts-controller.php
@@ -736,7 +736,8 @@ class WP_Test_REST_Posts_Controller extends WP_Test_REST_Post_Type_Controller_Te
 		) );
 		$request->set_body_params( $params );
 		$response = $this->server->dispatch( $request );
-		$this->check_create_post_response( $response );
+		$new_data = $response->get_data();
+		$this->assertEquals( "Rob O'Rourke's Diary", $new_data['title']['raw'] );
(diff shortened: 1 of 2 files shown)


tests/test-rest-posts-controller.php
```diff
@@ -712,6 +712,18 @@ public function test_create_post_with_invalid_date_gmt() {
...
+	public function test_create_post_with_quotes_in_title() {
+		wp_set_current_user( $this->editor_id );
+
+		$request = new WP_REST_Request( 'POST', '/wp/v2/posts' );
+		$params = $this->set_post_data( array(
+			'title' => "Rob O'Rourke's Diary",
+		) );
+		$request->set_body_params( $params );
+		$response = $this->server->dispatch( $request );
+		$this->check_create_post_response( $response );
```
danielbachhuber: Can we have an explicit test for the response title, to make this test more precise?

danielbachhuber: 301f33b

sanchothefat: I approve of this test. Also worth testing emoji and other unicode characters.

PR #1777, opened 2015-11-30
Add `?parent=<id>` support for attachments; update Post `_link`

### LINKED ISSUE #1767: ?post_parent=ID is ignored on /media endpoint
Issue: `site/wp-json/wp/v2/media?post_parent=ID` returns the same result as `site/wp-json/wp/v2/media`

There is no way to get attached images now, since _embedded['http://v2.wp-api.org/attachment'] was removed.

I am using beta7.

danielbachhuber:
Also stubs `get_collection_params()` abstraction for Posts controller to
make it possible for attachments controller to add `?parent=<id>` support

Fixes #1767

### DIFF
--- a/lib/endpoints/class-wp-rest-posts-controller.php
+++ b/lib/endpoints/class-wp-rest-posts-controller.php
@@ -15,20 +15,7 @@ class WP_REST_Posts_Controller extends WP_REST_Controller {
 
 		$base = $this->get_post_type_base( $this->post_type );
 
-		$posts_args = array(
-			'context'               => array(
-				'default'           => 'view',
-			),
-			'page'                  => array(
-				'default'           => 1,
-				'sanitize_callback' => 'absint',
-			),
-			'per_page'              => array(
-				'default'           => 10,
-				'sanitize_callback' => 'absint',
-			),
-			'filter'                => array(),
-		);
+		$posts_args = $this->get_collection_params();
 
 		register_rest_route( 'wp/v2', '/' . $base, array(
 			array(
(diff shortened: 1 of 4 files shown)


danielbachhuber:
@WP-API/amigos #reviewmerge

lib/endpoints/class-wp-rest-posts-controller.php
```diff
@@ -88,6 +75,7 @@ public function get_items( $request ) {
 		$args                   = array();
 		$args['paged']          = $request['page'];
 		$args['posts_per_page'] = $request['per_page'];
+		$args['post_parent']    = $request['parent'];
```
danielbachhuber: One thing that's a little dodgy is that _any_ controller extending this controller could use `parent` in the request, because we pass all arguments present in the request.

ioanlucut:
Hi! I am using version 2 Beta 10. Can you confirm that this functionality still works ? In my case, it really doesn't work. I am getting empty JSON back when I specify the parent request parameter. Thanks. (Wp v4.4.1)

danielbachhuber:
> Can you confirm that this functionality still works ?

Yes, it does.

> In my case, it really doesn't work. I am getting empty JSON back when I specify the parent request parameter.

Please open a new issue for support requests, and we'll help you debug the issue.

ioanlucut:
Hi, thanks for your answer! Actually I found out why it doesn't work. It's because a media file can be attached only to one post, therefore by simply adding a new media to a post without checking if it's already attached to another one, doesn't mean it is associated with that specific post.

I find this really awkward, why can't a media item belong to more than one posts ? Anyways, I know this is a Wordpress issue.

Thanks a lot for your answer.

danielbachhuber:
> I find this really awkward, why can't a media item belong to more than one posts ? Anyways, I know this is a Wordpress issue.

Right — we still need to sort that out https://github.com/WP-API/WP-API/issues/1987

PR #1780, opened 2015-12-01
Filter the responses by context after processing additional fields

joehoyle:
This means one can register an aditional field for specific contexts, as we'll 
interpret the `schema`


danielbachhuber:
It really feels like all of these transformational controller methods need to be abstracted to a new class.

danielbachhuber:
@joehoyle looks like something's bust with this

danielbachhuber:
Closing in favor of #2067

PR #1786, opened 2015-12-03
Add readme.txt, include deprecation notice

krogsgard:
This should be applied to version 1 of the API on WordPress.org. It includes a deprecation notice in the description.

### DIFF
--- a/readme.txt
+++ b/readme.txt
@@ -1,0 +1,1414 @@
+=== WP REST API (WP API) ===
+Contributors: rmccue, rachelbaker
+Tags: json, rest, api, rest-api
+Requires at least: 3.9
+Tested up to: 4.4
+Stable tag: 1.2.4
+License: GPLv2 or later
+License URI: http://www.gnu.org/licenses/gpl-2.0.html
+
+Access your site's data through an easy-to-use HTTP REST API (Version 1, deprecated).
+
+== Description ==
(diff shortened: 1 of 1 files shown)


readme.txt
```diff
@@ -0,0 +1,1414 @@
+=== WP REST API (WP API) ===
+Contributors: rmccue, rachelbaker
+Tags: json, rest, api, rest-api
+Requires at least: 3.9
+Tested up to: 4.4
+Stable tag: 1.2.4
+License: GPLv2 or later
+License URI: http://www.gnu.org/licenses/gpl-2.0.html
+
+Access your site's data through an easy-to-use HTTP REST API (Version 1, deprecated).
```
danielbachhuber: Couple of things:

1. We aren't tracking the `readme.txt` in the Github repo, so this shouldn't be merged to master.
2. We need to be careful about the language we're using, as (I don't believe) we recommend using v2 on production sites. If you need a production REST API, you should use v1.

krogsgard: @danielbachhuber I was under the impression that V2 is production ready now. Either way, there does need to be some note in the readme that there is a V2. Right now, when you Google "WordPress REST API" the old plugin result shows up first, and it'd be easy to install and use it without knowing you aren't on the right one.

Re: putting readme.txt in git, it'd be nice for it to be in git so that others can help contribute to it.

joehoyle:
Let's align on what we are saying we recommend in prod in todays meeting

danielbachhuber:
I've added a deprecation notice to the v1 plugin: http://plugins.trac.wordpress.org/changeset/1313888

PR #1793, opened 2015-12-07
Expose Post collection query params

danielbachhuber:
Including `author`, `order`, `orderby` and `status`

See #924


danielbachhuber:
@WP-API/amigos #reviewmerge

joehoyle:
:+1: looks good, tracking progress in #924 - @danielbachhuber does this trickle down to the `media` endpoint too?

danielbachhuber:
> does this trickle down to the media endpoint too?

Yes, because they extend the Posts controller

Issue #1706, opened 2015-10-31
Consistently show the `context` param as an endpoint argument for GET requests


rachelbaker:
In `WP_REST_Controller->get_collection_params()` we define the top-level collection params:
 * `page`
 * `per_page`
 * `search`

We should add to that the `context` param as well, so we have it consistently in the collection params.  

Additionally, I don't think we define the params for our single item endpoints.  We should communicate the context param there as well.

danielbachhuber:
Bumping to Beta 7. This bug isn't a regression, so it doesn't need to be addressed before we land Beta 6.

PR #1795, opened 2015-12-07
Standardize declaration of `context` param for collections across controllers

danielbachhuber:
See #1706

### DIFF
--- a/lib/endpoints/class-wp-rest-users-controller.php
+++ b/lib/endpoints/class-wp-rest-users-controller.php
@@ -766,13 +766,10 @@ class WP_REST_Users_Controller extends WP_REST_Controller {
 	 */
 	public function get_collection_params() {
 		$query_params = parent::get_collection_params();
-		$query_params['context'] = array(
-			'default'            => 'view',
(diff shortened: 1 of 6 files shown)


danielbachhuber:
@WP-API/amigos #reviewmerge

lib/endpoints/class-wp-rest-users-controller.php
```diff
@@ -766,13 +766,10 @@ public function get_item_schema() {
...
-		$query_params['context'] = array(
-			'default'            => 'view',
-			'description'        => 'Change the response format based on request context.',
-			'enum'               => array( 'embed', 'view', 'edit' ),
-			'sanitize_callback'  => 'sanitize_key',
-			'type'               => 'string',
-		);
+
+		$params['context']['default'] = 'view';
+		$params['context']['enum'] = array( 'embed', 'view', 'edit' );
```
joehoyle: Collection params are not json schema, so I don't think this `enum` will do anything, so far as I can tell

danielbachhuber: It doesn't right now, but I think it should. #1796

PR #1801, opened 2015-12-08
Ignore sticky posts by default

### LINKED ISSUE #1623: GET for posts returns one extra result up to per_page=15
I have found that posts endpoint is returning one extra result for all requests up to and including setting the per_page param to 15, change this to 16 or greater and it returns the number requested. If you add the param page=n where n is 2 or greater then it returns what you would expect.

For the sake of explanation I made a video showing this in action in the postman chrome extension but unfortunately I can't upload that to here. If you want it let me know

danielbachhuber:
Sticky posts are unrestful functionality. But, they can be re-enabled
for posts only with `filter[ignore_sticky_posts]=false`

Fixes #1623


danielbachhuber:
@WP-API/amigos #reviewmerge

PR #1803, opened 2015-12-08
In text strings, use `id` instead of `ID`

danielbachhuber:
`ID` is an implementation artifact. Our Resources use `id`.


danielbachhuber:
@WP-API/amigos #reviewmerge

kadamwhite:
+1

Issue #1453, opened 2015-08-03
Adding comment as author/member not working


DmitryOverchuk:
When i trying to add new comment to post as admin, everything works pefrect, as response i receiving added comment.

But when i'm doing the same as author or member, i receive no response and new comment doesn't appear.

rachelbaker:
@DmitryOverchuk Depending on your comment moderation settings, the comment can be created but not approved.  Can you check to see if the comment is created, but being held for approval?

DmitryOverchuk:
Occasionally comment can be added (even if i add it as author(user role)), but not always.

When comment successfully created, it already approved and visible with other post comments

![2015-08-03 21 17 43](https://cloud.githubusercontent.com/assets/5804318/9044443/568d5544-3a25-11e5-9b64-cb533dfb6856.png)

This user has author role.

Here is my discussion settings:

![2015-08-03 21 26 10](https://cloud.githubusercontent.com/assets/5804318/9044552/173f6638-3a26-11e5-83d2-ce8d0eae0323.png)

DmitryOverchuk:
Also if comment is not created i receive empty response body from server

sogko:
Seeing similar issue, unable to create comments, even for admin author.

Getting an empty response body from server and no comment is created (check using wp-admin and authenticated GET /comments)

What are the minimum required POST data for POST /comments?

danielbachhuber:
> What are the minimum required POST data for POST /comments?

For authenticated users, we only require the `post` argument when creating a comment.

PR #1804, opened 2015-12-08
Explicitly require `post` argument when creating a comment

danielbachhuber:
Because the internals require `post` when creating a comment, explicitly
marking it as required will provide a more helpful error message when
the argument isn't provided in the request.

Discovered in #1453


rachelbaker:
@danielbachhuber We are actually unwinding Comments and their dependency on Posts in core.  You can create Comments that aren't attached to Posts now (comment_post_ID is not required in wp_insert_comment()).

danielbachhuber:
@rachelbaker Ah, makes sense. I've created a separate issue to eventually fix that.

PR #1806, opened 2015-12-08
Include `full` image size in attachment `sizes` attribute

### LINKED ISSUE #365: Include information about the original image in `attachment_meta` for media
Currently in `attachment_meta.sizes`, you can find information about all the thumbnails generated, but information about the original uploaded image is scattered on the `source` field and on `attachment_meta`, with no mime-type information. It would be nice if it was consistent and I could get information about all the sizes at once.

danielbachhuber:
Fixes #365


danielbachhuber:
@WP-API/amigos #reviewmerge

PR #1824, opened 2015-12-09
Rename `register_api_field()` to `register_rest_field()`

### LINKED ISSUE #1321: Can register_api_field be better named?

danielbachhuber:
Introduces a `register_api_field()` function for backwards compat, which
calls `_doing_it_wrong()`

Fixes #1321

### DIFF
--- a/plugin.php
+++ b/plugin.php
@@ -231,7 +231,7 @@ if ( ! function_exists( 'rest_authorization_required_code' ) ) {
 	}
 }
 
-if ( ! function_exists( 'register_api_field' ) ) {
+if ( ! function_exists( 'register_rest_field' ) ) {
 	/**
 	 * Registers a new field on an existing WordPress object type.
 	 *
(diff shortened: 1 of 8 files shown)


danielbachhuber:
@WP-API/amigos #reviewmerge

joehoyle:
@danielbachhuber are you thinking the backcompat function will also go into core when the time comes?

danielbachhuber:
> are you thinking the backcompat function will also go into core when the time comes?

Nope. We can keep it in the plugin indefinitely. If the user has the plugin activated, then they'll be able to use the function. When they deactivate the plugin, they should update function calls.

PR #1826, opened 2015-12-10
Respect core's `comment_registration` setting

danielbachhuber:
If it's enabled, require users to be logged in to comment.


danielbachhuber:
@WP-API/amigos #reviewmerge

joehoyle:
Restarted build, some one-off SSL handshake failure with the SVN repo

PR #1827, opened 2015-12-10
Default to wildcard when searching users

### LINKED ISSUE #1771: Searching users requires wildcards
Unlike posts, which does wildcard search by default, `/wp/v2/users` requires an exact match on any of the searched fields, because this is the internal behaviour of `WP_User_Query`. One cal work around this by doing `/wp/v2/users?search=*ada*` to match "adam" etc.

I propose we auto add the wildcards on to the search query in users to match that of `/wp/v2/posts`

danielbachhuber:
It's more consistent with Posts

Fixes #1771


danielbachhuber:
@WP-API/amigos #reviewmerge

PR #1802, opened 2015-12-08
Move tags and categories to top-level endpoints

danielbachhuber:
Tags are now accessible at `/wp/v2/tags`, and categories accessible at
`/wp/v2/categories`. Post terms reside at `/wp/v2/posts/<id>/tags` and
`/wp/v2/<id>/categories`

Although this introduces the risk of collision between custom post types
and custom taxonomies, said collision isn't currently handled in
WordPress core. Developers can use `rest_base` to ensure CPTs and
taxomonies have distinct slugs.

Fixes #829

PR #1828, opened 2015-12-10
Bring the wp-api.js library up to date for v2 of the REST API

adamsilverstein:
* Initial pass at bringing the library up to date
* Correct model endpoints that have changed
* Verified all models read correctly
* Verified Post create, read, destroy (trash)
* Clean up doc block language for models
* Add additional JSDOC3 Doc Blocks for models
* rename internal BaseModel to WPApiBaseModel
* user model defaults to 'me' route if no user id is passed
* remove slug default from Post model, passing a blank slug was causing Create to fail.
* Consider eliminating defauls, the api already has defaults?
* Correct overriding of 'delete', backbone model funciton is `destroy`

### DIFF
--- a/wp-api.js
+++ b/wp-api.js
@@ -1,13 +1,25 @@
-( function( WP_API_Settings, Backbone, _, window, undefined ) {
+(function( window, undefined ) {
+
+	'use strict';
+
+	function WP_API() {
+		this.models = {};
+		this.collections = {};
+		this.views = {};
+	}
+
+	window.wp = window.wp || {};
+	wp.api = wp.api || new WP_API();
+
+})( window );
+
+(function( window, undefined ) {
+
 	'use strict';
 
 	window.wp = window.wp || {};
-
-	wp.api = {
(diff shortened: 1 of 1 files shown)


danielbachhuber:
:ship: :ship:

wp-api.js
```diff
@@ -270,24 +298,17 @@
...
-			 *
-			 * @returns {string}
-			 */
-			url: function() {
-				var id = this.get( 'id' );
-				id = id || '';
-
-				return WP_API_Settings.root + 'wp/v2/taxonomies/' + this.get( 'taxonomy' ) + '/terms/' + id;
-			},
+			urlRoot: WP_API_Settings.root + 'wp/v2/terms/tag',
```
joehoyle: This will now need updated per https://github.com/WP-API/WP-API/pull/1802

adamsilverstein:
@joehoyle the js client now traverses the schema to determine models to build, so no update will be required :)

PR #1829, opened 2015-12-10
Link Types to their Collections and vice versa

### LINKED ISSUE #1818: Link Posts to their Type and vice versa
Same as #1817 and #1814

Posts should link to their Type with `about` relation. Types should link to their Post collection with `collection` relation.

danielbachhuber:
Fixes #1818

### DIFF
--- a/lib/endpoints/class-wp-rest-post-types-controller.php
+++ b/lib/endpoints/class-wp-rest-post-types-controller.php
@@ -42,7 +42,7 @@ class WP_REST_Post_Types_Controller extends WP_REST_Controller {
 			if ( is_wp_error( $post_type ) ) {
 				continue;
 			}
-			$data[ $obj->name ] = $post_type;
+			$data[ $obj->name ] = $this->prepare_response_for_collection( $post_type );
 		}
 		return $data;
 	}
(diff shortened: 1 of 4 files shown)


danielbachhuber:
@WP-API/amigos #reviewmerge

rachelbaker:
@danielbachhuber would it also be worthwhile to add `rest_base` in the post_type response data?

joehoyle:
I'm a pretty big -1 on adding `rest_base` anywhere - it doesn't have much meaning in the context of just the API, and encourages manual building of arbitrary URLs which I think we'd be best to stick to just stick to `_links`.

joehoyle:
It's not clear to me that `collection` is the correct relation, from http://tools.ietf.org/html/rfc6573

>  When included in a resource that represents a member of a collection,
   the 'collection' link relation identifies a target resource that
   represents a collection of which the context resource is a member.

> For example, if a resource represents a single product in a catalog,
   that same representation may include a link to a resource that
   represents a product group to which this single product belongs

If anything `collection` would be a relation on a single post, pointing back to the collection - or the link of a single post could also point to the post type object via the `collection` relation also (akin to a product pointing to a "product group" object. 

The `collection` relation on a post type object (as this PR makes it) should be pointing to a collection / collection object of `post type objects`. it's saying "this resource is a member of the collection x", not "the items in this collection are here".

If anything we'd want the `item` relation 

> When included in a resource that represents a collection, the 'item'
   link relation identifies a target resource that represents a member
   of that collection.

> For example, if a resource represents a catalog of products, that
   same representation may include one or more links to resources that
   represent members of that catalog.

However, what we really wan't is `items` but these no spec for that currently.

PR #1831, opened 2015-12-10
Make `prepare_*` filters more consistent

danielbachhuber:
* Pass `$response` variable, to more clearly communicate expected value
* PHPdoc clarifications and corrections


danielbachhuber:
@WP-API/amigos #reviewmerge

adamsilverstein:
@danielbachhuber - looks good, thanks for cleaning up and completing.

PR #1838, opened 2015-12-11
Introduce `get_context_param()` for getting the magical context param

danielbachhuber:
Abstracting this to a common method lets us reuse it between `GET`
collection and `GET` single instance

See #1706


danielbachhuber:
@WP-API/amigos #reviewmerge

danielbachhuber:
Adding some tests.

danielbachhuber:
@WP-API/amigos #reviewmerge

danielbachhuber:
Feeling much better about this with tests.

Issue #1834, opened 2015-12-10
Update the JS Lib URLs for Tags and Categories


joehoyle:
In #1802 we moved taxonomies to the root, and renamed `tags` and `categories` so we need to update the JS library to reflect this.

danielbachhuber:
This is no longer an issue in the JS client, and the latest version will be synced over shortly by @adamsilverstein

PR #1841, opened 2015-12-11
Update the terms collections URLs in wp-api.js

joehoyle:
See #1834

A user can no longer just pass `taxonomy` to the collection, they'll need to
know the `rest_base` of the taxonomy, unfortunatly.

### DIFF
--- a/wp-api.js
+++ b/wp-api.js
@@ -308,7 +308,7 @@
 		{
 			idAttribute: 'id',
 
-			urlRoot: WP_API_Settings.root + 'wp/v2/terms/tag',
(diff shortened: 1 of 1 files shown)


joehoyle:
@adamsilverstein feedback welcome here if you think there might be a better way to handle this.

rmccue:
@joehoyle You probably want to patch the source on [the original repo](https://github.com/WP-API/client-js/tree/master/js).

joehoyle:
My bad! Would still like @adamsilverstein's input

PR #1846, opened 2015-12-11
Fix link relations for individual types and taxonomies

danielbachhuber:
`collection` should go to the collection of types and taxonomies,
respectively. While imperfect, `item` is the best definition we have for
the link to the collection of items.

From https://github.com/WP-API/WP-API/pull/1829#issuecomment-163777701


joehoyle:
I've seen a few PRs merged without review, was there communication somewhere else for peer review for these?

I don't think `item` is a correct relation - we should use a custom `wp` one if there is not the right relation already defined, we do this plenty elsewhere.

rmccue:
Agreed, we should DIY the relation here.

Issue #1794, opened 2015-12-07
Remove `core` submodule


danielbachhuber:
Once WordPress 4.4 is released, we don't need to submodule `core` any more as the infrastructure will be included in WordPress core.

In doing so, we'll want to make sure we bump the minimum supported version to WordPress 4.4.

PR #1847, opened 2015-12-11
Remove infrastructure tests

danielbachhuber:
Now that the infrastructure is committed to core, they're redundant

Related #1794


danielbachhuber:
@WP-API/amigos #reviewmerge

PR #1852, opened 2015-12-13
Prevent error notice when `show_in_rest` isn't set for a post type

### LINKED ISSUE #1850: Notice: Undefined property: stdClass::$show_in_rest
Occurs with WP 4.4, plugin beta9 when querying by single post:
`/lib/endpoints/class-wp-rest-posts-controller.php on line 911`

No other plugins active. Twenty Fifteen is theme.

danielbachhuber:
Fixes #1850


danielbachhuber:
@WP-API/amigos #reviewmerge

PR #1853, opened 2015-12-13
Fix link relations for individual types and taxonomies, part 2

danielbachhuber:
Use custom `https://api.w.org/items` relation instead of incorrect
`item` relation.

See https://github.com/WP-API/WP-API/pull/1846#issuecomment-164086269


danielbachhuber:
@WP-API/amigos #reviewmerge

joehoyle:
:+1:

PR #1858, opened 2015-12-13
Remove unneeded calls to `assertNotInstanceOf` and `rest_ensure_response`

danielbachhuber:
`dispatch()` has only returned `WP_REST_Response` for quite a while.


danielbachhuber:
@WP-API/amigos #reviewmerge

joehoyle:
makes sense

PR #1864, opened 2015-12-15
Add `edit` context for `wp/v2/taxonomies`; require `manage_terms` cap

### LINKED ISSUE #1185: '/taxonomies' response for edit/ embed context returns inconsistent response for unprivileged access
https://github.com/WP-API/WP-API/issues/702#issuecomment-66400990
> Additionally, right now setting context either succeeds (returns all data for that context) or fails (returns no data/an error), which is kind of nice in that you immediately know if you can't access the data. Returning a subset of data if you don't have full permissions means you can't necessarily be sure if there's more data to access.

It should either return no data or error for consistency.

danielbachhuber:
Fixes #1185

### DIFF
--- a/lib/endpoints/class-wp-rest-taxonomies-controller.php
+++ b/lib/endpoints/class-wp-rest-taxonomies-controller.php
@@ -43,10 +43,10 @@ class WP_REST_Taxonomies_Controller extends WP_REST_Controller {
 		}
 		$data = array();
 		foreach ( $taxonomies as $tax_type => $value ) {
-			$tax = $this->prepare_item_for_response( $value, $request );
-			if ( is_wp_error( $tax ) ) {
+			if ( empty( $value->show_in_rest ) || ( 'edit' === $request['context'] && ! current_user_can( $value->cap->manage_terms ) ) ) {
 				continue;
 			}
+			$tax = $this->prepare_item_for_response( $value, $request );
 			$tax = $this->prepare_response_for_collection( $tax );
 			$data[ $tax_type ] = $tax;
 		}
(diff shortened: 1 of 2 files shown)


danielbachhuber:
@WP-API/amigos #reviewmerge

danielbachhuber:
@joehoyle and I discussed some:

> joehoyle [7:08 AM] hmm
> ​[7:08] why is `labels` not part of the view context?
> ​[7:09] for example, I'd want the `search_items` label it build view-only UI
> danielbachhuber [7:10 AM] core doesn't expose `search_items` on the frontend
> ​[7:10] historically, our distinction between `view` and `edit` is backend and frontend
> joehoyle [7:10 AM] I thought it was "I want to get this data to perform an edit on it"
> ​[7:10] that's how I've always thought about it anyway
> ​[7:11] I'd expect a read-only admin user to be able to ​_see_​ everything using context=view
> danielbachhuber [7:12 AM] well, I suppose this is yet another problem with the context abstraction
> ​[7:12] meta only has `edit` context
> joehoyle [7:12 AM] interesting
> ​[7:12] I don't know what that is. to me, context=edit implies authorization(edited)
> danielbachhuber [7:14 AM] that is true, and in this pull request `context=edit` requires authorization
> joehoyle [7:19 AM] so, for taxonomy labels, I don't think there's a private data issue, i think it should be in `view`
> ​[7:19] infact, context=edit doesn't make any sense for taxonomies, as you can't edit them
> ​[7:19] IMO
> danielbachhuber [7:21 AM] ok. pretty sure that's different than how we've been operating
> joehoyle [7:22 AM] hmm yeah sounds like we need to get on the same page with that `context` means - do we have docs written for it?
> danielbachhuber [7:22 AM] nope

danielbachhuber:
@rachelbaker Where was the conversation on this?

PR #1865, opened 2015-12-15
Support for querying comments where `post=0`

### LINKED ISSUE #1832: Allow comments with a `comment_post_ID` of `0` in our GET requests
Users with the `moderate_comments` permission should be allowed to view comments that are not attached to a Post.  

#### GET Comments Collection:

We don't check that the `comment_post_ID !==0` in `WP_REST_Comments_Controller->get_items()` before checking the read post permission anyone that even has comments without a post relationship gets this response:
```
[ ]
```

No error, just an empty response with a PHP Notice in their log about trying to get the property on a non-object.

#### GET Single Comment

```
$post = get_post( $comment->comment_post_ID );
		if ( empty( $post ) ) {
			return new WP_Error( 'rest_post_invalid_id', __( 'Invalid post id.' ), array( 'status' => 404 ) );
		}
```

danielbachhuber:
Fixes #1832


joehoyle:
@danielbachhuber want to fix this up?

danielbachhuber:
@joehoyle it's blocked on core right now, because of https://core.trac.wordpress.org/ticket/35090

danielbachhuber:
This is going to need https://core.trac.wordpress.org/changeset/36381 to end up in a nightly to start passing. We can just skip the test on latest (or add a filter shim)

danielbachhuber:
@WP-API/amigos #reviewmerge

PR #1866, opened 2015-12-15
Fix invalid capability for comments get items permissions callback

danielbachhuber:
`manage_comments` doesn't exist; `moderate_comments` does


danielbachhuber:
@WP-API/amigos #reviewmerge

PR #1867, opened 2015-12-15
Use proper status code when failing to get comments of private post

danielbachhuber:


danielbachhuber:
@WP-API/amigos #reviewmerge

lib/endpoints/class-wp-rest-comments-controller.php
```diff
@@ -354,7 +354,7 @@ public function get_items_permissions_check( $request ) {
 			$post = get_post( (int) $request['post'] );
 
 			if ( $post && ! $this->check_read_post_permission( $post ) ) {
-				return new WP_Error( 'rest_cannot_read_post', __( 'Sorry, you cannot read the post for this comment.' ) );
+				return new WP_Error( 'rest_cannot_read_post', __( 'Sorry, you cannot read the post for this comment.' ), array( 'status' => rest_authorization_required_code() ) );
```
rachelbaker: @danielbachhuber Would a `403` status code be more accurate here?  Even if a user was logged-in they may not have permission to read private posts.

danielbachhuber: `rest_authorization_required_code()` returns `401` when the user is logged out, and `403` when the user is logged in?

joehoyle: `401` is basically "if you are not logged in, login and retry, then we'll tell you if you can access this" so I think we are good.

PR #1878, opened 2015-12-20
Adjust Posts Pagination Headers for `filter` Params

### LINKED ISSUE #1876: "filter[posts_per_page]" not affecting the "X-WP-TotalPages" response header
Hi there,

I'm using `filter[posts_per_page]` to retrieve a collection of posts for my application. The problem I'm facing is that the `filter[posts_per_page]` parameter is not updating the `X-WP-TotalPages` header as returned in the response by the server.

However, if I use `per_page` parameter in the request, it's updating the `X-WP-TotalPages` accordingly.

Consider the following request to the server:

    $ GET /wp/v2/posts?filter[posts_per_page]=5

The headers returned are as follows:

    X-WP-Total → 45
    X-WP-TotalPages → 5

Now if I use the `per_page` parameter like the following:

    $ GET /wp/v2/posts?per_page=5

The headers returned are like the following:

    X-WP-Total → 45
    X-WP-TotalPages → 9

It can be seen that the later one updated the `X-WP-TotalPages` header accordingly.

I think that the `X-WP-TotalPages` should be updated whenever a request is made via `filter[posts_per_page]` parameter.

rachelbaker:
Use the `posts_per_page` and `paged` values from `$query_args` instead of only from the shortened Request params when calculating the `X-WP-TotalPages` header as well as the `prev` and `next` link headers.  This prevents incorrect headers when either the `filter[posts_per_page]` or `filter[paged]` request params are used, as well as hooks into the `rest_post_query` filter.

Fixes #1876

### DIFF
--- a/lib/endpoints/class-wp-rest-posts-controller.php
+++ b/lib/endpoints/class-wp-rest-posts-controller.php
@@ -111,24 +111,28 @@ class WP_REST_Posts_Controller extends WP_REST_Controller {
 
 		$response = rest_ensure_response( $posts );
 		$count_query = new WP_Query();
+
+		// Store paged value for pagination headers then unset for count query.
+		$page = (int) $query_args['paged'];
 		unset( $query_args['paged'] );
+
 		$query_result = $count_query->query( $query_args );
 		$total_posts = $count_query->found_posts;
 		$response->header( 'X-WP-Total', (int) $total_posts );
-		$max_pages = ceil( $total_posts / $request['per_page'] );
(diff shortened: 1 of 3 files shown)


rachelbaker:
@WP-API/amigos #reviewmerge

rachelbaker:
Even though it looks like only the Posts endpoints use the `filter` param, the `Users` endpoint does include a filter (`rest_user_query`) so 5640d87 fixes the pagination headers there as well.

PR #1879, opened 2015-12-20
Change order of Users response to match schema order

rachelbaker:
Related to #1708, this adjusts the order of the User resources so that `id` is the first property and the remaining properties match the Schema order.


rachelbaker:
@WP-API/amigos #reviewmerge

danielbachhuber:
It would be neat if we were able to just reorder based on the schema.

PR #1880, opened 2015-12-21
Allow Comments to be created with a passed `author_ip`

### LINKED ISSUE #1788: Comments: cannot send true IP address
I see that this line: 
`$prepared_comment['comment_author_IP'] = '127.0.0.1';`
from: https://github.com/WP-API/WP-API/blob/074dab5071d0a3c685a73c6b843af00b390ad771/lib/endpoints/class-wp-rest-comments-controller.php#L182

causes the IP address to be unable to be set to whomever is calling the API. This seems like a problem. If the end-user is calling the API, then their IP address should be sent. If a server is proxying the comment, then its IP address should be sent -- and possibly be able to be overridden via WP's pre_comment_user_ip hook (https://codex.wordpress.org/Plugin_API/Filter_Reference/pre_comment_user_ip).

For example, I would like to use pre_comment_user_ip to replace the REMOTE_ADDR with X_FORWARDED_FOR, since I am proxying the comment API post.

I was able to get around the API forcing localhost to be the IP by this hacked up WP filter:

```
function proxy_ip( $prepared_comment )
{
    // Use IP set in the REMOTE_ADDR server variable by default
    $CLIENT_IP = $_SERVER['REMOTE_ADDR'];
    if (!empty($_SERVER['X_FORWARDED_FOR'])) {
        $X_FORWARDED_FOR = explode(',', $_SERVER['X_FORWARDED_FOR']);
    }
    elseif (!empty($_SERVER['HTTP_X_FORWARDED_FOR'])) {
        $X_FORWARDED_FOR = explode(',', $_SERVER['HTTP_X_FORWARDED_FOR']);
    }

    if (!empty($X_FORWARDED_FOR)) {
        $CLIENT_IP = trim($X_FORWARDED_FOR[0]);
        //$CLIENT_IP = preg_replace('/[^0-9a-f:\., ]/si', '', $CLIENT_IP);
    }

    $prepared_comment['comment_author_IP'] = $CLIENT_IP;
    return $prepared_comment;
}
add_filter( 'rest_pre_insert_comment', 'proxy_ip' );
```

rachelbaker:
This PR allows the `comment_author_IP` value to be set when creating a comment.  While working on this, I remembered why we don't handle this field...  while with `wp_insert_comment()` you can as of WP 4.3 set the `comment_author_IP` **BUT** you cannot modify this field in `wp_update_comment()`.  
### Questions:
1. Do we allow this property to be set when creating a Comment, BUT not allow it to be edited?  How should that be reflected in the schema?
2. Should we add a filter for the default `comment_author_IP` value (currently `127.0.0.1`)?
## UPDATE:

This PR has been refreshed and as of WP 4.5 the `comment_author_ip` property can be modified.  View details on the refresh: https://github.com/WP-API/WP-API/pull/1880#issuecomment-217732303

Fixes #1788


rachelbaker:
@WP-API/amigos thoughts?

lib/endpoints/class-wp-rest-comments-controller.php
```diff
@@ -699,6 +707,14 @@ protected function prepare_item_for_database( $request ) {
 			$prepared_comment['comment_author_url'] = $request['author_url'];
 		}
 
+		if ( isset( $request['author_ip'] ) ) {
+			if ( filter_var( $request['author_ip'], FILTER_VALIDATE_IP, FILTER_FLAG_IPV4 ) ) {
```
danielbachhuber: What if we added `ip_address` as a new format in our schema, so validation could happen much earlier (and in batch)?

rachelbaker:
Related core ticket: https://core.trac.wordpress.org/ticket/35276

danielbachhuber:
Something I've been thinking about recently with regards to whether or not fields should be editable: import / export. If someone were to build a import / export workflow based on the REST API, which seems totally reasonable to me, it seems like a reasonable assumption they should be able to update most any fields for resources.

rachelbaker:
> Something I've been thinking about recently with regards to whether or not fields should be editable: import / export. If someone were to build a import / export workflow based on the REST API, which seems totally reasonable to me, it seems like a reasonable assumption they should be able to update most any fields for resources.

@danielbachhuber, can you please add those thoughts to https://core.trac.wordpress.org/ticket/35276?  Additional supporting opinions will help.

rachelbaker:
In WP 4.5 this will no longer be blocked on core.  I will pick this back up after release.

caglaroner:
@rachelbaker & @danielbachhuber  Great job on the WP-API and thank you very much for all your contributions. 

Our team is currently blocked on this issue (more specifically the post comment part) and we would hate to branch out and patch it on our end. 

Is there any timeline or progress on this issue?

rachelbaker:
@WP-API/amigos This PR has been refreshed and is ready for review.
- Handles setting and updating the `author_ip` property on comment create & update
- Uses 127.0.0.1 as the default `author_ip` value in the schema
- Adds a new `ipv4` string format (which matches the following example on the JSON Schema site: http://json-schema.org/example2.html) 
- Adds the IP validation utility function `rest_is_ip_address()` which ideally I would like to get into WP core.

PR #1881, opened 2015-12-21
Fix fatal error when `prepare_item_for_database` returns a WP_Error

rachelbaker:
Discovered while working on #1880:
`WP_REST_Comments_Controller->prepare_item_for_database()` can return a WP_Error, but we currently assume it won't.  This can result in a Fatal error when trying to use the WP_Error object as an array.


rachelbaker:
@WP-API/amigos #reviewmerge

danielbachhuber:
Can we add a test for the case in which it returns a fatal error?

rachelbaker:
Replaced by #1917

PR #1882, opened 2015-12-21
Allow Terms Collection Params to be Filtered

rachelbaker:
Adds a filter for the `prepared_args` before passing them to `get_terms()` (matching behavior to the Posts and Users collections), and then adjusting the pagination headers so that the `prepared_args` are used instead of only the Request params.  

See #1878

### DIFF
--- a/lib/endpoints/class-wp-rest-terms-controller.php
+++ b/lib/endpoints/class-wp-rest-terms-controller.php
@@ -113,8 +113,13 @@ class WP_REST_Terms_Controller extends WP_REST_Controller {
 		}
 
 		$response = rest_ensure_response( $response );
+
+		// Store pagation values for headers then unset for count query.
+		$per_page = (int) $prepared_args['number'];
+		$page = ceil( ( ( (int) $prepared_args['offset'] ) / $per_page ) + 1 );
 		unset( $prepared_args['number'] );
 		unset( $prepared_args['offset'] );
+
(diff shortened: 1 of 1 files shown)


lib/endpoints/class-wp-rest-terms-controller.php
```diff
@@ -91,6 +91,20 @@ public function get_items( $request ) {
...
+		 * Enables adding extra arguments or setting defaults for a terms
+		 * collection request.
+		 *
+		 * @see https://developer.wordpress.org/reference/functions/get_terms/
+		 *
+		 * @param array           $prepared_args Array of arguments to be
+		 *                                       passed to get_terms.
+		 * @param WP_REST_Request $request       The current request.
+		 */
+		$prepared_args = apply_filters( 'rest_terms_query', $prepared_args, $request );
```
rachelbaker: This filter naming matches other similar filters, but `pre_rest_terms_query` would be a better match for Core's naming convention.  @rmccue @joehoyle thoughts?  Should we change these now for Posts, Users, etc?

danielbachhuber: I don't have a strong preference either way, but renaming would make sense to me.

danielbachhuber: Created a separate issue for this #1937

PR #1886, opened 2015-12-21
Defer to `edit_posts` of the custom post type

danielbachhuber:
These private query variables should be accessible to CPTs when a user
can't create normal posts.

### DIFF
--- a/tests/test-rest-pages-controller.php
+++ b/tests/test-rest-pages-controller.php
@@ -47,6 +47,23 @@ class WP_Test_REST_Pages_Controller extends WP_Test_REST_Post_Type_Controller_Te
 
 	}
 
+	public function test_get_items_private_filter_query_var() {
+		// Private query vars inaccessible to unauthorized users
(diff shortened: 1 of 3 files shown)


danielbachhuber:
Need to add some tests around private query var handling.

rmccue:
:+1: on this, good spot.

danielbachhuber:
@WP-API/amigos #reviewmerge

tests/test-rest-pages-controller.php
```diff
@@ -47,6 +47,25 @@ public function test_get_items() {
 
 	}
 
+	public function test_get_items_private_filter_query_var() {
+		// Private query vars inaccessible to unauthorized users
+		wp_set_current_user( 0 );
```
danielbachhuber: I wanted to test contributors here too, as they don't have the ability to edit pages, but I ran into #1892

joehoyle:
:+1:

PR #1889, opened 2015-12-22
check that the post type has a show_in_rest property 

foxpaul:
When hitting an endpoint (not /revisions) for a post that is in the 'revision' state, we get an 
undefined property when calling check_is_post_type_allowed.

Notice: Undefined property: stdClass::$show_in_rest in rest-api/lib/endpoints/class-wp-rest-posts-controller.php on line 911


danielbachhuber:
Thanks for the PR, @foxpaul ! This was actually fixed in #1852, and will be included in beta 10

PR #1894, opened 2015-12-23
Add `edit` context for `wp/v2/types`; require `edit_posts` cap

### LINKED ISSUE #1183: '/types' with edit context returns list of types with unprivileged request.
It should return no data/ error when edit context is used by unprivileged user.

danielbachhuber:
Fixes #1183


danielbachhuber:
@WP-API/amigos #reviewmerge

PR #1905, opened 2015-12-28
Add filter for get_endpoint_args_for_item_schema

wpsmith:
I would like to see a filter for `get_endpoint_args_for_item_schema` so this core code for the item schema can be extended instead of re-written and extended.


danielbachhuber:
Thanks for the PR — not quite ready to add a filter here though, see #1911

Issue #1609, opened 2015-09-28
Limit number of requested items for collection queries


Toddses:
No way to query for all terms. With `get_terms()` from WordPress you can pass the `number` arg as 0, or it defaults to 0, to get all the terms with the associated taxonomy. The terms controller uses `per_page` to fill out the `number` arg. The default for `per_page` is 10, and you can't pass 0 to this argument (or any negative numbers, because it sanitizes the value with `absint`). This effectively makes it impossible to query for all terms in a taxonomy.

donaldallen:
Can't figure this one out either.

donaldallen:
How are you using filtering on this request?

I've tried:

```
wp-json/wp/v2/terms/community_area?filter[per_page]=100
```

And I'm still only getting back 10 results. Is per_page incorrect usage? I tried this as well:

```
wp-json/wp/v2/terms/community_area?filter[number]=100
```

Toddses:
'....?per_page=100'

Asssuming you're using v2.0. Don't need the filter on this one.

donaldallen:
So even if I do that, I only get back 10 results. If I go into ```class-wp-rest-terms-controller.php``` and change:

```
$prepared_args['number']  = $request['per_page'];
```

to:

```
$prepared_args['number']  = 100;
```

That works. When I var_dump the $request, I'm not seeing per_page or anything else I pass through.

Toddses:
Strange - I don't seem to be having this issue. I can pass per_page and get the expected results. What version of the API are you running?

donaldallen:
Args for my custom post type:

```
    $args = [
        'label'               => __('community', 'something'),
        'description'         => __('Community', 'something'),
        'labels'              => $labels,
        'supports'            => ['title', 'thumbnail', 'editor', 'excerpt'],
        'hierarchical'        => false,
        'public'              => true,
        'show_in_rest'        => true,
        'show_ui'             => true,
        'show_in_menu'        => true,
        'show_in_nav_menus'   => false,
        'show_in_admin_bar'   => false,
        'menu_position'       => 31,
        'menu_icon'           => 'dashicons-admin-multisite',
        'can_export'          => true,
        'has_archive'         => true,
        'exclude_from_search' => false,
        'publicly_queryable'  => true,
        'capability_type'     => 'page',
    ];
```

And the args for my taxonomy:

```
    $args = [
        'labels'                     => $labels,
        'hierarchical'               => true,
        'public'                     => true,
[...]

PR #1906, opened 2015-12-28
Sanitize per_page with intval to allow for -1

wpsmith:
Sanitize `per_page` with `intval` to allow for `-1` value. See [WP_Query:Pagination Parameters](https://codex.wordpress.org/Class_Reference/WP_Query#Pagination_Parameters). Obviously, this is less than ideal for any REST request but since it is allowable in core on page render, why not allow it for REST API?


danielbachhuber:
We don't permit unbounded queries through the REST API, and will likely cap the number of items in the future. See https://github.com/WP-API/WP-API/issues/1609#issuecomment-146568663

PR #1907, opened 2015-12-28
Cast integer with (int) over intval for speed

wpsmith:
While probably negligible, casting the integer `(int)` is faster than using `intval()`. Just an idea.


danielbachhuber:
> casting the integer (int) is faster than using intval()

Can you point to data that supports this?

wpsmithtwc:
Here you go: https://hakre.wordpress.com/2010/05/13/php-casting-vs-intval/

danielbachhuber:
Thanks!

PR #1908, opened 2015-12-28
Fix PHP docs. get_item is passed an array as opposed to a WP_REST_Request in several places.

wpsmith:
`get_item` is passed an array as opposed to a `WP_REST_Request` in several places (NOTE, line numbers may change but are correct at the time of submission):
* class-wp-rest-attachments-controller.php, line [85](https://github.com/WP-API/WP-API/blob/develop/lib/endpoints/class-wp-rest-attachments-controller.php#L85), [125](https://github.com/WP-API/WP-API/blob/develop/lib/endpoints/class-wp-rest-attachments-controller.php#L125)
* class-wp-rest-comments-controller.php, line [203](https://github.com/WP-API/WP-API/blob/develop/lib/endpoints/class-wp-rest-comments-controller.php#L203), [267](https://github.com/WP-API/WP-API/blob/develop/lib/endpoints/class-wp-rest-comments-controller.php#L267)
* class-wp-rest-posts-controller.php, line [235](https://github.com/WP-API/WP-API/blob/develop/lib/endpoints/class-wp-rest-posts-controller.php#L235), [309](https://github.com/WP-API/WP-API/blob/develop/lib/endpoints/class-wp-rest-posts-controller.php#L309)
* class-wp-rest-terms-controller.php, line [202](https://github.com/WP-API/WP-API/blob/develop/lib/endpoints/class-wp-rest-terms-controller.php#L202), [260](https://github.com/WP-API/WP-API/blob/develop/lib/endpoints/class-wp-rest-terms-controller.php#L260)
* class-wp-rest-users-controller.php, line [192](https://github.com/WP-API/WP-API/blob/develop/lib/endpoints/class-wp-rest-users-controller.php#L192), [258](https://github.com/WP-API/WP-API/blob/develop/lib/endpoints/class-wp-rest-users-controller.php#L258), [316](https://github.com/WP-API/WP-API/blob/develop/lib/endpoints/class-wp-rest-users-controller.php#L316)

The way I see it there are 3 different possible fixes:
1.  The documentation (this PR) needs to be corrected to allow for a generic array.
2. Each of these instances need to instantiate and use a `WP_REST_Request`
3. Refactor `get_item` to use a helper function to maintain type integrity though I am not sure what this would look like per se.

### DIFF
--- a/lib/endpoints/class-wp-rest-comments-controller.php
+++ b/lib/endpoints/class-wp-rest-comments-controller.php
@@ -110,7 +110,7 @@ class WP_REST_Comments_Controller extends WP_REST_Controller {
 	/**
 	 * Get a comment.
 	 *
-	 * @param  WP_REST_Request $request Full details about the request.
+	 * @param  WP_REST_Request|array $request Full details about the request.
 	 * @return WP_Error|WP_REST_Response
 	 */
 	public function get_item( $request ) {
--- a/lib/endpoints/class-wp-rest-controller.php
+++ b/lib/endpoints/class-wp-rest-controller.php
@@ -23,7 +23,7 @@ abstract class WP_REST_Controller {
 	/**
 	 * Get one item from the collection.
 	 *
-	 * @param WP_REST_Request $request Full data about the request.
+	 * @param WP_REST_Request|array $request Full data about the request.
 	 * @return WP_Error|WP_REST_Response
 	 */
 	public function get_item( $request ) {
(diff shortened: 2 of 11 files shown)


danielbachhuber:
Hm. I think we should be instantiating and using `WP_REST_Request`

joehoyle:
Yeah I'd rather use a `WP_REST_Request`, I don't like the pattern of passing an array, we get away with this because the request object of course implements `ArrayAccess`, however it's a bit annoying as in my own code if done something like `function get_item( WP_REST_Request $request )` which fatals if you pass it an array.

danielbachhuber:
Thanks again for bringing this to our attention, @wpsmith. I've created pull request to address each instance you've identified.

PR #1913, opened 2015-12-28
Be consistent with type and format for URI.

wpsmith:


rachelbaker:
@wpsmith Thank you for the patch.  :cake:

PR #1928, opened 2016-01-01
Pass ID via object when hydrating response.author, including embedded author object if available

westonruter:
Fixes issue where a Backbone model needs an attributes object to be passed into `wp.api.models.User()`, not a bare ID, when hydrating `response.author` in `TimeStampedMixin.parse()`.

Additionally, if the request includes `_embed`, then the embedded author object is supplied instead of just `{ id: response.author }`.

### DIFF
--- a/wp-api.js
+++ b/wp-api.js
@@ -134,6 +134,7 @@
 		 * @returns {*}.
 		 */
 		parse: function( response ) {
+			var authorAttributes;
 
 			// Parse dates into native Date objects.
 			_.each( parseable_dates, function ( key ) {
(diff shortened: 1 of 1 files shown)


adamsilverstein:
@westonruter thanks for the PR's and for using the JS library! 

Sorry for the confusion here: development work on the JS library is happening in a separate repo and merged for releases into the main WP-WPI plugin. See https://github.com/WP-API/client-js

Can you please resubmit your PR's against that repo (note the build files in /js)? Also, you will find the current version has undergone some serious restructuring, although your issues may still apply.

Thanks!

westonruter:
Cheers. See https://github.com/WP-API/client-js/pull/61

PR #1929, opened 2016-01-01
Convert parsed author back to ID in toJSON

westonruter:
The `TimeStampedMixin.parse( response )` function will convert `response.author` from a ID to a `User` model object. However, the `TimeStampedMixin.toJSON()` method is not properly reversing this process to replace a `User` object with the original ID.

See also #1928.

### DIFF
--- a/wp-api.js
+++ b/wp-api.js
@@ -124,6 +124,11 @@
 				}
 			});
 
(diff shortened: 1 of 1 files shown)


westonruter:
Moved to https://github.com/WP-API/client-js/pull/62

PR #1932, opened 2016-01-04
Use `WP_REST_Request` internally in `WP_REST_Attachments_Controller`

danielbachhuber:
See #1908

### DIFF
--- a/lib/endpoints/class-wp-rest-attachments-controller.php
+++ b/lib/endpoints/class-wp-rest-attachments-controller.php
@@ -82,13 +82,14 @@ class WP_REST_Attachments_Controller extends WP_REST_Posts_Controller {
 
 		$this->update_additional_fields_for_object( $attachment, $request );
 
-		$response = $this->get_item( array(
-			'id'      => $id,
-			'context' => 'edit',
-		) );
+		$attachment_location = rest_url( '/wp/v2/' . $this->get_post_type_base( $attachment->post_type ) . '/' . $id );
+		$get_request = new WP_REST_Request( 'GET', $attachment_location );
+		$get_request->set_param( 'id', $id );
+		$get_request->set_param( 'context', 'edit' );
+		$response = $this->get_item( $get_request );
 		$response = rest_ensure_response( $response );
 		$response->set_status( 201 );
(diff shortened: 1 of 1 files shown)


danielbachhuber:
@WP-API/amigos #reviewmerge

lib/endpoints/class-wp-rest-attachments-controller.php
```diff
@@ -122,10 +123,11 @@ public function update_item( $request ) {
 			update_post_meta( $data['id'], '_wp_attachment_image_alt', $request['alt_text'] );
 		}
 
-		$response = $this->get_item( array(
-			'id'      => $data['id'],
-			'context' => 'edit',
-		));
+		$attachment_location = rest_url( '/wp/v2/' . $this->get_post_type_base( get_post_type( $data['id'] ) ) . '/' . $data['id'] );
```
joehoyle: Is it necessary that we pass the method and url to `WP_REST_Request`? Given we are calling `get_item` I don't think it should be required to have a "correct" `WP_REST_Request` for that function. `get_item` just says "get me one of these items, this request object is the source request, by the way". Given we didn't need to generate the object URLs before, I don't see why we need to do it now.

danielbachhuber: > Is it necessary that we pass the method and url to WP_REST_Request?

Huh. That's actually a good question. How do you feel about it? Is there harm in including it?

In my opinion, we'd ideally be calling through the server: https://github.com/WP-API/WP-API/issues/1218#issuecomment-103655685 If we call through the server, we'd need the URL on the request object so the server knows how to appropriately route it.

joehoyle: THere's now harm, it's just added complexity, and I'd rather that creating a `WP_REST_Request` manually wasn't a PITA, so the less stuff the better IMO. However, happy to be convinced otherwise, just based of simpler-is-better, not needing to "know" the URL of the resource is good.

danielbachhuber: I understand that. However, to date, I'd always thought `WP_REST_Request` objects *needed* a URL, just as they need a method, because the URL is the canonical resource identifier. Without a URL, the request object could be passed into different methods missing its original context.

PR #1939, opened 2016-01-04
Use `WP_REST_Request` internally in `WP_REST_Users_Controller`

danielbachhuber:
See #1908


danielbachhuber:
@WP-API/amigos #reviewmerge

PR #1941, opened 2016-01-04
Clean up permissions checks in `WP_REST_Terms_Controller`

danielbachhuber:
* Use `show_in_rest` instead of `public`; rename method.
* Return more descriptive `WP_Error` responses.
* Move 404 check for update term for consistency.


danielbachhuber:
@WP-API/amigos #reviewmerge

PR #1942, opened 2016-01-04
Use `show_in_rest` to determine publicness for post types

### LINKED ISSUE #1895: Inconsistent publicness checks between types and taxonomies
Taxonomies use `show_in_rest` to determine whether the taxonomy is accessible from `/wp/v2/taxonomies`

Types use `public` to determine whether the post type is accessible from `/wp/v2/types`.

I'd _think_ we should use `show_in_rest` to determine whether a type is exposed. `public` was introduced in fb4f59b5a4ad79e112b4770b547ca915b0d516e5

Discovered in #1894

danielbachhuber:
Fixes #1895

### DIFF
--- a/lib/endpoints/class-wp-rest-post-types-controller.php
+++ b/lib/endpoints/class-wp-rest-post-types-controller.php
@@ -36,8 +36,8 @@ class WP_REST_Post_Types_Controller extends WP_REST_Controller {
 	 */
 	public function get_items( $request ) {
 		$data = array();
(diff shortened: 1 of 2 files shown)


danielbachhuber:
@WP-API/amigos #reviewmerge

picklepete:
Edit: woops, wrong PR, see my comment [here](https://github.com/WP-API/WP-API/pull/1279#issuecomment-189213180).

PR #1945, opened 2016-01-04
Don't include method and URL in internal `WP_REST_Request` (Comments)

danielbachhuber:
While this change doesn't have any functional changes, it's in line with
our pattern throughout the rest of the codebase.

See https://github.com/WP-API/WP-API/pull/1932#discussion_r48764666


danielbachhuber:
@WP-API/amigos #reviewmerge

PR #1946, opened 2016-01-04
Don't include method and URL in internal `WP_REST_Request` (Term)

danielbachhuber:
Related #1945


danielbachhuber:
@WP-API/amigos #reviewmerge

PR #1949, opened 2016-01-05
Flag sticky if password not empty instead of isset

westonruter:
If attempting to submit a `password` consisting of an empty string while also submitting true value for `sticky`, the API is returning an error:

> A post can not be sticky and have a password.

I believe it should be considering an empty string to be as if the value was not set.

### DIFF
--- a/lib/endpoints/class-wp-rest-posts-controller.php
+++ b/lib/endpoints/class-wp-rest-posts-controller.php
@@ -747,7 +747,7 @@ class WP_REST_Posts_Controller extends WP_REST_Controller {
 		}
 
 		// Post password.
-		if ( isset( $request['password'] ) ) {
+		if ( isset( $request['password'] ) && '' !== $request['password'] ) {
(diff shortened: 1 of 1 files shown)


danielbachhuber:
@westonruter Can you update the tests to fix the build?

joehoyle:
I'm not sure why this is required - this means I can't set a password to string `0`.

danielbachhuber:
So the most precise check would be to ensure `password` is either:

1. Not set.
1. An empty string.

westonruter:
Right, my choice of `empty` was short-sighted. 

So instead of:

```php
if ( ! empty( $request['password'] ) ) {
```

It should be:

```php
if ( isset( $request['password'] ) && '' !== $request['password'] ) {
```

Amended commit in b6a349d.

danielbachhuber:
@westonruter can you add to the `test_create_post_with_password_and_sticky_fails` test to include a condition where the password is an empty string?

PR #1951, opened 2016-01-05
In `prepare_date_response` parse $date if available

adamsilverstein:
- Moves the $date_gmt blank check below the check for $date available, so $date is used if passed
- Fixes an issue where creating a new post with the API returned an object with all datetime fields set to null.
- After this fix, the API properly returns parsed dates for `date` and `date_modified` and `null` for `date_gmt` and `date_modified_gmt`, matching the database state


danielbachhuber:
> Fixes an issue where creating a new post with the API returned an object with all datetime fields set to null.

@adamsilverstein Can you share an example of this, so we have full history of the issue?

adamsilverstein:
@danielbachhuber:

Sure, I am using the JS client to create a new post. the newly created post has null values returned for the date fields by the api before this patch.

Before the patch the following two date values are null, after the patch they show the proper date:

screenshot

PR #1961, opened 2016-01-07
Add `include` parameter to `GET /wp/v2/posts`

danielbachhuber:
Permits limiting results to a specific defined set

See #1668

PR #1962, opened 2016-01-07
Ensure `GET /wp/v2/posts` respects `order` and `orderby` params

danielbachhuber:
Instead of silently discarding them.

Discovered in #1961
Related #924


danielbachhuber:
@WP-API/amigos #reviewmerge

PR #1963, opened 2016-01-07
Remove duplicated query params from Terms controller

danielbachhuber:
`page`, `per_page` and `search` are inherited from the base controller


danielbachhuber:
@WP-API/amigos #reviewmerge

Issue #1668, opened 2015-10-20
Add `include` parameter for all collections requests


danielbachhuber:
Each of these requests should support an `include` parameter (and `orderby=include`) where the value is one or more resource IDs:

* [x] `GET /wp/v2/posts`
* [x] `GET /wp/v2/comments`
* [x] `GET /wp/v2/users`
* [x] `GET /wp/v2/terms/<taxonomy>`

From https://github.com/WP-API/WP-API/issues/1587

danielbachhuber:
Also, when we're implementing this, we should look into the history of why `post__in` isn't a public WP_Query var to make sure we aren't creating a security issue.

danielbachhuber:
> we should look into the history of why post__in isn't a public WP_Query var to make sure we aren't creating a security issue.

Just to follow up on this — there isn't the potential for a security issue because `GET /wp/v2/posts include=1,2` is functionally equivalent to making two separate `GET` requests. We're still checking read permission on each post. I suppose `post__in` isn't public because WordPress themes don't typically check read permissions.

joehoyle:
Hmm, I'm not sure - even with `post__in` WP_QUery adds it's other conditions, so you can circuit any checks, for example it will do `SELECT * FROM x WHERE post_status=publish AND ID IN (1,2,3,)`

danielbachhuber:
:ship: :ship:

PR #1964, opened 2016-01-07
Add `include` parameter to `GET /wp/v2/users`

danielbachhuber:
See #1668


danielbachhuber:
@WP-API/amigos #reviewmerge

PR #1966, opened 2016-01-07
Cut Travis' load in half

danielbachhuber:
Ensures only one build is created on pull requests, instead of two.

See https://github.com/alleyinteractive/wordpress-fieldmanager/pull/430#issuecomment-169696750

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -77,8 +77,9 @@ git:
   depth: 1
 
 branches:
-    only:
-        - master
+  only:
+    - master
+    - develop
 
 notifications:
   email:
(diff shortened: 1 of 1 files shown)


danielbachhuber:
Hm. No builds created :(

mboynes:
I think you would need to also add `develop`, if that's what most of your PRs go into.

danielbachhuber:
> I think you would need to also add develop, if that's what most of your PRs go into.

Derp. Good call :)

danielbachhuber:
Huh. I wonder why `codecov` build isn't triggered.

danielbachhuber:
`codecov` is in there now:

image

PR #1968, opened 2016-01-08
Add `include` parameter to `GET /wp/v2/comments`

danielbachhuber:
Also explicitly declares `order` and `orderby` parameters

See #1668


danielbachhuber:
@WP-API/amigos #reviewmerge

PR #1973, opened 2016-01-08
Permit logged-in users to query for media

### LINKED ISSUE #1959: GET request for media files results in an empty array
Hi

I have been trying to fetch all the media files from the REST API using restangular, but no matter how I try it fails, so I decided to test the media endpoint via PostMan app and realized that although I can see the media files json object in my browsers when I browse to /wp-json/wp/v2/media but using PostMan or using restangular it always results in an empty array.

The funny part is that during the time I was pulling my hair for getting the media files, the files showed up only once, and after one refresh they were gone forever, this could be a bug in the wp rest plugin.

danielbachhuber:
Fixes #1959

### DIFF
--- a/lib/endpoints/class-wp-rest-posts-controller.php
+++ b/lib/endpoints/class-wp-rest-posts-controller.php
@@ -1604,7 +1604,7 @@ class WP_REST_Posts_Controller extends WP_REST_Controller {
 			),
 		);
 		$params['status'] = array(
-			'default'           => 'publish',
+			'default'           => 'attachment' === $this->post_type ? 'inherit' : 'publish',
 			'description'       => __( 'Limit result set to posts assigned a specific status.' ),
 			'sanitize_callback' => 'sanitize_key',
 			'type'              => 'string',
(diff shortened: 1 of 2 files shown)


danielbachhuber:
@WP-API/amigos #reviewmerge

lib/endpoints/class-wp-rest-posts-controller.php
```diff
@@ -1604,7 +1604,7 @@ public function get_collection_params() {
 			),
 		);
 		$params['status'] = array(
-			'default'           => 'publish',
+			'default'           => 'attachment' === $this->post_type ? 'inherit' : 'publish',
```
joehoyle: If I do a request logged out for `/media` and there's a bunch of posts with `post_status=inherit` which are attached to private posts, what does the response look like? Does it just remove the attachments I can't view (thus I get 6 results rather than the default 10) or do I get an error, or somethign else?

danielbachhuber: > Does it just remove the attachments I can't view (thus I get 6 results rather than the default 10)

Yes, which is a consistent pattern we have throughout the controllers.

joehoyle: :+1:

PR #1975, opened 2016-01-08
Declare `parent` query param for Pages

danielbachhuber:
Adds tests for Post and Page query params

See #924


danielbachhuber:
@WP-API/amigos #reviewmerge

PR #1976, opened 2016-01-09
Remove unnecessary since it exists in permission callback

frozzare:
`get_item` in users controller has a check for invalid user that is unnecessary since the same check exists in the [permission callback](https://github.com/WP-API/WP-API/blob/develop/lib/endpoints/class-wp-rest-users-controller.php#L398) so the check in `get_item` will never return a `WP_Error`

### DIFF
--- a/lib/endpoints/class-wp-rest-users-controller.php
+++ b/lib/endpoints/class-wp-rest-users-controller.php
@@ -169,10 +169,6 @@ class WP_REST_Users_Controller extends WP_REST_Controller {
 		$id = (int) $request['id'];
 		$user = get_userdata( $id );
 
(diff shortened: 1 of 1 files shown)


danielbachhuber:
I'm not sure I agree with this, because I'd expect to get an error returned, and not a fatal, if I called `get_item()` directly.

frozzare:
It depends on how it should be used, if it should be used to call directly the error check should exists but if not I don't think is necessary to have it there since it will never fail there.

danielbachhuber:
Ok. We're going to keep it in for now, because it's safer code and there's no harm in keeping it. Thanks for the PR, though.

PR #1977, opened 2016-01-09
Introduce new class WP_REST_Item_Schema_Transformer.

joehoyle:
This class is responsible for transforming an internal REST API schema
into various other formats, such as the REST Server Endpoint arguments,
JSON schema etc.

This is currently non-working, proof of concept on how we could pull the schema stuff out of WP_REST_Controller. I havn't applied any internal API change to `get_item_schem` yet - but basucally we'd remove the JSON schema specific stuff (as JSON schema is just a transform on the internal schema) and get rid of the `arg_options` weirdness.

The "internal" REST API schema, is in practice inspired by JSON schema still, will additional things like the arg options also as part of the same language. Think of it as a "fields registration" language, but we won't call it that for political reasons :) 

The main reason we need this, is something like JSON schema it's self doesn't provide enough definition for things like endpoint arguments and validation etc.

### DIFF
--- a/lib/endpoints/class-wp-rest-controller.php
+++ b/lib/endpoints/class-wp-rest-controller.php
@@ -186,7 +186,7 @@ abstract class WP_REST_Controller {
 	}
 
 	/**
-	 * Get the item's schema, conforming to JSON Schema.
+	 * Get the item's schema.
 	 *
 	 * @return array
 	 */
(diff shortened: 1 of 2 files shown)


danielbachhuber:
What about being able to handle collection params interchangeably?

joehoyle:
I'm going to close this out, as I'm not confident this approach improves much, at the cost of another layer of leaky abstraction. Consider this rescinded.

PR #1978, opened 2016-01-09
Add latest client js including minified version

adamsilverstein:
* Full js client development log at https://github.com/WP-API/client-js
* When SCRIPT_DEBUG off, use minified script

JavaScript client change highlights:
* Api builds models and collections from Schema, no hard coded models or collections; any endpoint in the schema will be mapped; supports multiple schemas (apis); helper methods for post model: all helpers user _embed data if available, otherwise using ajax request.
* new: Async load of api, apps can use the deferred to time startup
* new: Ability to localize schema, falls back to making an ajax request & cached in session storage; develop branch of client js has code that only localizes data once per session
* new Helper methods for Posts model: getCategories and setCategories
* fix: cleanup TimeStampedMixin
* new: models/collections get defaults and options from schema
* fix: all code up to WordPress JavaScript coding standards
* new: helper method to get post author user model `getAuthorUser`
* new: helper method to get a featured image for a post: `getFeaturedImage`
* new: models protect methods based on endpoint methods (eg. destory is blocked if an endpoint doesn't have a 'DELETE' method)
* new: models contain a reference to the original route object
* note: all helper getters return a promise
* note: added latest docs here: https://github.com/WP-API/docs-v2/pull/82

On the roadmap:
* Complete helper method suite
* Add name mapping so we can get non-plural names for all models and plural names for all collections


westonruter:
Props list? :)

adamsilverstein:
@westonruter Sure! how do i do that? should i add a props line with @ mentions to the PR comment? 

i believe when this get merged into core everyone with commits will get props on the merge

westonruter:
Since this is essentially a squashed merge commit from the client-js repo, I suggest the commits in this PR be rebased into a single commit and that the commit's description include a link to GitHub compare view for the range of commits introduced, and include the output of `git shortlog -sn`, with @-mentions of contributors' GitHub usernames. The same info can be included in the PR description.

adamsilverstein:
@westonruter you make that sound so easy: i guess it is after you do it a few times :) I'll give it a shot.

adamsilverstein:
rework this pr with weston's suggestions

adamsilverstein:
@westonruter 

I created a new PR - https://github.com/WP-API/WP-API/pull/1981, that was easier :) I left off schema localization for now on the PHP side, I'm concerned about grabbing that data to every page request - working on ways to refine that.

PR #1981, opened 2016-01-10
Add latest client js including minified version

adamsilverstein:
- Development happening at https://github.com/WP-API/client-js

Full changeset: https://github.com/WP-API/client-js/compare/22b1e4219afcb2204d5b79a95b36a6971f7548a6...477a49f029c6503fe680322910bef20809ec552b

Props: @delphinus35, @adamsilverstein, @westonruter

Change highlights:
- Api builds models and collections from Schema, no hard coded models or collections; any endpoint in the schema will be mapped; supports multiple schemas (apis); helper methods for post model: all helpers user _embed data if available, otherwise using ajax request.
- new: Async load of api, apps can use the deferred to time startup
- new: Ability to localize schema, falls back to making an ajax request & cached in session storage; develop branch of client js has code that only localizes data once per session
- new Helper methods for Posts model: getCategories and setCategories
- fix: cleanup TimeStampedMixin
- new: models/collections get defaults and options from schema
- fix: all code up to WordPress JavaScript coding standards
- new: helper method to get post author user model `getAuthorUser`
- new: helper method to get a featured image for a post: `getFeaturedImage`
- new: models protect methods based on endpoint methods (eg. destory is blocked if an endpoint doesn't have a 'DELETE' method)
- new: models contain a reference to the original route object
- Ensure getters always return promise
- Add remapping for names, see https://github.com/WP-API/client-js/pull/75
- additional helpers: add getMeta, getTags, getRevisions #76
- When SCRIPT_DEBUG off, use minified script; update localized data


westonruter:
I suggest `amend`-ing commit description in e079114 to replace `master` with the commit at https://github.com/WP-API/WP-API/commit/7573f8a19e36f16d8d88445b418b3108186a84d1 (current state of `master`) so that the compare view will remain true after additional changes are pushed to the branch.

adamsilverstein:
@westonruter updated the comparison link in commit and pr, thanks for the tip. i'm doing the comparison link for HEAD in clientjs against where we started with last merge so you can review the full commit log.

adamsilverstein:
force pushing my amended commit message resulted in a new commit hash: https://github.com/WP-API/WP-API/commit/6a77c408b8bc09ae706fd8c396628e9c1d99682a

westonruter:
Perfect :+1:

rachelbaker:
@danielbachhuber are you keeping this open so it can be updated right before the Beta 11 release?  Or can it be merged now?

danielbachhuber:
> are you keeping this open so it can be updated right before the Beta 11 release? Or can it be merged now?

Heh. It didn't have the `review` label, so I forgot about it :(

rmccue:
@adamsilverstein Looks like this didn't include the `schema` data that the JS plugin uses, so it causes a bunch of JS errors on the page.

adamsilverstein:
@rmccue yes, i left that out on purpose: i don't want to localize the data on each page. I have a branch in the works that attempts to localize the data once and only once per session: https://github.com/WP-API/client-js/commits/features/only-localize-schma-once

The JS should accommodate and make an ajax request for the schema data if its not localized. I thought I had checked this,assuming you cleared cache/opened a new window? 

I will dig back in and make sure this is working properly. In the mean time, copy the localized data build over from the client repo.

adamsilverstein:
@rmccue testing with latest develop, client js plugin disabled. not getting js errors, setup from ajax callback works fine. might be related to model name changes? what are the specific errors you are getting?

adamsilverstein:
Actually: I am getting a JS error now: `Uncaught TypeError: Cannot read property 'default' of null` - I think this is coming from the session cached schema copy... digging.

adamsilverstein:
Thanks for pointing that out - removing the default post processing fixes the issue with session storage: fixed in https://github.com/WP-API/client-js/commit/06987b63992f0c0b3c3cdd6c640be578df64269d : update PR incoming

adamsilverstein:
@rmccue fixed in https://github.com/WP-API/WP-API/pull/2022

PR #1990, opened 2016-01-11
Drop `orderby=>term_id`; add tests for `orderby=>id`

### LINKED ISSUE #1972: GET Terms: Drop unnecessary `term_id` from orderby enum
`id` is the one true id for terms. But, in dropping `term_id`, we should make `orderby=>id` does what we expect it to.

danielbachhuber:
Fixes #1972


danielbachhuber:
@WP-API/amigos #reviewmerge

joehoyle:
:+1:

PR #1992, opened 2016-01-11
Remove unnecessary `is_wp_error()` check

danielbachhuber:
`rest_ensure_response()` always returns a response object

Added unnecessarily in efb41e467506fe4e24f4f2e4b66576f7f5faf64e


danielbachhuber:
@WP-API/amigos #reviewmerge

danielbachhuber:
> rest_ensure_response() always returns a WP_REST_Response object

As it turns out, this is incorrect.

Issue #153, opened 2014-04-29
Set appropriate HTTP status when passing through WP_Error


rmccue:
As noted by @tobych in #146, core functions that return `WP_Error`s currently all have a 500 status code. However, we should be able to parse out a better status code based on the code in the error.

danielbachhuber:
I skimmed through all instances where we're checking `is_wp_error()`. Here are the places we're looking at the return value of core PHP APIs:
- [x] `wp_insert_post()` https://github.com/WP-API/WP-API/blob/1d70d3958270ff2e46bb894033ad56ba1b91bd48/lib/endpoints/class-wp-rest-attachments-controller.php#L69-L71
- [x] `wp_insert_post()` https://github.com/WP-API/WP-API/blob/1d70d3958270ff2e46bb894033ad56ba1b91bd48/lib/endpoints/class-wp-rest-posts-controller.php#L192-L202
- [ ] `wp_set_object_terms()` https://github.com/WP-API/WP-API/blob/1d70d3958270ff2e46bb894033ad56ba1b91bd48/lib/endpoints/class-wp-rest-posts-terms-controller.php#L136-L140
- [ ] `wp_remove_object_terms()` https://github.com/WP-API/WP-API/blob/1d70d3958270ff2e46bb894033ad56ba1b91bd48/lib/endpoints/class-wp-rest-posts-terms-controller.php#L182-L186
- [ ] `get_term()` https://github.com/WP-API/WP-API/blob/1d70d3958270ff2e46bb894033ad56ba1b91bd48/lib/endpoints/class-wp-rest-terms-controller.php#L161-L167
- [ ] `wp_insert_term()` https://github.com/WP-API/WP-API/blob/1d70d3958270ff2e46bb894033ad56ba1b91bd48/lib/endpoints/class-wp-rest-terms-controller.php#L206-L218
- [ ] `wp_update_term()` https://github.com/WP-API/WP-API/blob/1d70d3958270ff2e46bb894033ad56ba1b91bd48/lib/endpoints/class-wp-rest-terms-controller.php#L269-L273
- [ ] `WP_User_Query()` https://github.com/WP-API/WP-API/blob/1d70d3958270ff2e46bb894033ad56ba1b91bd48/lib/endpoints/class-wp-rest-users-controller.php#L117-L120
[...]

PR #1993, opened 2016-01-11
Add a more helpful status code when Attachment insert fails

danielbachhuber:
This follows a similar pattern we use when creating a Post.

See #153


danielbachhuber:
@WP-API/amigos #reviewmerge

PR #1994, opened 2016-01-11
Remove another unnecessary `is_wp_error()` check

danielbachhuber:
`rest_ensure_response()` always returns a `WP_REST_Response` object


danielbachhuber:
@WP-API/amigos #reviewmerge

danielbachhuber:
> rest_ensure_response() always returns a WP_REST_Response object

As it turns out, this is incorrect.

Issue #1970, opened 2016-01-08
Add `exclude` parameter for all collections requests


danielbachhuber:
Each of these requests should support an `exclude` parameter where the value is one or more resource IDs:

* [x] `GET /wp/v2/posts`
* [x] `GET /wp/v2/comments`
* [x] `GET /wp/v2/users`
* [x] `GET /wp/v2/<taxonomy>`

Related #1668, #924

danielbachhuber:
:ship:

PR #2000, opened 2016-01-12
Add `exclude` param to `GET /wp/v2/users`

danielbachhuber:
See #1970, #924


danielbachhuber:
@WP-API/amigos #reviewmerge

PR #2005, opened 2016-01-12
Use discrete `title` for categories and tags

danielbachhuber:
They're separate schemas, so they should be named differently


danielbachhuber:
@WP-API/amigos #reviewmerge

rachelbaker:
@danielbachhuber Unit tests for `register_rest_field` with the Terms controller would be :100:

danielbachhuber:
> Unit tests for register_rest_field with the Terms controller would be

Good call. Updating.

danielbachhuber:
@WP-API/amigos updated with `test_get_additional_field_registration`

Issue #1979, opened 2016-01-10
Querying comments by post_slug is protected


patrickgalbraith:
Why does querying comments by `post_slug` require authentication but by `post_id` it doesn't?

This is causing me issues for something I am building since I cannot request the post and the comments in parallel since I don't know the post's ID from the URL. So I have to wait for the post request to finish before requesting the comments. I can't see any reason why most of those arguments are protected (see code below).

```php
$prepared_args = array(
  'number'  => $request['per_page'],
  'post_id' => $request['post'] ? $request['post'] : '', // <--- Not authenticated
  'parent'  => isset( $request['parent'] ) ? $request['parent'] : '',
  'search'  => $request['search'],
  'orderby' => $this->normalize_query_param( $order_by ),
  'order'   => $request['order'],
  'status'  => 'approve',
  'type'    => 'comment',
);

$prepared_args['offset'] = $prepared_args['number'] * ( absint( $request['page'] ) - 1 );

if ( current_user_can( 'edit_posts' ) ) {
  $protected_args = array(
    'user'         => $request['user'] ? $request['user'] : '',
    'status'       => $request['status'],
    'type'         => isset( $request['type'] ) ? $request['type'] : '',
    'author_email' => isset( $request['author_email'] ) ? $request['author_email'] : '',
    'karma'        => isset( $request['karma'] ) ? $request['karma'] : '',
    'post_author'  => isset( $request['post_author'] ) ? $request['post_author'] : '',
    'post_name'    => isset( $request['post_slug'] ) ? $request['post_slug'] : '', // <--- Authenticated :(
    'post_parent'  => isset( $request['post_parent'] ) ? $request['post_parent'] : '',
    'post_status'  => isset( $request['post_status'] ) ? $request['post_status'] : '',
    'post_type'    => isset( $request['post_type'] ) ? $request['post_type'] : '',
  );

  $prepared_args = array_merge( $prepared_args, $protected_args );
}
```

On another note the comments `_embedded` in the post also can't be ordered which is a pain since I am showing the comments in `asc` order and the api doesn't take into account the discussion settings in WordPress. Which means I need to do additional api requests for the comments.

![image](https://cloud.githubusercontent.com/assets/811945/12220651/4f23c406-b7c9-11e5-8121-5f345591c01e.png)

patrickgalbraith:
Also shouldn't attempting to use a protected argument throw some kind of error.

danielbachhuber:
[...]

Issue #1167, opened 2015-05-02
Audit filters and actions


danielbachhuber:
Let's make sure our controllers have a consistent pattern of filters and actions.

peterigz:
It would be nice if the "get_items" controller function had a filter before returning back the response, passing in the $query and the array of items. This way you can add more data based on the query as a whole rather then just individual items (with "rest_prepare_" filter). For example when retrieving all the posts for a particular category, it would be good to get the archive title/description back with the data as well, plus other tags that are available outside of *the loop*.

It would also be good if the data structure was slightly different for returning multiple items as well like:

    data {
      collection [post, post] ...
      archive_title: 'the title',
      archive_description: 'description',
      sidebar: '<markup>',
      etc.
    }

I'm loving this plugin, hope it becomes part of wordpress core soon :)

danielbachhuber:
We should make sure we consistently pass `WP_REST_Request` and `WP_REST_Response` to our actions and filters https://github.com/WP-API/WP-API/issues/1238#issuecomment-102147919

rachelbaker:
Punted to Beta 4

BE-Webdesign:
Should we also list the filters and what not on the v2.wp-api.org site?

danielbachhuber:
> Should we also list the filters and what not on the v2.wp-api.org site?

Yes, we could if someone wanted to set that up.

BE-Webdesign:
Ya I will.  What do we need to audit about the filters just to make sure we have a consistent structure and naming to them in the controllers?

danielbachhuber:
> What do we need to audit about the filters just to make sure we have a consistent structure and naming to them in the controllers?

More or less.

BE-Webdesign:
**extras.php**

| Line | Filter | Function |
| --- | --- | --- |
| :81 | `apply_filters( 'rest_avatar_sizes', array( 24, 48, 96 ) );` | `rest_get_avatar_sizes()` |

**lib/endpoints/class-wp-rest-attachments-controller.php**

| Line | Filter | Function |
| --- | --- | --- |
| :152 | `do_action( 'rest_insert_attachment', $attachment, $request, true );` | `create_item()` |
| :186 | `do_action( 'rest_insert_attachment', $data, $request, false );` | `update_item()` |
| :288 | `apply_filters( 'rest_prepare_attachment', $response, $post, $request );` | `prepare_item_for_response()` |

**lib/endpoints/class-wp-rest-comments-controller.php**

[...]

PR #2007, opened 2016-01-12
Add `rest_comment_query` for filtering `GET /wp/v2/comments`

danielbachhuber:
See #1167
Related https://github.com/WP-API/WP-API/issues/1979#issuecomment-170403518


danielbachhuber:
@WP-API/amigos #reviewmerge

PR #2008, opened 2016-01-12
Make `rest_prepare_term` dynamic based on taxonomy

danielbachhuber:
This is consistent with our Posts controller.

See #1167


danielbachhuber:
@WP-API/amigos #reviewmerge

PR #2009, opened 2016-01-12
Properly quote concatenated filter name string

danielbachhuber:
This ensures the parser can pick up the full representation.

See #1167


danielbachhuber:
@WP-API/amigos #reviewmerge

PR #2010, opened 2016-01-12
Use dynamic filter names for `rest_post_query` and `rest_terms_query`

danielbachhuber:
Because these two controllers can be instantiated for different types
and taxonomies, it makes sense for their filters to be named based on
the type / taxonomy

See #1167


danielbachhuber:
@WP-API/amigos #reviewmerge

PR #2011, opened 2016-01-12
Use `developer.wordpress.org` links instead of the Codex

danielbachhuber:
It's the latest and greatest.

See #1167


danielbachhuber:
@WP-API/amigos #reviewmerge

PR #1197, opened 2015-05-06
Pagination headers for comment collection queries

danielbachhuber:
Fixes #1186

PR #2015, opened 2016-01-13
Avoid a duplicate query for the comment count

rmccue:
Back in 768a2826eb20c86c189b692248419627e89db1c2 (#1197), we introduced the regular pagination headers for comment queries. However, apparently we missed that `WP_Comment_Query` has `SQL_CALC_FOUND_ROWS` built-in (but off by default), so we're doing a complex query twice here.

Attached patch uses `$query->found_comments` and `$query->max_num_pages` instead of redoing the query.

### DIFF
--- a/lib/endpoints/class-wp-rest-comments-controller.php
+++ b/lib/endpoints/class-wp-rest-comments-controller.php
@@ -64,7 +64,6 @@ class WP_REST_Comments_Controller extends WP_REST_Controller {
 	public function get_items( $request ) {
 		$prepared_args = $this->prepare_items_query( $request );
 
-
 		/**
 		 * Filter arguments, before passing to WP_Comment_Query, when querying comments via the REST API.
 		 *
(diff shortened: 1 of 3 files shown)


rmccue:
This was introduced originally in #1209, since (as the failing test shows) out-of-bounds pages will give back 0 for `SQL_CALC_FOUND_ROWS`. Unfortunately, this hurts common-case performance (in-bounds requests) to support an edge case (out-of-bounds requests).

Changed this to using the normal query on in-bounds requests, and only running the secondary query if we're out of bounds. Updating other controllers now.

rmccue:
@WP-API/amigos #reviewmerge

danielbachhuber:
Looks good :+1: And well-covered by existing tests.

danielbachhuber:
Related:
- https://core.trac.wordpress.org/ticket/29178#comment:3
- https://core.trac.wordpress.org/changeset/34544

Although, in this case, I think this PR is still better than what we had before.

PR #2022, opened 2016-01-13
Update js client, fix session storage bug

adamsilverstein:
not much is changing here despite the diff. two big functions got moved; for changes see see https://github.com/WP-API/client-js/pull/78


rmccue:
:heart_eyes:

Issue #2021, opened 2016-01-13
Conditionally define classes and functions


danielbachhuber:
In preparation for an eventual merge into core, we need to conditionally define our classes and functions. This will ensure a trunk WordPress instance running the REST API plugin doesn't fatal.

danielbachhuber:
:ship:

PR #2023, opened 2016-01-14
Conditionally require REST API classes

danielbachhuber:
When these eventually get committed into core, we want the plugin to
gracefully use the classes included in core

See #2021


danielbachhuber:
@WP-API/amigos #reviewmerge

Issue #642, opened 2014-11-26
GET resources by slug


danielbachhuber:
A pretty common need is to be able to get a post by a slug. What if would it look like if you could get any resource by id or slug?

* `Post` - Type would need to be indicated in the route, as posts of two different types can have the same `post_name`.
* `Term` - Taxonomy would need to be indicated in the route, as two terms of different taxonomies can have the same `slug`.
* `Comment` - No such thing as a comment slug.
* `Meta` - Meta keys are sometimes unique. We can't rely on it.
* `User` - User's slug is the `user_nicename` which can be relied upon to be unique (as it's used in author URLs) and has an index in the db.

danielbachhuber:
[This has been addressed](https://wordpress.slack.com/archives/core-restapi/p1427759082000510).

For querying by slugs, our pattern will be `/wp/posts?slug=<slug>`

Related #924

baffleinc:
Hmm... hitting `/wp-json/posts?slug=why-us` returns my 10 most recent blog posts instead of the why us page.. Am I missing something?

danielbachhuber:
> Am I missing something?

Not implemented in `develop` yet.

baffleinc:
Ah ok... what's the best solution at present?

danielbachhuber:
Use `name` instead

baffleinc:
Yeah... 

/wp-json/posts?name=why-us - 10 most recent posts
/wp-json/posts?filter[name]=why-us - empty array

both stopped working after they used to work A-OK. Only since updating to 1.2. I'm not crazy, the page exists: https://www.dropbox.com/s/27ceykrau45de62/Screenshot%202015-04-02%2010.43.58.png?dl=0

danielbachhuber:
> both stopped working after they used to work A-OK.

Oh. Can you open a new issue then? Might be a bug specific to 1.2

baffleinc:
Sure. I tried downgrading, uninstalling, reinstalling, turning all plugins off again and having the same problem

As it turns out no filters are working - can you imagine any reason why? 

1. http://www2.health.com.au/wp-json/posts/?filter[name]=why-us - returns empty array
2. http://www2.health.com.au/wp-json/posts/?filter[type][]=page - returns most recent posts (not pages)
3. http://www2.health.com.au/wp-json/posts/?filter[page]=3 - returns first page of posts (same as #2)

Would really appreciate some help as it has been working fine.

baffleinc:
[...]

Issue #576, opened 2014-11-04
Exposing rewrite rules through the API


goblindegook:
Some time ago I extended the API to expose permalink settings in a format fit for consumption by a Backbone.Marionette application. Here's the relevant class:

https://github.com/B3ST/B3-REST-API/blob/master/helpers/B3_RoutesHelper.php

(This handles root permalinks as well as routes for search, authors, post types—if registered with the `show_in_json` argument as `true`—taxonomies, posts, date archives and pages.)

My search-fu failed me and I couldn't find anything in the issue tracker, so before I go ahead and massage this code into a pull request, I'd like to make sure some kind of discussion has been had.

What format is the API supposed to return routes in? Should WP routes be considered a resource of their own? Should it be returned by the API at all?

danielbachhuber:
Flagged this as an enhancement for 2.1. It's quite related to https://github.com/WP-API/WP-API/issues/1776#issuecomment-162918432 too.

PR #2025, opened 2016-01-14
Remove `GET /wp/v2/comments` support for `post_*` params

danielbachhuber:
Querying by `post_type` and other params have performance implications
due to the join on the posts table. Furthermore, Comments is the only
endpoint to support this type of query, and it doesn't make sense to
`post_*` to other endpoints.

See #1979
Related #924

### DIFF
--- a/lib/endpoints/class-wp-rest-comments-controller.php
+++ b/lib/endpoints/class-wp-rest-comments-controller.php
@@ -612,11 +612,6 @@ class WP_REST_Comments_Controller extends WP_REST_Controller {
 				'type'         => isset( $request['type'] ) ? $request['type'] : '',
 				'author_email' => isset( $request['author_email'] ) ? $request['author_email'] : '',
 				'karma'        => isset( $request['karma'] ) ? $request['karma'] : '',
-				'post_author'  => isset( $request['post_author'] ) ? $request['post_author'] : '',
-				'post_name'    => isset( $request['post_slug'] ) ? $request['post_slug'] : '',
-				'post_parent'  => isset( $request['post_parent'] ) ? $request['post_parent'] : '',
-				'post_status'  => isset( $request['post_status'] ) ? $request['post_status'] : '',
-				'post_type'    => isset( $request['post_type'] ) ? $request['post_type'] : '',
 			);
 
 			$prepared_args = array_merge( $prepared_args, $protected_args );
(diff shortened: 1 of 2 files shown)


danielbachhuber:
@WP-API/amigos Thoughts on this?

joehoyle:
The only one I can see being useful in practice is `post_name`, as it's feasible you don't know the ID, though I've never been a fan of treating slugs specially. I'm, +1 for removing all these.

kadamwhite:
> The only one I can see being useful in practice is post_name, as it's feasible you don't know the ID, though I've never been a fan of treating slugs specially.

One of the first conclusions I came to building a rest application was that WP's ID was meaningless to the rest of my application, so literally every query ended up being a search by slug. They aren't "meant" to be treated specially, but they're a piece of usually-privileged-by-interface (URL) post content that actually does hold its significance across multiple systems, which ID only does through the semantic weight we give it as developers.

_i.e._, in #1979 I agree with this sentiment:

> This is causing me issues for something I am building since I cannot request the post and the comments in parallel since I don't know the post's ID from the URL.

If the performance is significantly impacted I could be swayed, but I'd be :-1: on removing the ability to search comments by `post_slug`. I've largely come to accept that to do things properly with this API I will always need to be making multiple requests, e.g. a discovery request to find the endpoint prior to querying against it, but for this particular case it feels like it's just an internal implementation detail that's forcing us to make two requests (one by name to get the post, then another by ID to get comments).

danielbachhuber:
> If the performance is significantly impacted I could be swayed, but I'd be :-1: on removing the ability to search comments by `post_slug`.

I think the problem you're trying to solve calls for a different solution, not to add `post_slug` as a query parameter to every endpoint.

I'd propose the problem you're trying to solve is that, within the context of a JS client / application where state is initially determined by URL path, it can be difficult to determine the post ID from the URL. We should solve that problem (#576) independently. If you have easy access to the post ID, or even the entire post object, I'm sure the pain you've felt in creating your application would be moot.

`post_slug` can't be considered unambiguous (#642), and shouldn't be exposed as a top-level query param across all of our controllers.

rmccue:
:-1: on removing this. I don't think we should be overly concerned about the performance here unless it turns out to be a real problem.

> Furthermore, Comments is the only endpoint to support this type of query

That's intentional though. Every type of object is different, and one of the things that defines comments is that they "belong" to a post. I think it makes sense to be able to query by these.

danielbachhuber:
> That's intentional though. Every type of object is different, and one of the things that defines comments is that they "belong" to a post. I think it makes sense to be able to query by these.

I thought comments now _optionally_ are assigned to a post?

If attachments can be optionally assigned to a post, why don't we support querying attachments by `post_slug`?

patrickgalbraith:
I suggested this in the other thread but why not do something like `/wp-json/wp/v2/comments?filter[post_name]=some-post` so `filter` becomes a way to pass arguments to modify the `WP_Comment_Query`. Similar to what is done on the posts endpoint that way we get both the functionality and it is consistent with the other endpoints.

danielbachhuber:
> I suggested this in the other thread but why not do something like /wp-json/wp/v2/comments?filter[post_name]=some-post so filter becomes a way to pass arguments to modify the WP_Comment_Query.

Filed this as #2066

Going to close the PR for now as it's already stale. We can continue discussing on the issue.

PR #2027, opened 2016-01-14
Add helpful description to `force` argument

danielbachhuber:
Distinguish between resources that support trashing, and those that
don't


danielbachhuber:
@WP-API/amigos #reviewmerge

rachelbaker:
@danielbachhuber waiting for tests to pass...

PR #2028, opened 2016-01-14
Require `force` argument for `DELETE /wp/v2/<taxonomy>/<id>`

danielbachhuber:
While terms don't support trashing, we've established a pattern of
requiring the `force` param for all `DELETE` requests, to give us the
opportunity to introduce trashing in the future.


danielbachhuber:
@WP-API/amigos #reviewmerge

PR #2029, opened 2016-01-14
Move permission check methods in `WP_REST_Comments_Controller`

danielbachhuber:
Placing them above the method they're checking for makes the code easier
to read top to bottom.


danielbachhuber:
@WP-API/amigos #reviewmerge

PR #2030, opened 2016-01-14
Move permission check methods in `WP_REST_Controller`

danielbachhuber:
Placing them above the method they're supposed to check makes it move
obviously a pattern to follow.


danielbachhuber:
@WP-API/amigos #reviewmerge

PR #2038, opened 2016-01-14
Call `prepare_item_for_response()` directly in `Comment::create_item()`

danielbachhuber:
This lets us pass the original request through to
`prepare_item_for_response()`, giving it genuine context, and avoids an
unnecessary call to `get_item()`


danielbachhuber:
@joehoyle how do you feel about this pattern? Creating a new `WP_REST_Request` object always felt a bit hacky to me — I think this is a bit more true.

joehoyle:
Yeah I'm +1 for this, seems more lightweight and I don't see anything important skipped.

PR #2044, opened 2016-01-15
Let featured images forever be known as featured media

### LINKED ISSUE #2033: Rename `featured_image` attribute to `featured_media`
Our `_link` is `https://api.w.org/featuredmedia` intentionally.

danielbachhuber:
Why shouldn't you be able to feature other media too?

Fixes #2033

### DIFF
--- a/lib/endpoints/class-wp-rest-posts-controller.php
+++ b/lib/endpoints/class-wp-rest-posts-controller.php
@@ -283,8 +283,8 @@ class WP_REST_Posts_Controller extends WP_REST_Controller {
 			}
 		}
 
-		if ( ! empty( $schema['properties']['featured_image'] ) && isset( $request['featured_image'] ) ) {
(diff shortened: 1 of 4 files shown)


danielbachhuber:
@WP-API/amigos #reviewmerge

joehoyle:
My only concern here is that we are telling clients "this is a piece of media", so the client then needs to work out if it's some media it can display as an image. I know that's probably the case right now (technically) as I could have an audio object as the "feature image", but given it's stored in the db as Thumbnail ID, I don't know if there's a legitimate case for saying WordPress will support anything other than images in that (protected) meta key.

danielbachhuber:
Should we change the `_link` to `featuredimage` then?

danielbachhuber:
We're sticking with `featured_media`. Discussion in https://wordpress.slack.com/archives/core-restapi/p1453386374005100

PR #2048, opened 2016-01-15
First round at turning protected methods into private methods

danielbachhuber:
Our controller classes are designed to expose a consistent abstraction.
The expected public methods include `get_items()`, `get_item()`,
`create_item()`, `update_item()`, `delete_item()`,
`get_items_permissions_check()`, and so on.

Methods unique to a specific class should be marked `private` instead
of `protected`, so we don't expose them as interfaces for subclasses and
retain the freedom to refactor the internal behavior of a class at a
future date.

See #1063

### DIFF
--- a/lib/endpoints/class-wp-rest-attachments-controller.php
+++ b/lib/endpoints/class-wp-rest-attachments-controller.php
@@ -320,7 +320,7 @@ class WP_REST_Attachments_Controller extends WP_REST_Posts_Controller {
 	 * @param array $headers HTTP headers from the request
 	 * @return array|WP_Error Data from {@see wp_handle_sideload()}
 	 */
-	protected function upload_from_data( $data, $headers ) {
+	private function upload_from_data( $data, $headers ) {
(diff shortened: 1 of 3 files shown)


danielbachhuber:
@WP-API/amigos Is everyone alright with the this design decision? There's a fair bit more of the codebase needing to be changed up, but I wanted to submit this pull request first to get approval. If everyone is fine with it, #reviewmerge

joehoyle:
I'm not a fan of this, as it blocks subclassing something like the posts controller and adjusting methods where wanted. That might not be the intended design pattern, but I think it's still useful. I'm not sure if we want to get into the whole closed vs open by default debate, but I'm typically on the side of "if you subclass, do so at your own risk, but you can still hack things if you wish"

danielbachhuber:
Ok. Given @rmccue isn't a fan either, I think we need to revisit each protected method one by one to determine which should be made private (or removed entirely). We can do this in person next week.

PR #2053, opened 2016-01-19
Provide more helpful error message/code on Post Create/Update fail

danielbachhuber:
Instead of returning a generic `rest_forbidden` error, let's return an
error message indicating which operation failed, and 401 instead of 403
when the request is unauthenticated.

From https://core.trac.wordpress.org/ticket/35527

### DIFF
--- a/lib/endpoints/class-wp-rest-posts-controller.php
+++ b/lib/endpoints/class-wp-rest-posts-controller.php
@@ -240,7 +240,10 @@ class WP_REST_Posts_Controller extends WP_REST_Controller {
 			return new WP_Error( 'rest_cannot_assign_sticky', __( 'You do not have permission to make posts sticky.' ), array( 'status' => rest_authorization_required_code() ) );
 		}
 
-		return current_user_can( $post_type->cap->create_posts );
+		if ( ! current_user_can( $post_type->cap->create_posts ) ) {
+			return new WP_Error( 'rest_cannot_create', __( 'Sorry, you are not allowed to create new posts.' ), array( 'status' => rest_authorization_required_code() ) );
+		}
+		return true;
 	}
 
 	/**
@@ -328,7 +331,7 @@ class WP_REST_Posts_Controller extends WP_REST_Controller {
 		$post_type = get_post_type_object( $this->post_type );
 
 		if ( $post && ! $this->check_update_permission( $post ) ) {
-			return false;
+			return new WP_Error( 'rest_cannot_edit', __( 'Sorry, you are not allowed to update this post.' ), array( 'status' => rest_authorization_required_code() ) );;
 		}
 
 		if ( ! empty( $request['password'] ) && ! current_user_can( $post_type->cap->publish_posts ) ) {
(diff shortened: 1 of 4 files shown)


danielbachhuber:
@WP-API/amigos #reviewmerge

kadamwhite:
:+1: Not a PHP expert personally but this looks good, and is definitely an improvement over the current state of affairs.

kadamwhite:
Technically the [difference b/w 401 and 403](https://en.wikipedia.org/wiki/List_of_HTTP_status_codes) looks to be a matter of "is not authenticated" (401) vs "is authenticated but does not have credentials" (403). If we're returning a 403 for PUT errors based on capacity, is there previously a place where we checked whether the user is authenticated and returned a 401 if not?

danielbachhuber:
> If we're returning a 403 for PUT errors based on capacity, is there previously a place where we checked whether the user is authenticated and returned a 401 if not?

I'm not sure I follow. Can you clarify?

lib/endpoints/class-wp-rest-posts-controller.php
```diff
@@ -240,7 +240,10 @@ public function create_item_permissions_check( $request ) {
 			return new WP_Error( 'rest_cannot_assign_sticky', __( 'You do not have permission to make posts sticky.' ), array( 'status' => rest_authorization_required_code() ) );
 		}
 
-		return current_user_can( $post_type->cap->create_posts );
+		if ( ! current_user_can( $post_type->cap->create_posts ) ) {
+			return new WP_Error( 'rest_cannot_create', __( 'Sorry, you are not allowed to create new posts.' ), array( 'status' => rest_authorization_required_code() ) );
```
joehoyle, line 244: Should we be using the post type labels here? Trying to create a page will show "posts" right now

danielbachhuber, line 244: Yes, although all of our strings need to be updated to this effect, so I'd rather file it as a secondary issue.

danielbachhuber, line 244: #2056

kadamwhite:
> I'm not sure I follow. Can you clarify?

Sorry for the delayed response. What I mean was, if I interpret Wikipedia correctly ( && if Wikipedia represents proper usage), then both POST _and_ PUT should generate a 401 if the user is unauthenticated:

> 401 Unauthorized: Similar to _403 Forbidden_, but specifically for use when authentication is required and has failed or has not yet been provided. 401 semantically means "unauthenticated", i.e. "you don't have necessary credentials".

and then, once logged in, both POST _and_ PUT should generate a 403 if the user does not have the requisite permissions:

> 403 Forbidden: The request was a valid request, but the server is refusing to respond to it. Unlike a _401 Unauthorized_ response, authenticating will make no difference.[38] 403 error semantically means "unauthorized", i.e. "you don't have necessary permissions for the resource".

It looks (based purely on this PR) like we blanket-group POST errors as 401 and PUT as 403, which looks incorrect. I have not read through the rest of the code to determine whether this is the case, however.

kadamwhite:
...I was misreading the code completely. Please disregard the noise.

PR #2054, opened 2016-01-19
Use more correct error code for `Comment::delete` permission check

danielbachhuber:


danielbachhuber:
@WP-API/amigos #reviewmerge

PR #2057, opened 2016-01-20
Make `rest_post_trashable` dynamic based on the post type

danielbachhuber:
See #1167


danielbachhuber:
@WP-API/amigos #reviewmerge

PR #2058, opened 2016-01-20
Make `rest_insert_post`/`rest_delete_post` actions dynamic

danielbachhuber:
For instance, pages would fire `rest_insert_page`

See #1167


danielbachhuber:
@WP-API/amigos #reviewmerge

Issue #2001, opened 2016-01-12
Add `offset` parameter for all collections requests


danielbachhuber:
Each of these requests should support an `offset` parameter where the value is some integer:
- [x] `GET /wp/v2/posts`
- [x] `GET /wp/v2/comments`
- [x] `GET /wp/v2/users`
- [x] `GET /wp/v2/<taxonomy>`

Related #924, https://github.com/WP-API/WP-API/pull/1471#issuecomment-132215677

joehoyle:
+1 from me

danielbachhuber:
:ship:

PR #2061, opened 2016-01-20
Add `offset` param to `GET /wp/v2/posts`

danielbachhuber:
See #2001


danielbachhuber:
@WP-API/amigos #reviewmerge

PR #2062, opened 2016-01-20
Add `offset` param for `GET /wp/v2/tags`

danielbachhuber:
See #2001


danielbachhuber:
@WP-API/amigos #reviewmerge

PR #2063, opened 2016-01-20
Move term->post relations to the posts controller

### LINKED ISSUE #2032: Make assigning terms to a post less wonky
To assign category ID 9 to post ID 1, I currently need to:

```
POST /wp/v2/posts/1/categories/9
```

This seems a bit wonky. It's also not possible to control the order of the terms, or assign more than one term at a time. See #1421

I'm not sure why we created separate endpoints for term management in the first place (I didn't search past discussion). What if, instead, we dropped the Post Terms controller, and magically registered `categories`, `tags` and other custom taxonomy-based attributes as an array of ids to the Post schema? This would be more in line with `author`, `featured_image` (which are ids that represent other resources).

cc @joehoyle 

From https://github.com/WP-API/WP-API/issues/2016#issuecomment-171392733

joehoyle:
Fixes #2032


lib/endpoints/class-wp-rest-posts-controller.php
```diff
@@ -138,10 +138,10 @@ public function get_items( $request ) {
 
 		if ( $total_posts < 1 ) {
 			// Out-of-bounds, run the query again without LIMIT for total count
-			unset( $query_args['paged'] );
+		unset( $query_args['paged'] );
 			$count_query = new WP_Query();
-			$query_result = $count_query->query( $query_args );
-			$total_posts = $count_query->found_posts;
+		$query_result = $count_query->query( $query_args );
+		$total_posts = $count_query->found_posts;
```
danielbachhuber: Funky indentation?

joehoyle: Fixed in  ba5106d

lib/endpoints/class-wp-rest-posts-controller.php
```diff
@@ -1555,6 +1597,16 @@ public function get_item_schema() {
 			);
 		}
 
+		$taxonomies = wp_list_filter( get_object_taxonomies( $this->post_type, 'objects' ), array( 'show_in_rest' => true ) );
+		foreach ( $taxonomies as $taxonomy ) {
+			$base = ! empty( $taxonomy->rest_base ) ? $taxonomy->rest_base : $taxonomy->name;
+			$schema['properties'][ $base ] = array(
+				'description' => sprintf( __( 'The terms assigned to the objet in the %s taxonomy.' ), $taxonomy->name ),
```
danielbachhuber: 'objet' should be 'object'

danielbachhuber:
@joehoyle I took the liberty of renaming the methods to be more in line with the other methods we have on the class. PR looks good to me otherwise :+1:

PR #2069, opened 2016-01-22
When a comment is already trashed, return `410:rest_already_trashed`

danielbachhuber:
This is a more correct error than `500:rest_cannot_delete`


danielbachhuber:
@WP-API/amigos #reviewmerge

Issue #2065, opened 2016-01-21
Plans to reinstate filtering taxonomy term collections?


kadamwhite:
Way back in #401 we had added the ability to use `get_terms` arguments via the `filter` parameter on taxonomy term collections. Per #924 the beta plugins have removed that support: neither `/tags` nor `/categories` is listed as having any support for the `filter` query property, which has been borne out by my own testing on beta 10.

Lacking filter support for these collections makes it cumbersome for example to search for an individual term by slug: right now it appears my best approach is to `?search=the-term-slug`, then iterate through all returned posts until I find a match (for a generic slug `search` may bring many unrelated terms along for the ride).

I'd like to know whether there are any plans to reinstate this filtering ability, and if not, how querying for individual terms is likely to be handled in upcoming releases?

danielbachhuber:
> neither /tags nor /categories is listed as having any support for the filter query property, which has been borne out by my own testing on beta 10.

I've filed a separate issue for this, as it merits some discussion #2066

> Lacking filter support for these collections makes it cumbersome for example to search for an individual term by slug: 

Why don't we just add a `slug` param to the terms endpoint?

kadamwhite:
> Why don't we just add a slug param to the terms endpoint?

I believe a `slug` query is an important component of any slug-supporting collection, so if we went that route I'd suggest we add it to `/posts`, etc as well, for consistency.

danielbachhuber:
I've added `slug` to `/wp/v2/categories`, `/wp/v2/tags`, `/wp/v2/posts`, and `/wp/v2/users`

PR #2072, opened 2016-01-22
Add `slug` param for `GET /wp/v2/posts`

danielbachhuber:
Related https://github.com/WP-API/WP-API/issues/2065#issuecomment-173988524
See #924


danielbachhuber:
@WP-API/amigos #reviewmerge

kadamwhite:
What happens with `/posts?slug=oneslug&filter[name]=another`, I assume we just get an empty response? I'm AFK atm so I'm not sure what we did with e.g. paging filter parameters, whether those could continue to be used in `filter` once they made their way to a top-level parameter

danielbachhuber:
> What happens with /posts?slug=oneslug&filter[name]=another, I assume we just get an empty response? 

`filter[name]` would take priority over `slug`, so the query would use `another` as the slug.

Issue #2118, opened 2016-01-26
Introduce new WP_REST_Controller protected class variables


danielbachhuber:
To make it easier to subclass `WP_REST_Controller`'s children and reuse their methods, I propose we introduce four protected class variables:
- `$namespace`
- `$rest_base`
- <del>`$plural_label`</del>
- <del>`$singular_label`</del>

Edit: As it turns out, we can't use variables in the labels :(

Related #2056

We'll need to update:
- [x] `WP_REST_Attachments_Controller`
- [x] `WP_REST_Comments_Controller`
- [x] `WP_REST_Meta_Controller`
- [x] `WP_REST_Post_Statuses_Controller`
- [x] `WP_REST_Post_Types_Controller`
- [x] `WP_REST_Posts_Controller`
- [x] `WP_REST_Revisions_Controller`
- [x] `WP_REST_Taxonomies_Controller`
- [x] `WP_REST_Terms_Controller`
- [x] `WP_REST_Users_Controller`

TimothyBJacobs:
Posted this in a PR, but should've here.

Something to keep in mind, IIRC, using the singular and plural label to build strings is verboten. It doesn't work for languages with more complex sentence structures. See [this core ticket](https://core.trac.wordpress.org/ticket/12968).

danielbachhuber:
> Something to keep in mind, IIRC, using the singular and plural label to build strings is verboten. It doesn't work for languages with more complex sentence structures.

Ah, damn. Didn't think of that.

Can you think of a way we can accommodate controller-specific language? (#2056) Or, are we better off just using "Resource" ?

TimothyBJacobs:
I think it'd be best to handle it the way core currently does, which is default to using "View Post" and support customization in the `$labels` registration arg.

danielbachhuber:
Huh. We have a ton of potential strings though.

TimothyBJacobs:
Yep, languages suck. We could default to something like.

> You cannot delete posts of _post-slug_ type.

But I'm not sure that is much better.

danielbachhuber:
@TimothyBJacobs going with a more generic "resource" for now. Thanks for flagging the issue!

PR #2119, opened 2016-01-26
Use class variables for easier subclassing in Terms controller

danielbachhuber:
See #2118


danielbachhuber:
@WP-API/amigos #reviewmerge

Although this doesn't cover all of the strings in the controller, I want to make sure the pattern is acceptable before I move on to other controllers.

rmccue:
:+1: :100: 

I know Posts is one where I've hit the can't-override-namespace issue before, so good idea.

PR #2120, opened 2016-01-26
Clean up unneeded variable assignment in test controllers

danielbachhuber:


danielbachhuber:
Tests pass. Scrutinizer seems to be down.

Issue #1796, opened 2015-12-07
GET collection with invalid context produces nonsensical response


danielbachhuber:
I'd expect to get an error when I provide an invalid context, not this nonsensical response:

```
local ➜  ~  http GET http://wordpress-develop.dev/wp-json/wp/v2/posts\?context\=foo
HTTP/1.1 200 OK
[{"_links":{"self":[{"href":"http:\/\/wordpress-develop.dev\/wp-json\/wp\/v2\/posts\/1"}],"collection":[{"href":"http:\/\/wordpress-develop.dev\/wp-json\/wp\/v2\/posts"}],"author":[{"embeddable":true,"href":"http:\/\/wordpress-develop.dev\/wp-json\/wp\/v2\/users\/1"}],"replies":[{"embeddable":true,"href":"http:\/\/wordpress-develop.dev\/wp-json\/wp\/v2\/comments?post=1"}],"version-history":[{"href":"http:\/\/wordpress-develop.dev\/wp-json\/wp\/v2\/posts\/1\/revisions"}],"https:\/\/api.w.org\/featuredmedia":[{"embeddable":true,"href":"http:\/\/wordpress-develop.dev\/wp-json\/wp\/v2\/media\/5"}],"https:\/\/api.w.org\/attachment":[{"href":"http:\/\/wordpress-develop.dev\/wp-json\/wp\/v2\/media?parent=1"}],"https:\/\/api.w.org\/term":[{"taxonomy":"category","embeddable":true,"href":"http:\/\/wordpress-develop.dev\/wp-json\/wp\/v2\/posts\/1\/terms\/category"},{"taxonomy":"post_tag","embeddable":true,"href":"http:\/\/wordpress-develop.dev\/wp-json\/wp\/v2\/posts\/1\/terms\/tag"},{"taxonomy":"post_format","embeddable":true,"href":"http:\/\/wordpress-develop.dev\/wp-json\/wp\/v2\/posts\/1\/terms\/post_format"}],"https:\/\/api.w.org\/meta":[{"embeddable":true,"href":"http:\/\/wordpress-develop.dev\/wp-json\/wp\/v2\/posts\/1\/meta"}]}}]
```

danielbachhuber:
We should be validating `context` on collection requests ([discussion](https://wordpress.slack.com/archives/core-restapi/p1449702766001911))

danielbachhuber:
@joehoyle where do you think we should include this validation?

PR #2122, opened 2016-01-26
Introduce a new `WP_REST_Request_Controller`

danielbachhuber:
It contains a common set of methods for validating, sanitizing, and
transforming request values based on registered arguments for the
endpoint.

See #1796

### DIFF
--- a/lib/endpoints/class-wp-rest-comments-controller.php
+++ b/lib/endpoints/class-wp-rest-comments-controller.php
@@ -89,20 +89,13 @@ class WP_REST_Comments_Controller extends WP_REST_Controller {
 	 * @return WP_Error|WP_REST_Response
 	 */
 	public function get_items( $request ) {
-		$prepared_args = array(
-			'comment__in'     => $request['include'],
-			'comment__not_in' => $request['exclude'],
-			'number'          => $request['per_page'],
-			'post_id'         => $request['post'] ? $request['post'] : '',
-			'parent'          => isset( $request['parent'] ) ? $request['parent'] : '',
-			'search'          => $request['search'],
-			'offset'          => $request['offset'],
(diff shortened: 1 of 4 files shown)


lib/endpoints/class-wp-rest-comments-controller.php
```diff
@@ -89,20 +89,13 @@ public function get_items_permissions_check( $request ) {
...
-			'status'          => 'approve',
-			'type'            => 'comment',
-			'no_found_rows'   => false,
-		);
+		$prepared_args = WP_REST_Request_Controller::prepare_query_args_from_request( $request );
+		$prepared_args['status'] = 'approve';
+		$prepared_args['type'] = 'comment';
+		$prepared_args['no_found_rows'] = true;
+		$prepared_args['parent'] = ! empty( $prepared_args['parent'] ) ? $prepared_args['parent'] : '';
+		$prepared_args['post_id'] = ! empty( $prepared_args['post_id'] ) ? $prepared_args['post_id'] : '';
```
danielbachhuber, line 97: This doesn't actually need to happen, as we're properly setting a default value.

lib/endpoints/class-wp-rest-request-controller.php
```diff
@@ -0,0 +1,25 @@
+<?php
+
+class WP_REST_Request_Controller {
+
+	/**
+	 * Sanitizes, validates, and transforms WP_REST_Request query arguments.
```
danielbachhuber, line 6: Maybe we should move validation and sanitization to separate methods?

Actually, we could pluck `validate_schema_property` and `sanitize_schema_property` into this class, and then use them for collection queries or CRUD operations on every request.

danielbachhuber:
@WP-API/amigos Thoughts on this so far?

danielbachhuber:
Handled by subsequent PRs

Issue #2056, opened 2016-01-20
Make error strings and argument descriptions controller-specific


danielbachhuber:
Throughout the codebase, we have static strings like 'Sorry, you are not allowed to create new posts.' and "The date the object was published, in the site's timezone.".

Generally, strings should be dynamic based on the controller.

Related https://github.com/WP-API/WP-API/pull/2053#discussion_r50206683

rmccue:
Could we offload this to the post type registration `labels`?

danielbachhuber:
> Could we offload this to the post type registration labels?

The entire string, or just the source of the subject?

danielbachhuber:
I think it's a lot of new strings to try and include in `labels`, so we should use generic "resource" language instead.

ramiy:
@danielbachhuber using labels is a very good idea, but you need to make sure that it can be translated well to all the languages. In some edge cases using "plural" in English is not necessary sounds good in other languages.

PR #2133, opened 2016-01-28
Use class variables for easier subclassing in Statuses controller

danielbachhuber:
See #2118, #2056


TimothyBJacobs:
Something to keep in mind, IIRC, using the singular and plural label to build strings is verboten. It doesn't work for languages with more complex sentence structures. See [this core ticket](https://core.trac.wordpress.org/ticket/14827).

PR #2136, opened 2016-01-28
Use class variables for easier subclassing in Comments controller

danielbachhuber:
See #2118, #2056


danielbachhuber:
Closing - variables in the strings won't work, per https://github.com/WP-API/WP-API/issues/2118#issuecomment-176247044

It'll be easier to just use 'resource' in the string.

PR #2139, opened 2016-01-29
Revert `$singular_label` and `$plural_label`; use generic "Resource"

danielbachhuber:
As it turns out, we can't use label variables in strings. Instead, we'll
just use a more generic "Resource" throughout.

See #2118, #2056


danielbachhuber:
@WP-API/amigos #reviewmerge

PR #2140, opened 2016-01-29
Use more generic "Resource" language in attachment controller

danielbachhuber:
This permits easier subclassing without needing to modify strings.

See #2056


danielbachhuber:
@WP-API/amigos #reviewmerge

PR #2142, opened 2016-01-29
Use class variables for easier subclassing in Post(s) controllers

danielbachhuber:
- Removes `get_post_type_base()` method, which isn't necessary.
- Fixes `parent` link for revisions, which was previously missing `v2` from `wp/v2`

See #2118


danielbachhuber:
@WP-API/amigos #reviewmerge

Issue #1872, opened 2015-12-18
Respect show_avatars option in user response


tharsheblows:
In WP_REST_Users_Controller->prepare_item_for_response would it be ok to make $data['avatar_urls'] dependent on the show_avatars option chosen in the Discussions settings?

joehoyle:
I don't think this makes sense, the Discussion settings is a visual thing for show avatars, and we'd be better off keeping the information in the user object.

However, once we add the settings / options endpoint this value would be sent in there so you front end could chose to render the avatar based off that.

tharsheblows:
If a site owner thinks they've disabled avatars for whatever reason (my concern is privacy), they shouldn't be publicly available. Although looking at it, would it be ok to make the embed context of avatar_urls conditional on the show_avatars option? 

I do realise it's possible to filter them out but think the general assumption is that if avatars are disabled through the settings, they're not publicly accessible.

It's also entirely possible I'm misreading the embed context! I'm taking it as "data which is publicly available and of the sort one would use when embedding bits and bobs on other sites" :)

joehoyle:
Given the default WordPress avatar settings, I'm not convinced there is a privacy issue there, as they are just generated from known information already. However, I think it's worth discussion. I'm not sure the admin really portrays it as a user data/privacy type thing, it's more about the visual display of the comments list.

You have embed mostly right, but it's not necessarily just for public display, for example you can fetch a Private Post in the embed context, which only people with permission will be able to view.

tharsheblows:
Oh no, sorry, I didn't mean to have a discussion about whether avatars are bad or good, just that if I decide I don't want them to be shown for whatever reason, they shouldn't be available for others to use. 

I was surprised that they are available in the public user response even after unticking the show avatars box. In the themes I do for clients, I simply don't include any code to show avatars if it's not in the design --  I assume this is what everyone else does, too! So I've never used that box to toggle visual display but did assume that I could use it to remove avatar_url from the users response. I can't think of a single case where a client doesn't have avatars on the site but would want them or even not hate them in embedded content. Users on a community site I developed would find it particularly unnerving if an avatar they never knew was associated with that account suddenly showed up somewhere else with their forum post. 

I know there is some concern about data being unexpectedly public with the API and that you're aware of it and conscious of avoiding those situations - respecting the site owner's choice for avatar visibility is something similar.

[...]

PR #2151, opened 2016-01-29
Respect `show_avatars` option when adding avatars to Users

nullvariable:
If `show_avatars=>false`, then `avatar_url` shouldn't be included on the User Resource.

See #1872

### DIFF
--- a/lib/endpoints/class-wp-rest-users-controller.php
+++ b/lib/endpoints/class-wp-rest-users-controller.php
@@ -492,7 +492,6 @@ class WP_REST_Users_Controller extends WP_REST_Controller {
 			'url'                => $user->user_url,
 			'description'        => $user->description,
 			'link'               => get_author_posts_url( $user->ID ),
-			'avatar_urls'        => rest_get_avatar_urls( $user->user_email ),
 			'nickname'           => $user->nickname,
 			'slug'               => $user->user_nicename,
 			'registered_date'    => date( 'c', strtotime( $user->user_registered ) ),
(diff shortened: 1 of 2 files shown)


lib/endpoints/class-wp-rest-users-controller.php
```diff
@@ -501,6 +500,12 @@ public function prepare_item_for_response( $user, $request ) {
 			'extra_capabilities' => $user->caps,
 		);
 
+		$schema = $this->get_item_schema();
+
+		if ( ! empty( $schema['properties']['avatar_urls'] ) ) {
+				$data['avatar_urls'] = rest_get_avatar_urls($user->user_email);
```
danielbachhuber: Indentation is a bit off here. Also `rest_get_avatar_urls()` needs spaces around the parameter: `rest_get_avatar_urls( $user->user_email );`

tests/test-rest-users-controller.php
```diff
@@ -897,7 +897,6 @@ public function test_get_item_schema() {
 		$properties = $data['schema']['properties'];
 
 		$this->assertEquals( 17, count( $properties ) );
-		$this->assertArrayHasKey( 'avatar_urls', $properties );
```
danielbachhuber, line 900: Let's keep this test here, as `avatar_urls` should be present by default.

tests/test-rest-users-controller.php
```diff
@@ -914,6 +913,24 @@ public function test_get_item_schema() {
...
+
+		$this->assertArrayHasKey( 'avatar_urls', $properties );
+
+		update_option( 'show_avatars', false );
+		$request = new WP_REST_Request( 'OPTIONS', '/wp/v2/users' );
+		$response = $this->server->dispatch( $request );
+		$data = $response->get_data();
+		$properties = $data['schema']['properties'];
+
+		$this->assertArrayNotHasKey( 'avatar_urls', $properties );
```
danielbachhuber, line 927: Tests look good.

PR #2153, opened 2016-01-29
Use class variables for easier subclassing in Post Meta controller

danielbachhuber:
See #2118


danielbachhuber:
@WP-API/amigos #reviewmerge

PR #2163, opened 2016-01-30
Return original resource with `DELETE /wp/v2/comments/<id>`

danielbachhuber:
Now that all resources require the `force` param, we don't need to wrap
delete responses with the `trash` state.

See #1260


danielbachhuber:
@WP-API/amigos #reviewmerge

Issue #2066, opened 2016-01-21
`filter` param for Users, Comments and Terms


danielbachhuber:
We have a `filter` param on `GET /wp/v2/posts` which permits clients to use `WP_Query` arguments.

Should we introduce an equivalent param for Users, Comments and Terms?

If we do, we'll need a layer of authorization to use the `filter` param, as core currently doesn't have a paradigm of public vs. private query vars for non-Posts.

kadamwhite:
If we provide a `slug` query parameter across relevant collection types, looking at `get_terms` much of the remaining functionality is actually replicated by the existing parameters `page`, `per_page`, `order`, etc. The main pieces of term-specific functionality that isn't supported by the current parameters (per #924 at least) is searching by Slug, as noted in #2065; `child_of`, which extends beyond what the `parent` property can do; the `*__like` methods; and `exclude_tree`. These would make sense to expose via `filter`, but the functionality that's replicated by api-specific parameters wouldn't need to be exposed via the `filter` shim.

> we'll need a layer of authorization

I do not think this will be a good user experience; I either think we should whitelist some keys and provide them publicly with no auth needed, or not pursue the filter parameter at all. I believe API users will just be frustrated if they have to authenticate to make a `child_of` or `slug` query, without caring about the nuance of why that is the case.

danielbachhuber:
> > we'll need a layer of authorization
> 
> I do not think this will be a good user experience; I either think we should whitelist some keys and provide them publicly with no auth needed, or not pursue the filter parameter at all.

It's not really a user experience consideration — it's a "are we exposing data that was previously private" consideration.

kadamwhite:
I don't mean to discount that, but I'd rather a custom slug parameter that I can use than a filter parameter that requires authentication.

Additionally, most of the properties I described could be deduced as-is by iterating through all terms and computing the relationships yourself; that's just horribly inefficient.

patrickgalbraith:
Some reference for the built in queries and the equivalent API parameters currently.

## WP_Comment_Query

https://codex.wordpress.org/Class_Reference/WP_Comment_Query

| Key | API |
| --- | --- |
| author_email | author_email |
| author__in |  |
| author__not_in |  |
| include_unapproved |  |
| fields |  |
| ID | id |
| comment__in | include |
| comment__not_in |  |
| karma | karma |
| number |  |
| offset |  |
| orderby | orderby |
| order | order |
| parent | parent |
| post_author__in |  |
[...]

PR #2165, opened 2016-01-30
Remove `GET /wp/v2/comments` support for `post_*` params

### LINKED ISSUE #1979: Querying comments by post_slug is protected

danielbachhuber:
The proper pattern is to use `GET /wp/v2/posts` to fetch the post IDs to
limit the request to.

We may add support for these again in #2066, but that's a later discussion.

Fixes #1979

### DIFF
--- a/lib/endpoints/class-wp-rest-comments-controller.php
+++ b/lib/endpoints/class-wp-rest-comments-controller.php
@@ -120,11 +120,6 @@ class WP_REST_Comments_Controller extends WP_REST_Controller {
 				'type'         => isset( $request['type'] ) ? $request['type'] : '',
 				'author_email' => isset( $request['author_email'] ) ? $request['author_email'] : '',
 				'karma'        => isset( $request['karma'] ) ? $request['karma'] : '',
-				'post_author'  => isset( $request['post_author'] ) ? $request['post_author'] : '',
-				'post_name'    => isset( $request['post_slug'] ) ? $request['post_slug'] : '',
-				'post_parent'  => isset( $request['post_parent'] ) ? $request['post_parent'] : '',
-				'post_status'  => isset( $request['post_status'] ) ? $request['post_status'] : '',
-				'post_type'    => isset( $request['post_type'] ) ? $request['post_type'] : '',
 			);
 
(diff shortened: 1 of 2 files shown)


danielbachhuber:
@WP-API/amigos #reviewmerge

PR #2167, opened 2016-01-30
Link taxonomy terms to the post type collections they support

### LINKED ISSUE #898: Taxonomy terms should include _links to post type collections
If a taxonomy supports a given post type, we should add `_links` to the post type collections, filtered to the specific taxonomy term.

Here's an introduction to linking: http://v2.wp-api.org/extending/linking/

danielbachhuber:
Fixes #898

### DIFF
--- a/lib/endpoints/class-wp-rest-terms-controller.php
+++ b/lib/endpoints/class-wp-rest-terms-controller.php
@@ -510,21 +510,23 @@ class WP_REST_Terms_Controller extends WP_REST_Controller {
 		}
 
 		$taxonomy_obj = get_taxonomy( $term->taxonomy );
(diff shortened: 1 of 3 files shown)


lib/endpoints/class-wp-rest-terms-controller.php
```diff
@@ -509,6 +509,24 @@ protected function prepare_links( $term ) {
 			}
 		}
 
+		$taxonomy_obj = get_taxonomy( $term->taxonomy );
+		if ( ! empty( $taxonomy_obj->object_type ) ) {
```
rmccue: Conditional should be flipped to return early.

PR #2177, opened 2016-01-30
Use roles rather than role in the users controller

### LINKED ISSUE #1280: User should have a role, or roles - not both

joehoyle:
we want to use roles throughout rather than a split `role`/`roles`.

Fixes #1280


tripflex:
::edit::  roles must be submitted as form array, unless the `$request['roles']` is typecasted to an array see #2592

Issue #1845, opened 2015-12-11
Ensure consistent contexts across controllers


danielbachhuber:
Most endpoints default to `context=view`, but `GET /wp/v2/users/<id>` defaults to `embed`: 2d4cdbc0b145fbe2f06a8789d311dc844dcd5a72

Some endpoints only have `view`, the meta endpoint only has `edit`, and other endpoints have `edit`.

For a client trying to making requests, these inconsistencies are really confusing. We should make sure we have consistent default contexts across controllers, and each controller supports similar context names.

Related #1797, #1796

Rediscovered in #1838

Previously #702

danielbachhuber:
I put together a little script to audit our contexts: https://gist.github.com/danielbachhuber/cdc34b82b4a2815fb718

As of #1838, here's the result:

```
+---------------------------------------------------------------+-----------------+-------------------+
| route                                                         | default_context | enum              |
+---------------------------------------------------------------+-----------------+-------------------+
| /                                                             | view            |                   |
| /wp/v2                                                        | view            |                   |
| /wp/v2/posts                                                  | view            | view, embed, edit |
| /wp/v2/posts/(?P<id>[\d]+)                                    | view            | view, embed, edit |
| /wp/v2/posts/(?P<parent_id>[\d]+)/meta                        | edit            | edit              |
| /wp/v2/posts/(?P<parent_id>[\d]+)/meta/(?P<id>[\d]+)          | edit            | edit              |
| /wp/v2/posts/(?P<parent_id>[\d]+)/revisions                   | view            | view              |
| /wp/v2/posts/(?P<parent_id>[\d]+)/revisions/(?P<id>[\d]+)     | view            | view              |
| /wp/v2/posts/(?P<post_id>[\d]+)/categories                    | view            | view, embed       |
| /wp/v2/posts/(?P<post_id>[\d]+)/categories/(?P<term_id>[\d]+) | view            | view, embed       |
| /wp/v2/posts/(?P<post_id>[\d]+)/tags                          | view            | view, embed       |
| /wp/v2/posts/(?P<post_id>[\d]+)/tags/(?P<term_id>[\d]+)       | view            | view, embed       |
| /wp/v2/pages                                                  | view            | view, embed, edit |
| /wp/v2/pages/(?P<id>[\d]+)                                    | view            | view, embed, edit |
| /wp/v2/pages/(?P<parent_id>[\d]+)/meta                        | edit            | edit              |
[...]

PR #2184, opened 2016-01-30
Provide `embed` and `view` as public contexts for users

### LINKED ISSUE #1869: Don't forcibly downgrade context in WP_REST_Users_Controller::get_items()
In `WP_REST_Users_Controller::get_items()`, we forcibly downgrade context if the user doesn't have `list_users` capability:

```
if ( ! current_user_can( 'list_users' ) ) {
	$prepared_args['has_published_posts'] = true;

	// Only display a public subset of information
	$request['context'] = 'embed';
}
```

This is inconsistent with the rest of our controllers. If the current user can't access the supplied context, we should return an error.

Related #1845

danielbachhuber:
Details historically in `view` have been moved to `edit`

Fixes #1869
Related #1845

### DIFF
--- a/plugin.php
+++ b/plugin.php
@@ -292,9 +292,6 @@ if ( ! function_exists( 'rest_validate_request_arg' ) ) {
 			if ( 'string' === $args['type'] && ! in_array( $value, $args['enum'] ) ) {
 				return new WP_Error( 'rest_invalid_param', sprintf( __( '%s is not one of %s' ), $param, implode( ', ', $args['enum'] ) ) );
 			}
-			if ( 'array' === $args['type'] && array_diff( $value, $args['enum'] ) ) {
-				return new WP_Error( 'rest_invalid_param', sprintf( __( '%s includes a value not in %s' ), $param, implode( ', ', $args['enum'] ) ) );
-			}
 		}
 
 		if ( 'integer' === $args['type'] && ! is_numeric( $value ) ) {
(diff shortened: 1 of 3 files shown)


danielbachhuber:
@WP-API/amigos #reviewmerge

plugin.php
```diff
@@ -289,9 +289,12 @@ function rest_validate_request_arg( $value, $request, $param ) {
 		$args = $attributes['args'][ $param ];
 
 		if ( ! empty( $args['enum'] ) ) {
-			if ( ! in_array( $value, $args['enum'] ) ) {
+			if ( 'string' === $args['type'] && ! in_array( $value, $args['enum'] ) ) {
 				return new WP_Error( 'rest_invalid_param', sprintf( __( '%s is not one of %s' ), $param, implode( ', ', $args['enum'] ) ) );
 			}
+			if ( 'array' === $args['type'] && array_diff( $value, $args['enum'] ) ) {
```
joehoyle: Hmm this is somewhat new behaviour, as this is really "array of type", the valid values being the enums. We might be better of using the jsonSchema way of doing this, rather than inventing our own.

danielbachhuber: Oh. I think I introduced this because `roles` ended up with an enum in my merge conflict:

```
'roles'           => array(
    'description' => __( 'Roles assigned to the resource.' ),
    'type'        => 'array',
    'context'     => array( 'edit' ),
    'enum'        => array_keys( $wp_roles->role_objects ),
),
```

I'll revert for now.

danielbachhuber: 9fc8248

PR #2199, opened 2016-01-31
Fix array indentation in attachments controller

danielbachhuber:


rmccue:
FWIW in future, I'm happy for just a single PR for all style changes.

danielbachhuber:
> FWIW in future, I'm happy for just a single PR for all style changes.

Makes sense. Sometimes I intentionally create separate PRs so they're easier to read and/or to avoid potential merge conflicts.

PR #2200, opened 2016-01-31
Include or exclude multiple comment authors with `GET` Comments

### LINKED ISSUE #2191: Include and exclude multiple authors with `GET /wp/v2/comments`
Currently, `GET /wp/v2/comments` supports `author`, which is only one author.

We should have `author` support multiple author ids. Also, we should have `author_exclude`, which would map to `author__not_in`.

From #924

danielbachhuber:
Fixes #2191

### DIFF
--- a/lib/endpoints/class-wp-rest-comments-controller.php
+++ b/lib/endpoints/class-wp-rest-comments-controller.php
@@ -85,7 +85,7 @@ class WP_REST_Comments_Controller extends WP_REST_Controller {
 		}
 
 		if ( ! current_user_can( 'edit_posts' ) ) {
-			$protected_params = array( 'author', 'karma', 'author_email', 'type', 'status' );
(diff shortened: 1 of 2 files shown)


danielbachhuber:
@WP-API/amigos #reviewmerge

lib/endpoints/class-wp-rest-comments-controller.php
```diff
@@ -85,7 +85,7 @@ public function get_items_permissions_check( $request ) {
 		}
 
 		if ( ! current_user_can( 'edit_posts' ) ) {
-			$protected_params = array( 'author', 'karma', 'author_email', 'type', 'status' );
+			$protected_params = array( 'author', 'author_exclude', 'karma', 'author_email', 'type', 'status' );
```
joehoyle, line 88: Why is author protocted?

PR #2202, opened 2016-01-31
Include or exclude multiple authors with GET Posts

### LINKED ISSUE #2186: Include and exclude multiple authors with `GET /wp/v2/posts`
Currently, `GET /wp/v2/posts` supports `author`, which is only one author.

We should have `author` support multiple author ids. Also, we should have `author_exclude`, which would map to `author__not_in`.

From #924

danielbachhuber:
Fixes #2186


danielbachhuber:
@WP-API/amigos #reviewmerge

rmccue:
:ok_hand:

PR #2203, opened 2016-01-31
Introduce `embed` context for Types controller

danielbachhuber:
For better parity with other controllers, it makes sense to have a
`embed` context. For now, `embed` is the type name and slug.

See #1845


danielbachhuber:
@WP-API/amigos #reviewmerge

PR #2205, opened 2016-01-31
Introduce `embed` and `edit` contexts for Statuses controller

danielbachhuber:
This creates better parity between contexts for our controllers. `embed`
is simply the name and slug of the status. `edit` context now requires
authorization, which is consistent with our other controllers.

See #1845


danielbachhuber:
@WP-API/amigos #reviewmerge

PR #2209, opened 2016-02-01
Enforce min 1 and max 100 values for `per_page` arg

### LINKED ISSUE #1609: Limit number of requested items for collection queries

danielbachhuber:
Fixes #1609

### DIFF
--- a/plugin.php
+++ b/plugin.php
@@ -318,6 +318,12 @@ if ( ! function_exists( 'rest_validate_request_arg' ) ) {
 			}
 		}
 
+		if ( isset( $args['minimum'] ) && isset( $args['maximum'] ) ) {
+			if ( $value > $args['maximum'] || $value < $args['minimum'] ) {
+				return new WP_Error( 'rest_invalid_param', sprintf( __( '%s must be between %d and %d' ), $param, $args['minimum'], $args['maximum'] ) );
+			}
+		}
+
 		return true;
 	}
 }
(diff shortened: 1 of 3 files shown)


danielbachhuber:
@WP-API/amigos #reviewmerge

plugin.php
```diff
@@ -318,6 +318,12 @@ function rest_validate_request_arg( $value, $request, $param ) {
 			}
 		}
 
+		if ( isset( $args['minimum'] ) && isset( $args['maximum'] ) ) {
```
joehoyle: It feels like we are just inventing this adhoc? What does it mean to have `[ "type" => "string", "minimum" => 1 ]`? For now, I'd be more in favour of just validating these in their own callback rather than abstracting `rest_validate_request_arg` to handle it until we have a solid spec on how to achieve this.

danielbachhuber: > It feels like we are just inventing this adhoc?

Huh? `minimum` and `maximum` are a part of the spec:
- https://github.com/json-schema/json-schema/wiki/Minimum-and-exclusiveminimum
- https://github.com/json-schema/json-schema/wiki/Maximum-and-exclusivemaximum

Although, I'll include a type check for `integer` and `number`

PR #2211, opened 2016-02-01
Pass original $request context to `prepare_items_query`

### LINKED ISSUE #2207: Pass WP_REST_Request to prepare_items_query in WP_REST_Posts_Controller
Use case:

I'm extending the built-in `WP_REST_Posts_Controller` class to provide extra collection parameters / `WP_Query` arguments (sort by meta field, filter by P2P connected posts, etc.) for a custom post type endpoint.

AFAICT there is no way to do this without duplicating all of the code in `WP_REST_Posts_Controller::get_items()` as I need the current `WP_REST_Request` object to add the appropriate query arguments.

danielbachhuber:
When subclassing, this is useful context to have access to.

Fixes #2207


danielbachhuber:
@WP-API/amigos #reviewmerge

Issue #538, opened 2014-10-28
Non-Authenticated Create New User


royboy789:
Referencing to a conversation with @rmccue at WCSF.. creating an issue for the pull

al3x-edge:
@royboy789 @danielbachhuber Is this something that made it into 2.0 beta or did it die out?

danielbachhuber:
Hasn't been implemented yet.

rachelbaker:
@royboy789 Is this something you still want to submit a PR for using the `develop` branch?

tinytoolbox:
@royboy789  I can't find anything in the docs about this so I assume it's not in yet.  Any update?

danielbachhuber:
@tinytoolbox it's not yet supported by the API.

tinytoolbox:
Thanks @danielbachhuber.  I think I know the answer but Is there a timetable for its inclusion?  @royboy789 mentions in the thread above that he has some code for this.  Does it work?  My team would happily contribute to this as we need it for our project anyway.

danielbachhuber:
> I think I know the answer but Is there a timetable for its inclusion?

When someone creates a satisfactory pull request to address the issue :)

> mentions in the thread above that he has some code for this. Does it work?

Not sure. Given the age of the ticket, I suspect the code is quite out of date.

royboy789:
@tinytoolbox This is some old code from a plugin I created for social login/registration but it might help you.

You'd probably need to add authentication and validation, but it works..

tinytoolbox:
Was there supposed to be something attached?  I did get the code.  Thanks

tinytoolbox:
*didn't

royboy789:
@tinytoolbox sorry:
https://github.com/royboy789/wp-api-social-login/blob/master/inc/social-routes.php

simonegiacco:
This solves the issue using register_new_user() instead of wp_insert_user(), just in case of a non-admin trying to add a user without providing the password (non-admins can only create users without providing the password indeed)

The only counter effect i found is that register_new_user() returns errors (if any) in HTML, and this is not a standard you are following on the API.
The alternative is to copy-paste most of the register_new_user() code.

@danielbachhuber This pull request follows [my messages](https://wordpress.org/support/topic/i-need-to-allow-new-users-to-create-an-account?replies=4#post-7967628) on the plugin support page.

I'm happy to help, please give me some feedbacks to let my modification suit your needs.

Thanks

[...]

PR #2212, opened 2016-02-01
Non-Authenticated Create New User

simonegiacco:
Solves #538

### DIFF
--- a/lib/endpoints/class-wp-rest-users-controller.php
+++ b/lib/endpoints/class-wp-rest-users-controller.php
@@ -27,7 +27,7 @@ class WP_REST_Users_Controller extends WP_REST_Controller {
 				'permission_callback' => array( $this, 'create_item_permissions_check' ),
 				'args'            => array_merge( $this->get_endpoint_args_for_item_schema( WP_REST_Server::CREATABLE ), array(
 					'password'    => array(
-						'required' => true,
+						//'required' => true,
 					),
 				) ),
 			),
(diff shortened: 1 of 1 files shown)


lib/endpoints/class-wp-rest-users-controller.php
```diff
@@ -27,7 +27,7 @@ public function register_routes() {
 				'permission_callback' => array( $this, 'create_item_permissions_check' ),
 				'args'            => array_merge( $this->get_endpoint_args_for_item_schema( WP_REST_Server::CREATABLE ), array(
 					'password'    => array(
-						'required' => true,
+						//'required' => true,
 					),
```
danielbachhuber, line 31: We can remove this definition entirely.

lib/endpoints/class-wp-rest-users-controller.php
```diff
@@ -285,21 +288,68 @@ public function create_item( $request ) {
...
+						  *
+						  * @since 4.4.0
+						  *
+						  * @param int $user_id ID of the newly created user.
+						  */
+						do_action( 'network_user_new_created_user', $user_id );
+					}
+				}
+			}else{
+				$user_id = register_new_user( $user->user_login, $user->user_email );
```
danielbachhuber, line 343: We'll need to internalize the functionality for `register_new_user()`, as we won't be able to return HTML errors if registration fails.

danielbachhuber:
@simonegiacco Can you add a few to several new tests covering this functionality, and the new success and error cases it provides?

lib/endpoints/class-wp-rest-users-controller.php
simonegiacco, line 343: ...and I guess strip_tags() is not an option,right? :)

Ok, I'll internalize the function, who can give me the error messages?

danielbachhuber:
We're not going to be including this in the core plugin at this time, see https://github.com/WP-API/WP-API/pull/2356#issuecomment-195234273

Issue #2194, opened 2016-01-31
rest_validate_request_arg() should validate `type=array`


danielbachhuber:
Just like we have type checks for `string` and `integer`, we should also have a type check for `array`.

However, when we do this, we'll need to address instances where we've typed the param to `array`, but also accept a CSV of ids.

Lastly, the type checks in this function should happen before any other checks.

From 470e59ad0a2b60270008e267c8f981527b833c79

danielbachhuber:
@WP-API/amigos [Here's an example](https://wordpress.org/support/topic/creating-post-with-category-doesnt-work) in the support forums where the behavior of the API is confusing, because the client is passing data in the wrong format and we don't strictly validate

BE-Webdesign:
I'll get the PR for this in top shape tonight and after your feedback it will probably be good to go.

BE-Webdesign:
Well I have tried a bunch of different ways to do it and none of them really turned out well I will pass this on to someone else who has a better idea of how to do it.  I may come back to this at a later date, but it would be more productive for me to work on the documentation portions of things, for the time being.

PR #2213, opened 2016-02-01
Fully validate `minimum`, `maximum` and their exclusive attributes

danielbachhuber:
See #2194


danielbachhuber:
@WP-API/amigos #reviewmerge

danielbachhuber:
Alternatively, we could package this entire library: https://github.com/justinrainbow/json-schema

joehoyle:
Looks like this library is not 5.2 compatible.

PR #2214, opened 2016-02-01
Fix PHP strict standards notice

danielbachhuber:
Bug introduced in #2211


jonathanbardo:
@danielbachhuber Can you update the DocBlock with the missing params just to be consistent with `WP_REST_Posts_Controller`?

danielbachhuber:
> Can you update the DocBlock with the missing params just to be consistent

Good call, done.

PR #2216, opened 2016-02-01
Expose taxonomy and post type capabilities in `context==edit`

danielbachhuber:


danielbachhuber:
@WP-API/amigos #reviewmerge

TimothyBJacobs:
This should ship with a giant warning that just because a user has the capability described, does not mean they will be able to perform the action.

danielbachhuber:
> This should ship with a giant warning that just because a user has the capability described, does not mean they will be able to perform the action.

I think I understand what you're saying, but I'm not sure where it would fit in. How would you present such a warning?

nylen:
I think it makes sense to add a warning to the documentation, with an example of a case or two that can be problematic.

This change was prompted by some discussion in Slack:  https://wordpress.slack.com/archives/core-restapi/p1454367860000776

To summarize - we do a lot of capabilities checking in Calypso, mainly so that we can go ahead and disable UI elements and/or show messages if we are certain that the user _will not_ be able to perform an action.  This isn't perfect - as you said, we still need to be sure we handle errors correctly - but it's still very useful.

lib/endpoints/class-wp-rest-taxonomies-controller.php
```diff
@@ -154,6 +155,12 @@ public function get_item_schema() {
 			'title'                => 'taxonomy',
 			'type'                 => 'object',
 			'properties'           => array(
+				'capabilities'     => array(
+					'description'  => __( 'All capabilities used by the resource' ),
```
joehoyle, line 159: Missing full-stop on the sentence here.

lib/endpoints/class-wp-rest-post-types-controller.php
```diff
@@ -127,6 +128,12 @@ public function get_item_schema() {
 			'title'                => 'type',
 			'type'                 => 'object',
 			'properties'           => array(
+				'capabilities'     => array(
+					'description'  => __( 'All capabilities used by the resource' ),
+					'type'         => array(),
```
joehoyle, line 133: should `type` not be `"array"`?

lib/endpoints/class-wp-rest-taxonomies-controller.php
```diff
@@ -107,6 +107,7 @@ public function prepare_item_for_response( $taxonomy, $request ) {
 		$data = array(
 			'name'         => $taxonomy->label,
 			'slug'         => $taxonomy->name,
+			'capabilities' => $taxonomy->cap,
```
joehoyle, line 110: Does it make sense to white-list the capabilities? I'm wondering if in other versions of WordPress or plugins making modifications with new capabilities that we want consistent data coming from the API.

danielbachhuber, line 110: > Does it make sense to white-list the capabilities?

Hm. I'm not sure I understand the use case of whitelisting.

As far as the data itself goes, it's always expected to be an array of key => value pairs.

PR #2218, opened 2016-02-01
Error early when user can't GET types or taxonomies when `context=edit`

danielbachhuber:


danielbachhuber:
@WP-API/amigos #reviewmerge

PR #2219, opened 2016-02-02
Allow customisation on post controllers

tuanmh:
This would be great for someone who wants to control/modify post controller instances initialised by WP_API.

### DIFF
--- a/plugin.php
+++ b/plugin.php
@@ -163,6 +163,7 @@ if ( ! function_exists( 'create_initial_rest_routes' ) ) {
 			if ( ! is_subclass_of( $controller, 'WP_REST_Controller' ) ) {
 				continue;
 			}
+			$controller = apply_filters( "wp_rest_on_create_" . $post_type->name. "_controller", $controller, $post_type->name );
 
 			$controller->register_routes();
 
--- a/plugin.php
+++ b/plugin.php
@@ -163,7 +163,7 @@ if ( ! function_exists( 'create_initial_rest_routes' ) ) {
 			if ( ! is_subclass_of( $controller, 'WP_REST_Controller' ) ) {
 				continue;
 			}
-			$controller = apply_filters( "wp_rest_on_create_" . $post_type->name. "_controller", $controller, $post_type->name );
+			$controller = apply_filters( 'wp_rest_on_create_' . $post_type->name. '_controller', $controller, $post_type->name );
 
 			$controller->register_routes();
 


danielbachhuber:
Can you explain in greater detail what you're trying to do?

tuanmh:
> Can you explain in greater detail what you're trying to do?

Currently, there isn't a way for other plugins to control Post Controller created by WP-API. Someone could use this filter to keep the initialised instance and re-use himself instead of creating new instance of controllers every time he needs - he just wants a single instance of that class created during the API call. 

This can be beneficial as well in case we have some filters in a custom post controller - creating new instances of the class will end it up with many instances serving same filter unnecessary and would lead to unexpected results. 

Hope this clears your doubts :)

danielbachhuber:
> Currently, there isn't a way for other plugins to control Post Controller created by WP-API.

Actually, you can subclass the controller and use your subclass by overriding the global value:

```
function dbx_modify_posts_rest_api_controller() {
    global $wp_post_types;

    if ( isset( $wp_post_types['post'] ) ) {
        $wp_post_types['post']->rest_controller_class = 'My_Subclass_Of_Posts_Controller';
    }
}
add_action( 'init', 'dbx_modify_posts_rest_api_controller', 12 );
```

> Someone could use this filter to keep the initialised instance and re-use himself instead of creating new instance of controllers every time he needs - he just wants a single instance of that class created during the API call.

The controller classes aren't designed to keep their state in this way — you're safe to always re-initialize them. They also have actions and filters at key points where you might interpret the request or modify the response.

> This can be beneficial as well in case we have some filters in a custom post controller - creating new instances of the class will end it up with many instances serving same filter unnecessary and would lead to unexpected results.

I think you'd likely be much better off subclassing, as I've depicted above.

tuanmh:
> Actually, you can subclass the controller and use your subclass by overriding the global value:
> This can be beneficial as well in case we have some filters in a custom post controller - creating new instances of the class will end it up with many instances serving same filter unnecessary and would lead to unexpected results.
> I think you'd likely be much better off subclassing, as I've depicted above.

Yes, this is exactly what we're doing. That's why I mentioned `custom post controller`, it might not be clear to you. 

> The controller classes aren't designed to keep their state in this way — you're safe to always re-initialize them. They also have actions and filters at key points where you might interpret the request or modify the response.

I didn't mean the standard 'actions' and 'filters' offered by WP-API. There are always other custom filters and actions rather the ones from WP-API, re-initialising the sub-class means that there are multiple instances handling these filters and actions at the same time unnecessary. Of course, there are also ways to workaround but won't be a proper way to solve it. 

Again, there are at least 2 reasons for this PR:
- Allow dev to control the custom post instances created by WP-API.
- As subclassing post controller, new methods & properties can be added and if they're crucial for the controller to work - there isn't a way to set these properties & call the methods at your own interest. For example your custom post controller needs an interface to work and I need a way to inject the concrete class of that interface from outside. There's no way to do that now, unless you initialise that interface by using a concrete class and this is not the proper way. This includes the handling filters & actions as I said above. 

This would just enable WP-API to be more customisable and extendable.

PR #2230, opened 2016-02-04
Introduce `edit` context for Terms controller

danielbachhuber:
While `edit` doesn't include additional fields, it gives us a paradigm
for adding `edit`-specific data in the future, and makes Terms more
    consistent with other controllers.

See #1845


danielbachhuber:
@WP-API/amigos #reviewmerge

PR #2233, opened 2016-02-04
Use `parent` instead of `parent_id` in Revisions controller

danielbachhuber:
`parent` is consistent with the attribute, and our other controllers.


danielbachhuber:
@WP-API/amigos #reviewmerge

kadamwhite:
I see no cause for concern in the diff, but have not tested locally

PR #2234, opened 2016-02-04
Use the term cache for post data

### LINKED ISSUE #2232: Post prepare should use post-term cache
Related to #2228, we should be using `get_the_terms` rather than `wp_get_object_terms` directly. This is a no-brainer as we don't need a complex query, and significantly reduces queries on the posts collection endpoint.

rmccue:
Fixes #2232. Reduces number of queries on the posts collection by `posts_per_page * count( taxonomies )`


danielbachhuber:
:dancers:

PR #2235, opened 2016-02-04
Add `embed` and `edit` contexts to Revisions for parity

### LINKED ISSUE #1845: Ensure consistent contexts across controllers

danielbachhuber:
`edit` is the same as `view`. `embed` is a subset of `view`, following
what we've established for Posts

Fixes #1845


danielbachhuber:
@WP-API/amigos #reviewmerge

PR #2236, opened 2016-02-04
Remove unused variables in `Attachments::create_item()`

### LINKED ISSUE #2081: Unused variable assignment in `Attachments::create_item()`
Not sure what the original intent was with:

```
$name       = basename( $file['file'] );
$name_parts = pathinfo( $name );
$name       = trim( substr( $name, 0, -(1 + strlen( $name_parts['extension'] ) ) ) );
```

Scrutinizer also reports this as unused:

```
$img_url_basename = wp_basename( $data['source_url'] );
```

Discovered in #2080

wesm87:
Fixes #2081

### DIFF
--- a/lib/endpoints/class-wp-rest-attachments-controller.php
+++ b/lib/endpoints/class-wp-rest-attachments-controller.php
@@ -73,10 +73,6 @@ class WP_REST_Attachments_Controller extends WP_REST_Posts_Controller {
 			return $file;
 		}
 
-		$name       = basename( $file['file'] );
-		$name_parts = pathinfo( $name );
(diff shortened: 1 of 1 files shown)


danielbachhuber:
@wesm87 Can you look into what these variables were there for in the first place?

wesm87:
@danielbachhuber It looks like `$img_url_basename` was used in a block of code that was removed in 0f6986e7f5284af7a46448b3212b4f79348d65b2 The rest I'm not so sure about... the first occurrence I can find is in 7ae965bb9fb8f107b5c75215a6fb4e3311851a7f and as far as I can tell they've never actually been used anywhere.

danielbachhuber:
The first block actually represents a bug needing to be fixed: when creating an attachment, the default title name should be the attachment name.

PR #2241, opened 2016-02-08
Require minimum value of 1 for `page` param

### LINKED ISSUE #2124: Inconsistent results on /categories?page=0
When i query to mySite.com/wp-json/wp/v2/categories?page=0
I get the same results as page 2. 
Does it make sense?
If the parameter is index 1 based, shouldn't it be ignored?

Thanks for your time.

Btw: this took me some headaches until i realized it

danielbachhuber:
Fixes #2124


danielbachhuber:
@WP-API/amigos #reviewmerge

kadamwhite:
:+1:

PR #2242, opened 2016-02-08
Respect `<!--nextpage-->` flag in post content

### LINKED ISSUE #1206: Posts with pages issue with rendered content
This all refers to the posts controller.

If a post utilises the nextpage separator to paginate the post then currently you can't retrieve the content for those individual pages within the post using ?page=2. There's a couple of problems: When setup_postdata is used in the prepare_item_for_response function, the $wp_query global is not used to store the query, so some of the globals that setup_postdata sets such as $page are not set correctly (because it relies on retrieving the query vars) so the rendered content does not display properly.

Secondly, the rendered content doesn't use get_the_content to render it, so it will just render the full content of the post, rather then just the page you want.

The other issue is in the get_items function the page arg is unset and paged is used instead, but page is required as a query var to get the right page in the post.

danielbachhuber:
When present, `content.rendered` includes the first page, and `content.pages` is all remaining pages.

Fixes #1206

### DIFF
--- a/tests/test-rest-posts-controller.php
+++ b/tests/test-rest-posts-controller.php
@@ -570,6 +570,23 @@ class WP_Test_REST_Posts_Controller extends WP_Test_REST_Post_Type_Controller_Te
 		$this->assertEquals( 403, $response->get_status() );
 	}
 
+	public function test_get_item_paged_content() {
+		$post_content = <<<EOT
+This is my first page of content.
+<!--nextpage-->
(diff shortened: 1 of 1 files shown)


danielbachhuber:
Wontfix. I actually think this is a bad idea, and a responsibility of the presentational layer https://github.com/WP-API/WP-API/issues/1206#issuecomment-181421197

PR #2246, opened 2016-02-09
Clean up attachment controller

### LINKED ISSUE #2081: Unused variable assignment in `Attachments::create_item()`

JPry:
A few updates to the Attachments controller:
- Remove unused variables
- Update docblocks to use `stdClass` instead of `object` where appropriate.

Fixes #2081.

### DIFF
--- a/lib/endpoints/class-wp-rest-attachments-controller.php
+++ b/lib/endpoints/class-wp-rest-attachments-controller.php
@@ -83,13 +83,9 @@ class WP_REST_Attachments_Controller extends WP_REST_Posts_Controller {
 			return $file;
 		}
 
-		$name       = basename( $file['file'] );
-		$name_parts = pathinfo( $name );
-		$name       = trim( substr( $name, 0, -(1 + strlen( $name_parts['extension'] ) ) ) );
-
-		$url     = $file['url'];
-		$type    = $file['type'];
-		$file    = $file['file'];
+		$url  = $file['url'];
+		$type = $file['type'];
+		$file = $file['file'];
 
 		// use image exif/iptc data for title and caption defaults if possible
 		// @codingStandardsIgnoreStart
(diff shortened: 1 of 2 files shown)


danielbachhuber:
Thanks @JPry. I'll land this first thing for Beta 13

lib/endpoints/class-wp-rest-attachments-controller.php
```diff
@@ -83,13 +83,9 @@ public function create_item( $request ) {
 			return $file;
 		}
 
-		$name       = basename( $file['file'] );
-		$name_parts = pathinfo( $name );
-		$name       = trim( substr( $name, 0, -(1 + strlen( $name_parts['extension'] ) ) ) );
```
danielbachhuber, line 88: From the historical record, this block was used in v1 to set a default `$title` ([ref](https://github.com/WP-API/WP-API/blob/1f85ba4bf409d1b531fdc19d7bdfbcb7c1717aeb/lib/class-wp-json-media.php#L229-L231)):

```
$name       = basename( $file['file'] );
$name_parts = pathinfo( $name );
$name       = trim( substr( $name, 0, -(1 + strlen($name_parts['extension'])) ) );
$url     = $file['url'];
$type    = $file['type'];
$file    = $file['file'];
$title   = $name;
```

It doesn't appear we have a default `$request['title']` when empty now.

This code was originally from https://github.com/WP-API/WP-API/commit/7ae965bb9fb8f107b5c75215a6fb4e3311851a7f#diff-1dd6759fb788e439e3b8c9bd01884447R131

danielbachhuber, line 88: Bug that needs to be fixed.

danielbachhuber:
@JPry I've cherry-picked your doc block fixes to another PR, and will handle the bug fix in a separate PR

Issue #1749, opened 2015-11-18
Design of autosaves


joehoyle:
I was recently working on something that wanted to make use of autosaving, as this exists in the WordPress admin, it makes sense to be able to something similar via the API.

I had to dig through how WordPress handles autosaves in the database, I've specced that our below at a high level:

- Autosaves to draft posts are just straight updates to the post, pretty basic
- Autosaves to published posts is more complex:
  - A revision is created for each use that edits a published post on "autosave" happening
  - A post has a "new autosave" if it has a revision who's post_date is later then the post_parent's publish date
  - Autosave revisions are identified by a special post_name
  - Choosing to restore from autosave essentially does a restore to revision on the autosave revision

In terms of how to handle this in the API, I'm not totally sure - if we allowed creating of revisions (we currently don't) it would be possible for a client to implement the above rules/flow, however I'd say that's very convoluted and it would be nice if we avoid that.

We may want to scale back the supported features for autosaves, e.g. having autosaves for each user is added complexity.

The case of draft posts maybe should be handled differently, as WP makes a different case for that (it's basically just saving the post every 30s). We may want to look at other APIs here for prior art too.

danielbachhuber:
Because post types can conditionally support revisions, we can't treat autosaves as future revisions.

Instead, we'll need to create `/wp/v2/posts/<id>/autosaves` for a client to create and get autosaves.

danielbachhuber:
@joehoyle moving this to beta 13

ellatrix:
I found myself in need of this too.

PR #2253, opened 2016-02-09
Autosave endpoint for posts

joehoyle:
See #1749

This is an initial stab at a way for clients to push autosave / partial data to existing posts. This is largerly modelled of the Jetpack / WordPress.com Autosave endpoint.

This is currently partial, and largely untested. This is a direct way of solving this problem, more RESTful, novel approach suggestions welcome :)


kadamwhite:
I'm 👎  on this for now, it's an interesting approach but autosave feels like something we should address more mindfully unless this is necessary for 4.7, which I do not believe it is.

joehoyle:
Ok, I'm going to close this out as I don't think this approach will be taken in the longer term.

PR #2256, opened 2016-02-10
Ensure search is an empty string

### LINKED ISSUE #2255: Comments collection accidentally generates a search query
We default `search` to `null`, but internally, `WP_Comment_Query` checks that it's an empty string. This generates a search query which looks like:

``` sql
SELECT SQL_CALC_FOUND_ROWS sz_comments.comment_ID
FROM sz_comments
WHERE ( comment_approved = '1' )
AND comment_post_ID IN ( 85 )
AND comment_type IN ('')
AND (comment_author LIKE '%%'
OR comment_author_email LIKE '%%'
OR comment_author_url LIKE '%%'
OR comment_author_IP LIKE '%%'
OR comment_content LIKE '%%')
ORDER BY sz_comments.comment_date_gmt ASC, sz_comments.comment_ID ASC
LIMIT 10
```

:hear_no_evil:

Core needs to be patched to have a better check, but we should work around this in the meantime.

cc @rachelbaker our comment overlord

rmccue:
See #2255; if it's an empty value (like null, our default) but not an empty string, this will generate a useless search query.

Fixes #2255 for the API, core should be patched too.

### DIFF
--- a/lib/endpoints/class-wp-rest-comments-controller.php
+++ b/lib/endpoints/class-wp-rest-comments-controller.php
@@ -138,6 +138,9 @@ class WP_REST_Comments_Controller extends WP_REST_Controller {
 		if ( empty( $request['offset'] ) ) {
 			$prepared_args['offset'] = $prepared_args['number'] * ( absint( $request['page'] ) - 1 );
 		}
+		if ( empty( $request['search'] ) ) {
+			$prepared_args['search'] = '';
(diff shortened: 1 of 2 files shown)


rmccue:
@WP-API/amigos #reviewmerge

danielbachhuber:
Adding a test case.

PR #2257, opened 2016-02-10
Use cached get_the_terms() for terms-for-post

### LINKED ISSUE #2228: Terms get_items should check post cache
Rather than calling `wp_get_object_terms`, we should use `get_the_terms`, or check `get_object_term_cache` ourselves. This will require doing the sorting ourselves though; we may be able to optimise for the common use case of the defaults.

rmccue:
When hitting `/wp/v2/categories?post={id}`, we can use `get_the_terms()` instead of `get_terms()` to get the cached data.

Fixes #2228.


rmccue:
@WP-API/amigos #reviewmerge

Bonus points if you find anything I missed. :)

danielbachhuber:
@rmccue What about more test cases?

PR #2261, opened 2016-02-10
Use our Spy REST Server in tests to call `rest_post_dispatch` filter

### LINKED ISSUE #2260: Posting a comment with post id throws an error
I've been experimenting with versions of the REST API for quite a while now and have recently updated to the latest beta12 (0549f8e). My application is posting comments with the `post` field set to the id of the post as per the documentation. The request payload contains the following:

```
{
  author: 1,
  author_name: "Commenter",
  content: "My comment"
  post: 1
}
```

Currently, this throws an error in the `WP_REST_Comments_Controller` on [line 73](https://github.com/WP-API/WP-API/blob/2.0-beta12/lib/endpoints/class-wp-rest-comments-controller.php#L73) after saving the comment in the database and getting the updated collection of comments. It appears the request object is checked for the post field and handled as an array, rather than an integer. The behaviour was introduced in [this commit](14ab402) to allow for multiple post ids to be checked when obtaining a collection of comments.

Is the documentation outdated in this regard, am I using the endpoint incorrectly or have I hit upon a bug? Thanks!

danielbachhuber:
Also fixes up errors caused in `rest_post_dispatch` call.

Fixes #2260

### DIFF
--- a/lib/endpoints/class-wp-rest-comments-controller.php
+++ b/lib/endpoints/class-wp-rest-comments-controller.php
@@ -70,7 +70,7 @@ class WP_REST_Comments_Controller extends WP_REST_Controller {
 	public function get_items_permissions_check( $request ) {
 
 		if ( ! empty( $request['post'] ) ) {
-			foreach ( $request['post'] as $post_id ) {
+			foreach ( (array) $request['post'] as $post_id ) {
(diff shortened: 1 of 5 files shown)


danielbachhuber:
@WP-API/amigos #reviewmerge

Issue #2114, opened 2016-01-25
Should callbacks distinguish between URL and body params?


mineralwassermedium:
Hello

When I make a POST request to the endpoint /users/<id> to update data of a user, I can update any user. For example: 

I have two users: 

``` javascript
{id: 1, slug: 'user1'}
{id: 2, slug: 'user2'}
```

Now I am logged in as user1 and make the post request to the endpoint /users/1 with the request body:

``` javascript
{id: 2, slug: 'user3', first_name: 'Jon'}
```

and the former user2 is now

``` javascript
{id: 2, slug: 'user3', first_name: 'Jon'}
```

Is that supposed to work?

danielbachhuber:
Internally, our callbacks don't currently distinguish between arguments included in the URL and arguments included in the body. Because you've included `id` in the body, it overrides the `id` specified by the named group regex for the path.

I'll flag this issue for discussion as to whether we should make our internal callbacks more precise. Thanks for reporting!

danielbachhuber:
After some discussion, our callbacks should check the 'id' in the URL to be the most precise.

However, we need to change a lot of references to the `$request['id']`, and change many of our tests, in order to handle this.

danielbachhuber:
#2260 is another reason we need to update all of our controllers:
- [x] Attachments
- [ ] Comments
- [ ] Statuses
- [ ] Types
- [ ] Taxonomies
- [ ] Terms
- [ ] Users

BE-Webdesign:
For the update_item and update_item_permissions_check should the $id = (int) $request['id']; be changed to $id = (int) $request->get_url_params()['id'];  If that's is the case I can help out with these and can do that on the other controllers that would need this distinction.  If I misunderstand the problem let me know.  Thank you.

danielbachhuber:
I appreciate the offer, @BE-Webdesign. There's a lot of code churn in these changes though, so I think it would be most efficient for me to just sweep through and make all of the changes.

I'll set aside some time in the next couple of days to mark some more [good first bugs](https://github.com/WP-API/WP-API/issues?q=is%3Aopen+is%3Aissue+label%3A%22Good+First+Bug%22) if you're looking for issues to dive in on.

BE-Webdesign:
Okay no problem @danielbachhuber.  I will get on the good first bugs that already exist in the meantime. If you have the time to answer, was I at least on the right track with the solution? I would like to know whether I am understanding the implementation of the API better or if I am completely not understanding it.

[...]

PR #2262, opened 2016-02-10
Distinguish between url, body, and query params in Attachments

danielbachhuber:
We need to be precise to ensure our callbacks are handling the correct
request data.

See #2114
Related #2260


danielbachhuber:
@WP-API/amigos #reviewmerge

Issue #2144, opened 2016-01-29
Support for date queries with `GET /wp/v2/posts`


danielbachhuber:
It would be nice if we had top-level support for date queries with `GET /wp/v2/posts`.

`WP_Query` currently supports a ton of date filters:

![image](https://cloud.githubusercontent.com/assets/36432/12675466/5a190996-c63f-11e5-9969-df25e278a38b.png)

We should pick which fields would be sensible to support, and add them to `GET /wp/v2/posts`

Related #924

kadamwhite:
Anecdotally, I have most often leveraged `year`, `month`, `day`, `before` and `after`.

bjork:
To me it would make sense to expose the whole `date_query` but not the shorthand (root level) parameters to avoid having too many ways to do the same thing. The only disadvantage is that encoding arrays within arrays as HTTP query parameters is not the most intuitive thing and is specific to PHP. But, expressing complex structures as a single string is always going to be at least somewhat complex.

rmccue:
Agreed with @bjork, although I'd also be happy to support both `before` and `after` as top-level and support ISO8601 dates for them. You can implement year/month/date/etc on top of this pretty easily, i.e. `year => 2015` maps directly to `before=2016-01-01T00:00:00Z&after=2015-01-01T00:00:00Z`.

bponghneng:
Wow, literally just registered at wordpress.org then logged in here to ask whether this is supported, or am I missing something in the WP_REST_Post_Types_Controller class. Glad to see it's on the list for coming soon. Queries using before and after would do very nicely, IMO.

Looks fairly trivial, but I just took a whack at it and couldn't get it to work in an hour of flogging. I did manage to get `year` and `monthnum` queries working with half a dozen edited lines of code in a subclass of WP_REST_Post_Types_Controller. So I think I'm missing something simple. My first guess is the missing piece is somewhere in REST API allowed query vars since `before` and `after` aren't in the WP allowed vars.

Anyway, I'd be delighted to contribute in this small area if someone wouldn't mind fielding some questions from a WP newbie.

wesm87:
@bpongvh I just started looking at the `tax_query` issue and I think you're on the right track, you'll probably want to start by adding `before` and `after` to the `$rest_valid` array in `WP_REST_Posts_Controller::get_allowed_query_vars()` and then look at `WP_REST_Posts_Controller::get_items()`. I could be way off track, though.

bponghneng:
[...]

PR #2266, opened 2016-02-11
Add date query to posts endpoints

BE-Webdesign:
Currently I exposed everything for date parameters.  Whether we want
just date query or just top level or make date query params top level
and then parse a date query, is up for discussion. I am up for improving
this, this is a start and something that works.

**If you want to use the top level just pass them as arguments.**
GET wp/v2/posts/?year=2016&monthnum=1

For posts from January, 2016.

**If you want to use date query do something like:**
GET /wp/v2/posts/?date_query[0][after][year]=2001&date_query[0][before][2020]

This will return posts after the year 2001 and before 2020.


BE-Webdesign:
This is for #2144, new to GitHub should have referenced that above.

joehoyle:
This is looking like a good first step @BE-Webdesign, I think we'll likely want to look at the param names, for example I think we're unlikely to match the wp_query names like `w` as those are very cryptic. I seem to remember discussing with @rmccue and @danielbachhuber briefly as to whether it would be best to _just_ support `date_query` rather than all the other shorthand ones - however we should probably talk about this.

Code wise, this is very neat and clear - I'd typically steer away from fixing whitespace / yoda conditions at the same time though, unless it's that piece of code you are changing.

This is a good PoC and we should look to iterate on it after discussing at the next meeting in slack :+1:

BE-Webdesign:
@joehoyle As my schedule is right now I should be able to attend the meeting and am up for helping out with any changes the team needs me to make.  Thanks, for the tips I will not change the yoda conditions, makes sense.  I did that in my other tax_query support so I will remove that at some point today in both of them.

rmccue:
Per https://github.com/WP-API/WP-API/issues/2144#issuecomment-179703893 I think we're best off only supporting `before` and `after` at the top level, and supporting `filter[date_query]` separately. The other query vars are a bit complex and inconsistent, and 80% of expected behaviour can be supported with just `before` and `after`.

rmccue:
(Agreed with @joehoyle on the code quality though, this is a super neat PR :+1: )

BE-Webdesign:
@rachelbaker , @joehoyle, @rmccue, @danielbachhuber , is this latest revision on the right track?

lib/endpoints/class-wp-rest-posts-controller.php
```diff
@@ -1581,6 +1592,11 @@ public function get_collection_params() {
 
 		$params['context']['default'] = 'view';
 
+		$params['after'] = array(
+			'description'       => __( 'The after parameter in a date query, send as an ISO8601 or strtotime() compliant date string. See wp-includes/date.php' ),
```
danielbachhuber: Accepting `strtotime()`-compatible values seems a bit sketchy to me, given we don't accept them anywhere else in the API and they're very much a PHP-ism.

If we change `format=>date-time`, then we get validation for ISO8601 for free and don't have to introduce a validation callback.

danielbachhuber:
> is this latest revision on the right track?

I think so, aside from my one comment.

lib/endpoints/class-wp-rest-posts-controller.php
BE-Webdesign: @danielbachhuber - Excellent suggestion, will get on that.

BE-Webdesign:
@danielbachhuber Let me know if I need to tweak this anymore.  Thanks for the feedback! I am learning a lot faster so big thanks to all of the WP-API team for helping me out!

danielbachhuber:
> Let me know if I need to tweak this anymore.

Can you add some tests covering both valid and invalid date parameters supplied?

BE-Webdesign:
> Can you add some tests covering both valid and invalid date parameters supplied?

I have to learn how to do that first, but I believe I can get that done.

BE-Webdesign:
@danielbachhuber I think this is ready to go, maybe?  Before and after would need to be added to posts, pages, and media on #924.

lib/endpoints/class-wp-rest-posts-controller.php
```diff
@@ -102,6 +102,16 @@ public function get_items( $request ) {
 		$args['post_status']          = $request['status'];
 		$args['s']                    = $request['search'];
 
+		// Set before into date query. Date query must be specified as an array of an array.
+		if ( isset( $request['before'] ) ) {
+			$args['date_query'][0]['before'] = $request['before'];
```
danielbachhuber, line 108: To prevent a PHP notice, can we create the `date_query` array before trying to assign the value?

lib/endpoints/class-wp-rest-posts-controller.php
```diff
@@ -1581,6 +1592,12 @@ public function get_collection_params() {
 
 		$params['context']['default'] = 'view';
 
+		$params['after'] = array(
+			'description'        => __( 'The after parameter in a date query, send as an ISO8601 compliant date string.' ),
```
danielbachhuber: Clients don't necessarily know what a "date query" is, given this is a WordPress-ism. Also, "after parameter" and "send" are implied in the context. What about something like:

"Limit response to resources published after a given ISO8601 compliant date."

danielbachhuber:
> I think this is ready to go, maybe?

Tests look pretty good. I left a few small follow-up comments to address.

> Before and after would need to be added to posts, pages, and media on #924.

Yep, I've got a script to update that.

BE-Webdesign:
Okay, thanks for the comments.  I will make those changes now!

PR #2268, opened 2016-02-11
Update Categories tests to properly set url, body, and query params

danielbachhuber:
See #2114


danielbachhuber:
Backing out of this decision.

Issue #2248, opened 2016-02-09
Respect `show_avatars` option when adding avatars to comments


danielbachhuber:
Needs to be the same behavior as #2151

danielbachhuber:
:hammer:

Issue #1488, opened 2015-08-20
Use short CURIEs instead of full URIs


rmccue:
One small annoyance in the API right now is that link relations are full URIs. In JavaScript, this means that while you have  `_links.self` and `_links.author`, you have to do `_links["http://v2.wp-api.org/term"]`. This is a bit ugly, and can be confusing.

We should instead switch to [CURIEs](https://en.wikipedia.org/wiki/CURIE). These are an abbreviated form of the full URI that's easier to work with, but can be expanded out to the full URI if needed. They look like this:

```json
{
    "_links": {
        "curies": [
            {
                "name": "wp",
                "href": "http://v2.wp-api.org/{rel}",
                "templated": true
            }
        ],

        "wp:term": [
            {
                "taxonomy": "category",
                "href": "..."
            },
            {
                "taxonomy": "post_tag",
                "href": "..."
            }
        ]
    }
}
```

malfborger:
+1 !

scottopolis:
+1

danielbachhuber:
Closing in favor of https://core.trac.wordpress.org/ticket/34729

danielbachhuber:
Actually, I just realized this will only affect our endpoints.

danielbachhuber:
We should handle this magically in the server: https://core.trac.wordpress.org/ticket/34729

danielbachhuber:
Re-opening, because we need to add a shim to the plugin when the server code lands.

danielbachhuber:
@rmccue @joehoyle thoughts on how we can shim this into the plugin?

joehoyle:
So I've been looking into doing this, we don't have good hooks in place to modify the response links, best I can see is to either run a subclass of the server class in the plugin, or filter the server response and look for links recursively (bit of a hack).

danielbachhuber:
> best I can see is to either run a subclass of the server class in the plugin

I think subclassing seems like the cleanest implementation at this point

PR #2271, opened 2016-02-11
Respect-`show_avatars`-comments-#2248

BE-Webdesign:
Possible fix for #2248.  Respects the show_avatars option when
retrieving avatars data for comments. Modeled after #2151.

### DIFF
--- a/lib/endpoints/class-wp-rest-comments-controller.php
+++ b/lib/endpoints/class-wp-rest-comments-controller.php
@@ -886,7 +886,7 @@ class WP_REST_Comments_Controller extends WP_REST_Controller {
 			),
 		);
 
-		if ( true === get_option( 'show_avatars' ) ) {
+		if ( '1' === get_option( 'show_avatars' ) ) {
 			$avatar_properties = array();
 
 			$avatar_sizes = rest_get_avatar_sizes();
--- a/lib/endpoints/class-wp-rest-comments-controller.php
+++ b/lib/endpoints/class-wp-rest-comments-controller.php
@@ -557,7 +557,7 @@ class WP_REST_Comments_Controller extends WP_REST_Controller {
 
 		$schema = $this->get_item_schema();
 
-		if ( ! empty( $schema['properties']['avatar_urls'] ) ) {
+		if ( ! empty( $schema['properties']['author_avatar_urls'] ) ) {
 			$data['author_avatar_urls'] = rest_get_avatar_urls( $comment->comment_author_email );
 		}
 
@@ -886,7 +886,7 @@ class WP_REST_Comments_Controller extends WP_REST_Controller {
 			),
 		);
 
-		if ( get_option( 'show_avatars' ) ) {
+		if ( 1 === get_option( 'show_avatars' ) ) {
 			$avatar_properties = array();
 
 			$avatar_sizes = rest_get_avatar_sizes();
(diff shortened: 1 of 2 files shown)


lib/endpoints/class-wp-rest-comments-controller.php
```diff
@@ -899,6 +885,29 @@ public function get_item_schema() {
 				),
 			),
 		);
+
+		if ( get_option( 'show_avatars' ) ) {
```
rmccue, line 892: This would be better off as a strict check (`=== true`) if we can.

rmccue:
Looks pretty good :+1: Small change in the check, but otherwise seems pretty sensible.

lib/endpoints/class-wp-rest-comments-controller.php
BE-Webdesign, line 892: Okay, will fix.  In the codex it is stored as 1 and 0 integer type.  Is this changed to a boolean or should I check for 1 === get_option( 'show_avatars' )?

BE-Webdesign:
Hallelujah!

lib/endpoints/class-wp-rest-comments-controller.php
```diff
@@ -899,6 +885,29 @@ public function get_item_schema() {
 				),
 			),
 		);
+
+		if ( '1' === get_option( 'show_avatars' ) ) {
```
joehoyle: I think this should be a `if ( get_option...` the strict type check, while correct for most cases, is somewhat unintuitive, as this _really_ should be a bool. +1 for allowing any truthy value.

joehoyle:
Should we also apply the same to the users controller? The database option name is `show_avatars` which makes it seem like it's not comment specific, though I understand that's where the option is in the dashbaord?

BE-Webdesign:
@joehoyle I believe this was already done in users #2151.  @rmccue suggested the strict check so if == is the way over === we can decide that in the meeting maybe?

danielbachhuber:
I'm :-1: on a strict check in this context. A filter from a plugin or theme should be able to return a truthy or falsy value, and see the API work as expected.

BE-Webdesign:
@danielbachhuber So do you believe we should put it back to:

if ( get_option( 'show_avatars' ) )
**or use**
if (  true == get_option( 'show_avatars' ) )
**or something else entirely?**

danielbachhuber:
I think we should use `if ( get_option( 'show_avatars' ) )`

rmccue:
I'm OK with the truthy/loose check so long as it doesn't cause side effects (i.e. we don't accidentally catch `'n'` or something).

danielbachhuber:
> I'm OK with the truthy/loose check so long as it doesn't cause side effects

Core uses `if ( ! get_option( 'show_avatars' ) )`: https://core.trac.wordpress.org/browser/trunk/src/wp-includes/pluggable.php#L2356

I think we're fine using `if ( get_option( 'show_avatars' ) )`. @BE-Webdesign if you want to make that final change, we can get this landed.

BE-Webdesign:
Okay, I will change it back give me a second.

danielbachhuber:
Build fails because of https://github.com/WP-API/WP-API/issues/1488#issuecomment-184664633

Merging because curies need to be fixed in this release regardless.

Issue #2145, opened 2016-01-29
Support taxonomy queries


danielbachhuber:
It would be nice if we had top-level support for tax queries with `GET /wp/v2/posts`.

Here's `WP_Query` currently supports:

![image](https://cloud.githubusercontent.com/assets/36432/12675613/5f311076-c640-11e5-958a-7079d19e8f1b.png)

Related #924

bjork:
See [my comment on #2144](https://github.com/WP-API/WP-API/issues/2144#issuecomment-176990148). Like `date_query`, exposing `tax_query` would be very straightforward.

kadamwhite:
Re-purposing this issue to serve as a general TODO list for taxonomy handling, because the current interface is inconsistent at best and unworkable in many situations.

I took a crack at coming up with some stories for taxonomy querying in general: **As an API consumer, I...**
- Should be able to request records that are associated with a specific taxonomy term
- Should be able to request records that are associated with multiple terms of a single taxonomy (AND)
- Should be able to request records that are associated with any of several terms of a single taxonomy (OR)
- Should be able to request records that are associated with specific terms of multiple taxonomies
- Should be able to request records that are associated with multiple terms of multiple taxonomies (AND)
- Should be able to request records that are associated with any of several terms of multiple taxonomies (OR)
- Should be able to request records that are NOT associated with a specific taxonomy term
- Should be able to request records that are NOT associated with multiple terms of a single taxonomy
- Should be able to request records that are NOT associated with any of several terms of multiple taxonomies
- Should be able to request records that are NOT associated with any of several terms of multiple taxonomies
- Should get an empty collection back if no records match the provided conditions

(This list assumes the queried taxonomies are marked to be shown in the REST API.)

I would like to see the taxonomy query interface hit the above points (or at least the first four!) in as terse a query syntax as is feasible.

kadamwhite:
I also believe we should strive for consistency and make this apply to CPTs and any resource for which a taxonomy is registered, not just posts.

kadamwhite:
An assumption of the above list is that the interface would look the same or similar regardless of the taxonomy. So these would satisfy:

```
?terms[]=catID1&terms[]=tagID&terms[]=catID2
?tax[category][]=catID1&tax[category][]=catID2&tax[tag][]=tagID
```

but these would not:

```
?filter[tag]=tagSlug&filter[cat]=catID&tax[custom_tax]=termId
```

kadamwhite:
[...]

PR #2273, opened 2016-02-13
Add tax query #2145

BE-Webdesign:
Adds tax query support to GET wp/v2/posts with custom validation function for #2145.  Let me know what we can do to improve this.

**Simple call for posts in the uncategorized category taxonomy.**
/wp/v2/posts?tax_query[0][taxonomy]=category&tax_query[0][field]=slug&tax_query[0][terms]=uncategorized

**More Complex:**
/wp/v2/posts?tax_query[relation]=OR&tax_query[0][taxonomy]=category&tax_query[0][field]=slug&tax_query[0][terms][0]=quotes&tax_query[1][relation]=AND&tax_query[1][0][taxonomy]=post_format&tax_query[1][0][field]=slug&tax_query[1][0][terms][0]=post-format-quote&tax_query[1][1][taxonomy]=category&tax_query[1][1][field]=slug&tax_query[1][1][terms][0]=wisdom


BE-Webdesign:
Test failing because tax_query is potentially a slow query so is it not a good idea to support this? Need assistance!  Is my validation what is throwing that error?

joehoyle:
This goes somewhat hand-in-hand with date query, I'm not sure whether we want to match the same naming as WP internally for the fields, or if we should support the full tax query. 

A first iteration may be good to allow `tax_query` as part of the filter param?

BE-Webdesign:
So at this point it is probably ridiculous to merge this one line of code in.  So @joehoyle if you want to add this into #2287 it would probably make sense to do it there instead of 8 commits for one line of code haha!  If I am missing something that needs to be done with tax_query as a filter param let me know.  Currently it works well with something like:

GET wp/v2/posts/?filter[tax_query][0][taxonomy]=category&filter[tax_query][0][field]=slug&filter[tax_query][0][terms]=uncategorized

I think this combined with Joe's #2287 should solve #2145 and #2277 and add some nice parameters to handle taxonomy querying.

joehoyle:
I think we'll still want this PR, as it adds it to filter. I didn't use this technique in #2287, as we need to providing whitelisting for the taxonomies in `filter[tax_query]` which I believe we don't currently have. @BE-Webdesign I thought you had done something like that already when you make `tax_query` a top level param?

BE-Webdesign:
@joehoyle I had a validation function that validated the tax_query itself which did include white listing minus the show_in_rest => true.  I will try to figure out a way to add that in.  Do we want that in depth of validation of the filter param if tax_query is set or do we just want to white list taxonomies that are show_in_rest => true?

BE-Webdesign:
Definitely will need some feedback on the direction for this latest commit.  I tested it with #2287 and there are no conflicts.  So if you wanted to find posts that were both in tags 1 and categories 1 then with #2287 you could do something like this.

**GET wp/v2/posts/?tags=1&categories=1&filter[tax_query][relation]=AND**

I think this in combination with @joehoyle 's work will make handling more robust tax queries a lot simpler.  Feedback is always appreciated!

danielbachhuber:
> Definitely will need some feedback on the direction for this latest commit.

Holy moly. I need to think on this one for a little bit.

BE-Webdesign:
@danielbachhuber I spoke briefly with @rachelbaker one night about validation of the filter param.  We both kind of came to the conclusion that it would be overkill.  So we can decide whether we want to selectively validate certain arguments passed into the filter or whether we want none at all. I think validating all of them would just be too much code and as new WP_Query args are added it would make maintaining the REST API more difficult as well.

I'm pretty sure WP_Query() validates the arguments passed to it so the validation only really serves as a way to tell the user they made a bad request, which might be important, might not.  As we have more top level params, that cover a lot of use cases, I imagine the filter param will be used less and probably by power users who will understand WP_Query so validation may not be necessary at all.  @joehoyle , @rmccue , @rachelbaker , @danielbachhuber thoughts would be appreciated!

joehoyle:
@BE-Webdesign the problem with not sanitizing / filtering the tax_query is we would be potentially allowing taxonomies to be queried that have specifically not been opted into the REST API via `show_in_rest`. This is why we can't rely on WP_Query to do this, as that will allow any taxonomy that has a query var to be used.

I think the sanitize function for tax_query shouldn't be too difficult when it gets down to it. Especially if we on'y check the taxonomy names, rather than a full sanitization of all values.

BE-Webdesign:
@joehoyle Okay so cut this down to just checking for valid taxonomy names which will be queried against for show_in_rest => true?  Will other filter params be validate/sanitized at some point as well?

lib/endpoints/class-wp-rest-posts-controller.php
```diff
@@ -1724,4 +1726,138 @@ public function validate_user_can_query_private_statuses( $value, $request, $par
...
+						return $is_valid;
+					}
+					continue;
+				default :
+					// Returns true for validation by default.
+					$is_valid = true;
+					continue;
+			}
+		}
+		return $is_valid;
```
kjbenk, line 1817: If there are no `$values` then would't this code throw an error since `$is_valid` is not declared anywhere?  I would just `return true` here if you want that to be the default value, or declare `$is_valid = true` before the foreach.

BE-Webdesign, line 1817: Ya I should put an empty check there and have some declaration of $is_valid, most likely this is being tossed anyways for something simpler.  This is more of an experiment rather than what is being merged.  But thank you for pointing that out because that is a glaring mistake haha :smile:

BE-Webdesign:
Going to simplify this some more.

BE-Webdesign:
Failing for use of tax_query potentially being a slow query.

kjbenk:
@BE-Webdesign I know it does not solve the problem but you can add this [comment](https://github.com/WordPress-Coding-Standards/WordPress-Coding-Standards/wiki/Whitelisting-code-which-flags-errors#slow-taxonomy-queries) to all the places you use `tax_query` in your tests.
- https://github.com/WP-API/WP-API/pull/2273/files#diff-dea402c127c0b4c164758886c78a4b09R485
- https://github.com/WP-API/WP-API/pull/2273/files#diff-dea402c127c0b4c164758886c78a4b09R505
- https://github.com/WP-API/WP-API/pull/2273/files#diff-dea402c127c0b4c164758886c78a4b09R520

BE-Webdesign:
@kjbenk Thanks for the tip!!! Going to put it in.

tests/test-rest-posts-controller.php
```diff
@@ -482,6 +482,7 @@ public function test_get_items_valid_tax_query_in_filter() {
 
 		$request = new WP_REST_Request( 'GET', '/wp/v2/posts' );
 		$request->set_param( 'filter', array(
+			// WPCS: tax_query ok.
 			'tax_query' => array(
```
kjbenk: You need to add the comments inline.  So this would be the correct syntax `'tax_query' => array( // WPCS: tax_query ok.` or at least I think :)

lib/endpoints/class-wp-rest-posts-controller.php
```diff
@@ -1784,4 +1786,86 @@ public function validate_user_can_query_private_statuses( $value, $request, $par
...
+				switch ( $filter_param ) {
+					case 'tax_query' :
+						$is_valid = $this->rest_validate_tax_query( $value, $request, $filter_param );
+						// If an error is found exit early and it will throw an error for request->has_valid_params();
+						if ( is_wp_error( $is_valid ) ) {
+							return $is_valid;
+						}
+						continue;
+					default :
+						continue;
```
kadamwhite, line 1813: Nitpick, but for a single-case `switch` I'd vote in favor of an `if`. (unless this is a code standard of which I was not aware)

BE-Webdesign: Yeah it was originally an entire validation of tax_query, but now it handles only one case.  If this ever wanted to be expanded though it should go back to switch.  I can make a modification to it sometime soon.

BE-Webdesign:
Closed in favor of #2748.

PR #2281, opened 2016-02-16
Add testing for cached get_the_terms() call

### LINKED ISSUE #2228: Terms get_items should check post cache

rmccue:
Following on from #2257, we need unit tests, since there's a bug. :|

Fixes #2228.


rmccue:
Tests failing because WP nightly build is broken.

danielbachhuber:
Build fails because of https://github.com/WP-API/WP-API/issues/1488#issuecomment-184664633

Merging because curies need to be fixed in this release regardless.

Issue #2277, opened 2016-02-14
Post from category giving all posts


kdpuvvadi:
I'm trying to get the posts from specific category, when i try to use `https://www.example.com/wp-json/wp/v2/posts?categories=38` it is giving all posts rather than posts from that specific category.

LukePettway:
Try doing:

https://www.example.com/wp-json/wp/v2/posts/38

Luciaisacomputer:
Try doing the following

https://www.example.com/wp-json/wp/v2/posts?filter[cat]=38

kdpuvvadi:
it works but why doesn't `posts?categories="category-id"` work even though it is provided by WP REST API and `posts?tags="tag-id"` is also not working

Luciaisacomputer:
Could you post a link to where you are seeing that? I don't think I've done it that way before so I am curious.

kdpuvvadi:
https://www.telugushowtime.com/wp-json/wp/v2/posts?tags=165

rmccue:
@kdpuvvadi Neither of these work; I'm not sure what indicated that they should?

You need to use `filter[cat]=...` and `filter[tag_id]=...` per the [WP Query](https://codex.wordpress.org/Class_Reference/WP_Query#Category_Parameters) documentation.

kdpuvvadi:
i've seen them in the category and tag details api, and for category `filter[cat]=...`  is working but 'filter[tag_id]=...' is giving all posts even though the results not relevant.

joehoyle:
I think we should add this, I'll take it on

kdpuvvadi:
I've given my site url above, Please check with it. Thank you.

danielbachhuber:
Landed in #2287

kdpuvvadi:
So, how can i filter posts with tag?

PR #2287, opened 2016-02-17
Support taxonomy collection args on posts endpoint.

joehoyle:
See #2277


lib/endpoints/class-wp-rest-posts-controller.php
```diff
@@ -1695,6 +1709,18 @@ public function get_collection_params() {
 		$params['filter'] = array(
 			'description'       => __( 'Use WP Query arguments to modify the response; private query vars require appropriate authorization.' ),
 		);
+
+		$taxonomies = wp_list_filter( get_object_taxonomies( $this->post_type, 'objects' ), array( 'show_in_rest' => true ) );
+		foreach ( $taxonomies as $taxonomy ) {
+			$base = ! empty( $taxonomy->rest_base ) ? $taxonomy->rest_base : $taxonomy->name;
+
+			$params[ $base ] = array(
+				'description'       => sprintf( _( 'Limit result set to all items that have the specified term assigned in the %s taxonomy.' ), $base ),
```
danielbachhuber: `__()` instead of `_()`

danielbachhuber:
Should we error if an invalid term id is provided?

joehoyle:
I think not, all the checking for a large amount of terms would be slow IMO.

danielbachhuber:
Ok, just one nit then

joehoyle:
:+1: fixed in 6c1a4d1

kadamwhite:
This is just by ID, correct—so `tags=tagslug` wouldn't work? Fine if so, just making sure I understand the code

kadamwhite:
> Should we error if an invalid term id is provided?

I would expect to just get an empty collection back

kadamwhite:
Final Q: does this give any ability to query by multiple categories?

danielbachhuber:
> This is just by ID, correct

Correct

> Final Q: does this give any ability to query by multiple categories?

Yes

danielbachhuber:
> Final Q: does this give any ability to query by multiple categories?

Although, I believe this is `AND` and not `OR`

BE-Webdesign:
When the filter[tax_query] is in you can specify that in addition to get either `AND` or `OR`. I also believe the default is OR but could be wrong don't remember. https://github.com/WP-API/WP-API/pull/2273

kadamwhite:
> > Final Q: does this give any ability to query by multiple categories?
> 
> Yes

Can you demonstrate what that query would look like?

PR #2288, opened 2016-02-18
Remove post meta link from post response.

joehoyle:
Now we no longer have meta in the core plugin.


rachelbaker:
Removed the related unit test that also needed to be removed.  Ready for Review.

danielbachhuber:
:rabbit:

PR #2291, opened 2016-02-19
Add date query for comments

BE-Webdesign:
This should add before and after parameters to the top level of query arguments for the comments endpoint.  Similar to #2266 and the same functionality pretty much that goes for #2144 using posts.  Use ISO8601 compliant date time strings!

Example:

**GET wp/v2/comments/?after=2015-01-01T00:00:00Z&before=2017-01-01T00:00:00Z**
For comments between January 1st, 2015 and January 1st, 2017.

Let me know what improvements can be made!


danielbachhuber:
:ship: :ship: :ship:

PR #2295, opened 2016-02-20
Subclass `WP_REST_Server` to apply curies to WP 4.4 responses

### LINKED ISSUE #1488: Use short CURIEs instead of full URIs

danielbachhuber:
Sadly, a filter would've made this much easier, but the filter we need
doesn't exist.

Fixes #1488


danielbachhuber:
@WP-API/amigos #reviewmerge

danielbachhuber:
There's another test failing because curie transformation happens in `WP_REST_Server::get_response_links`, which we never call inside of our `WP_Test_Spy_REST_Server::dispatch()` method. Open to ideas on how to inject that.

danielbachhuber:
Closing in favor of #2412

PR #2298, opened 2016-02-21
Add rendered and raw to attachment's caption

miya0001:
HTML is allowed in attachment's caption.
Do they need `raw` and `rendered` like `the_content`?

https://github.com/WordPress/WordPress/blob/master/wp-includes/media.php#L1725


danielbachhuber:
I don't think we need to break `caption` into `rendered` and `raw`. That use of `texturize()` is specific to the `[gallery]` shortcode.

miya0001:
Thanks @danielbachhuber 

I thought it will make need to break schema If specifications would be changed in the WordPress core (e.g. add filter hook), because html is  allowed in here like `the_content`.

But I also think you are right too. :smile:

PR #2303, opened 2016-02-23
Fix slug and date validation for new, unpublished posts

adamsilverstein:
When trying to save/update unpublished posts with the current validation in place, I am getting an error:

> "Invalid parameter(s): date_gmt (The date you provided is invalid.), slug (slug is not of type string)"

screenshot

In my testing, newly created posts (via api or wp-admin) get a `post_date_gmt` of `0000-00-00 00:00:00` - the js client can't parse that as a date and stores '`null`. Sending this value pack to the API via a PUT request results in a `rest_invalid_date` error before this patch. Checking for $value truthy before attempting to parse the date fixes the issue.

Newly created, unpublished posts have a `post_name` aka `slug` that is empty. The API returns this as an empty string (`"slug": ""`) - sending an empty string back to the API before this PR results in the `rest_invalid_param` error.  I tracked this down to applying `sanitize_title` to the slug, this is somehow turning the empty string into an empty object. Changing the sanitization callback to `sanitize_key` fixes this issue and I believe thats the appropriate sanitization call to use here in any case. 

also ensure the empty keys are not used in `handle_terms`


BE-Webdesign:
I think in the handle_terms() you want to leave it as it was and drop the `
|| empty( $request[ $base ] )` So that you could pass in empty arrays for tags or categories.

BE-Webdesign:
For the conditional to array_map in handle_terms() maybe change the conditional to be

``` php
if ( is_array( $request[ $base ] ) ) {
    $terms = array_map( 'absint', $request[ $base ] );
}
```

adamsilverstein:
@BE-Webdesign good suggestion, that fixes the issue for me and the failing test now passes. The one test that is failing agains trunk is also failing in develop: its not related to this fix.

lib/endpoints/class-wp-rest-posts-controller.php
```diff
@@ -1367,7 +1370,7 @@ public function get_item_schema() {
 					'type'        => 'string',
 					'context'     => array( 'view', 'edit', 'embed' ),
 					'arg_options' => array(
-						'sanitize_callback' => 'sanitize_title',
+						'sanitize_callback' => 'sanitize_key',
```
danielbachhuber, line 1373: We need to use `sanitize_title()` because core uses `sanitize_title()` for the slug, not `sanitize_key()`. See #1585

danielbachhuber:
@adamsilverstein Can you make sure there's test coverage for the issues you've found?

plugin.php
```diff
@@ -306,7 +306,7 @@ function rest_validate_request_arg( $value, $request, $param ) {
 		if ( isset( $args['format'] ) ) {
 			switch ( $args['format'] ) {
 				case 'date-time' :
-					if ( ! rest_parse_date( $value ) ) {
+					if ( $value && ! rest_parse_date( $value ) ) {
```
danielbachhuber, line 309: I don't think we should necessarily only permit null values for dates, but not for other types or formats.

@WP-API/amigos Should we give `null` a free pass in validation (e.g. `null` is always an acceptable value for a field)?

lib/endpoints/class-wp-rest-posts-controller.php
adamsilverstein, line 1373: Ok, I will try to dig in to see why this change resolves the issue I am seeing.

plugin.php
adamsilverstein, line 309: Thanks for the feedback: note that what I am trying to do is update a post that is new, and unpublished. I've tested with posts created in wp-admin or via the api. new, unpublished posts have '0000-00-00 00:00:00' for their date_gmt and blank slugs. Updates to the endpoint as an update with these values fails validation. 

perhaps the api should just not return a slug and date_gmt for new unpublished posts? i'm going to try adjusting the client to skip sending these values to see if that resolves my issue.

danielbachhuber, line 309: > new, unpublished posts have '0000-00-00 00:00:00' for their gmt_date

Oh, huh. Maybe we should be special casing these, and returning `null`.

adamsilverstein:
@danielbachhuber for now I added a patch to the client to have it skip sending the blank slug and null date_gmt when saving the post: https://github.com/WP-API/client-js/commit/8ed4be654f25daf6fd491944b07664d153bf8b34 

I think instead of changing the validation, we could change the return: and not return the null/blank values. Basically: if the values are not valid, the api should not return them.

plugin.php
rmccue, line 309: @danielbachhuber I'd be in favour of treating `null` as the value not being included. Maybe we should do an `array_filter` to exclude null before we validate?

Also, would this break things if we handled it that way?

danielbachhuber, line 309: > Also, would this break things if we handled it that way?

Potentially. Someone should look at existing validation libraries to see how null cases are handled.

adamsilverstein, line 309: The backbone client now includes a workaround for this issue.

rachelbaker:
> I think instead of changing the validation, we could change the return: and not return the null/blank values. Basically: if the values are not valid, the api should not return them.

@adamsilverstein did you want to update this PR, or open a new one?

adamsilverstein:
@rachelbaker Sure, I will work something up.

adamsilverstein:
@rachelbaker I created a PR with my alternate proposed solution: https://github.com/WP-API/WP-API/pull/2393 

I'm not sure what the best solution is, I this this is right - that the API should omit empty values in the return data; I'm curious to hear what you think or if this creates any issues I'm missing.

joehoyle:
I created https://github.com/WP-API/WP-API/issues/2657 to make a decision on `null` handling.

kadamwhite:
Decision in #2657 was that values should be sent as `null`, not omitted; #2696 captures implementing that change

adamsilverstein:
Created a new PR for this: https://github.com/WP-API/WP-API/pull/2763. Closing.

PR #2311, opened 2016-02-24
Remove unused code block in terms controller

danielbachhuber:


rmccue:
Oooooooooops! This needs reverting; the first `if` block there needs to check this variable, as if `$use_cache` is false, we need to do the full `get_terms` call.

PR #2317, opened 2016-02-24
Add timezone to date and date_gmt

miya0001:
If there isn't timezone, `new date( date_gmt )` always returns `date_gmt` as local time.

Before:

``` javascript
new date( '2016-02-24T11:45:36' ); // always returns same value at anywhere.
```

After:

``` javascript
new date( '2016-02-24T11:45:36+00:00' ); // returns local time.
```

### DIFF
--- a/lib/endpoints/class-wp-rest-posts-controller.php
+++ b/lib/endpoints/class-wp-rest-posts-controller.php
@@ -664,7 +664,12 @@ class WP_REST_Posts_Controller extends WP_REST_Controller {
 	protected function prepare_date_response( $date_gmt, $date = null ) {
 		// Use the date if passed.
 		if ( isset( $date ) ) {
-			return mysql_to_rfc3339( $date );
+			if ( get_option( 'gmt_offset' ) < 0 ) {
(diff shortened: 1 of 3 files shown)


rmccue:
IIRC this was actually an intentional removal, but then Chrome changed how they parse datetimes in `new Date()`. I'd recommend solving it on the client-side by doing:

``` js
new Date( date_gmt + 'Z' );
```

I can't find the original ticket/discussion about this though, so leaving open for now.

miya0001:
@rmccue Thanks! :smile:

rachelbaker:
Related 6d0ad766ca525c7a3aee921b66554ddf21a5c023

BE-Webdesign:
Closing this PR out as dates are this seemed to be deemed a won't change due to the weirdness that is dates in every language.  The API supports ISO 8601 strings and was changed to strip out time zones in [6d0ad76](https://github.com/WP-API/WP-API/commit/6d0ad766ca525c7a3aee921b66554ddf21a5c023).

PR #2318, opened 2016-02-25
Revert "Remove unused code block in terms controller"

rmccue:
Reverts WP-API/WP-API#2311


joehoyle:
@rmccue should we not add the `if` you referenced in https://github.com/WP-API/WP-API/pull/2311#issuecomment-188549081? Unless I'm missing something, this code is still doing nothing at all, we were just missing what it was supposed to be meant for?

rmccue:
@joehoyle Indeed, I did this via the GH UI though.

PR #2329, opened 2016-03-01
Bug fix: check if $data[$key][$attribute] is set before unset it

thebatclaudio:


rmccue:
Thanks for the pull request. Is there any reason for this PR? As far as I'm aware, PHP doesn't throw warnings or notices if you try to unset a non-existent variable/index, so it's safe to call this regardless.

thebatclaudio:
Hello, on Ubuntu 14.04 LTS with Nginx 1.4.6 and PHP 5.5.9 i get this error everytime. So I solved the problem in this way

rmccue:
What error are you getting?

thebatclaudio:
I get this error:

PHP Fatal error:  Cannot unset string offsets in /usr/share/nginx/website/wp-content/plugins/rest-api/lib/endpoints/class-wp-rest-controller.php on line 193

BE-Webdesign:
@rmccue There are other places in the WP API where there is an isset check before unset like in WP_REST_Controller->get_public_item_schema().  I tested it without the check and get Fatal errors as well.

danielbachhuber:
Thanks again @thebatclaudio

rmccue:
> I tested it without the check and get Fatal errors as well.

Curious. I suspect it's probably if `$data` is `''` or a similar default.

Thanks for the PR!

PR #2333, opened 2016-03-02
Support all args for terms-for-post

rmccue:
Following on from #2257: in #2311, @danielbachhuber noticed we weren't using the variable for the check. In the process of correcting that check, I discovered that the code will be broken with that check anyway.

We used to use `wp_get_object_terms` to get the uncached terms, but as it turns out, that doesn't support `include`, `exclude`, `hide_empty`, `search`, or `slug` anyway. These should be supported in `wp_get_object_terms` IMO, but we can't change this code in the meantime.

We could alternatively add the code to our `get_terms_for_post` call, but it'll be a pain as this stuff really needs to be done in the SQL query.

PR adds a failing test to show that `exclude` doesn't work currently.

### DIFF
--- a/tests/test-rest-categories-controller.php
+++ b/tests/test-rest-categories-controller.php
@@ -367,6 +367,31 @@ class WP_Test_REST_Categories_Controller extends WP_Test_REST_Controller_Testcas
 		$this->assertEquals( 'Cape', $data[0]['name'] );
 	}
 
+	public function test_get_items_post_with_exclude() {
+		$include = array();
+		$include[] = $this->factory->category->create( array( 'name' => 'Apple' ) );
+		$include[] = $this->factory->category->create( array( 'name' => 'Orange' ) );
+		$include[] = $this->factory->category->create( array( 'name' => 'Banana' ) );
+
+		// Ensure our query doesn't include this category
+		$bad_milhouse = $this->factory->category->create( array( 'name' => 'Grapefruit' ) );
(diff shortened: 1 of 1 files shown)


joehoyle:
I think this could be resolved by https://github.com/WP-API/WP-API/pull/2780, maybe @boonebgorges can chime in on that?

boonebgorges:
> I think this could be resolved by #2780, maybe @boonebgorges can chime in on that?

Yes. `wp_get_object_terms()` now supports all arguments available to `get_terms()`/`WP_Term_Query`. See https://core.trac.wordpress.org/changeset/38667

kadamwhite:
From the discussion, is this superseded by #2780? That PR looks like it's blocked until 4.7 ships, does that change any part of the equation?

kadamwhite:
@rmccue [confirmed in slack](https://wordpress.slack.com/archives/core-restapi/p1476714434007712) that this should be superseded by #2780; that PR depends on 4.7 so will be merged when applicable. Closing this one out in the meantime to reduce noise.

boonebgorges:
Just to chime in that I agree with closing this ticket in favor of #2780.

If the core merge doesn't happen for 4.7, and if the plugin plans to continue to support WP 4.6, _and_ if the team thinks that this feature is important enough, then it'd be possible to write a shim for 4.6. IMO it'd be more trouble than it's worth, and can be viewed as a progressive enhancement when running WP 4.7, but I wanted to throw it out there.

Issue #2331, opened 2016-03-01
Is the link to WP-API 1.2.5 intentional? (among other links)


mjacobsen4DFM:
The readme points to https://wordpress.org/plugins/json-rest-api/ within:
"The latest stable version is available from the [WordPress Plugin Directory](https://wordpress.org/plugins/json-rest-api/)." Which is the old version 1.2.5; which is really nothing like this version.

Likewise, the plugin page within WP admin points to http://wp-api.org/ for the "WP REST API Team" and "view details" points to an [information page](https://wirehub.dfmdev.com/wp-admin/plugin-install.php?tab=plugin-information&plugin=wp-api&TB_iframe=true&width=600&height=550)  which is regarding a [WP API version 1.0.3](https://wordpress.org/plugins/wp-api/), which is entirely different!

Is that all intentional since 2.0 is still beta?

rmccue:
Good catch, we should fix that :)

BE-Webdesign:
I can take care of this at some point today, will have a PR tonight.

mjacobsen4DFM:
It's always the little things. Brilliant code is fun... documentation, not so much.

BE-Webdesign:
@mjacobsen4DFM , @rmccue So #2334 patches up a portion of this issue however there is a plugin in the WordPress.org Plugin Repo with the name WP-API so the View Details link is being auto generated to match that plugin even though the two are not the same plugin.

rmccue:
Fixed in #2334.

PR #2334, opened 2016-03-02
Fix links in plugin docs #2331

BE-Webdesign:
This is a partial fix for #2331.  This fixes the Author URI to match v2.wp-api.org instead of v1 docs.  This also fixes a link in the Readme.md that points to v1 plugin instead of v2 (eeeek).  As for the second portion of #2331 there is a plugin in the WordPress.org repository with the slug wp-api, which is generating a View Details link if you are installing the develop branch of WP-API into your site.  Not sure how we want to solve that problem but I think that is less imperative than these other fixes.

### DIFF
--- a/README.md
+++ b/README.md
@@ -8,11 +8,11 @@ Access your WordPress site's data through an easy-to-use HTTP REST API.
 
 ## WARNING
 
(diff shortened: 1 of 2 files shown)


rmccue:
Thanks!

Re: `wp-api` issue, I don't think it's a big issue. If you clone the repo, you just need to specify that the directory is `rest-api`, not the default of `WP-API`.

PR #2335, opened 2016-03-02
Removed unnecessary global variables from users controller

claudiosanches:
No need for that global variables on `create_item()` or `get_item_schema()` methods.


danielbachhuber:
:+1: Thanks

PR #2336, opened 2016-03-02
Fix docblock for rest_insert_user

claudiosanches:
`get_user_by()` will return an instance of `WP_User`.

https://developer.wordpress.org/reference/functions/get_user_by/


danielbachhuber:
Not sure what the deal is with Scrutinizer.

PR #2337, opened 2016-03-02
Removed hard coded namespaces

claudiosanches:


claudiosanches:
Oops, that ones make sense to be hard coded if you try to extend :)
Sorry!

PR #2341, opened 2016-03-04
Add filter to taxonomy output in post object.

wkirby:
This allows users to filter the output of the taxonomy object when included on a post type object.

### DIFF
--- a/lib/endpoints/class-wp-rest-posts-controller.php
+++ b/lib/endpoints/class-wp-rest-posts-controller.php
@@ -1156,7 +1156,19 @@ class WP_REST_Posts_Controller extends WP_REST_Controller {
 		foreach ( $taxonomies as $taxonomy ) {
 			$base = ! empty( $taxonomy->rest_base ) ? $taxonomy->rest_base : $taxonomy->name;
 			$terms = get_the_terms( $post, $taxonomy->name );
-			$data[ $base ] = $terms ? wp_list_pluck( $terms, 'term_id' ) : array();
+			$term_ids = $terms ? wp_list_pluck( $terms, 'term_id' ) : array();
+
+			/**
+			 * Filter the output of taxonomy terms
+			 *
+			 * The dynamic portion of the hook name, $base, refers to the rest_base of the
+			 * taxonomy, which is either the taxonomy name or a manually configured base.
+			 *
+			 * @param Array   $term_ids      Array of term IDs on the Post object
+			 * @param Object  $taxonomy      Taxonomy that the $term_ids belong to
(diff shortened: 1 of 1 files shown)


danielbachhuber:
Why can't you do this in `"rest_prepare_{$this->post_type}"`?

wkirby:
You absolutely can; but if you want to modify the output of the taxonomy everywhere it appears, or just the output of the taxonomy (and no other attributes), this is a simpler approach, and seems like a really inexpensive change.

Compare:

``` php
// Filter Taxonomy Output to `id` => `slug` pairs for two post types
add_filter('rest_prepare_' . POST_TYPE, array(&$this, 'format_taxonomy_output'), 25, 3);
add_filter('rest_prepare_' . POST_TYPE_TWO, array(&$this, 'format_taxonomy_output'), 25, 3);

function format_taxonomy_output($response, $post, $request )
{
    if ( isset($response->data[TAXONOMY]) ) {
        $term_ids = $response->data[TAXONOMY];

        $term_objects = array_map(function($term_id) {
            return get_term_by('id', $term_id, TAXONOMY);
        }, $terms);

        $response->data[TAXONOMY] = wp_list_pluck($term_objects, 'slug', 'term_id');
    }

    return $response;
}
```

With:

``` php
// Filter Taxonomy Output to `id` => `slug` pairs everywhere it appears
add_filter('rest_prepare_terms_' . TAXONOMY, array(&$this, 'format_taxonomy_output'), 25, 3);

function format_taxonomy_output($term_ids, $taxonomy, $post )
{
    $term_objects = array_map(function($term_id) use ($taxonomy) {
        return get_term_by('id', $term_id, $taxonomy->name);
    }, $term_ids);

    return wp_list_pluck($term_objects, 'slug', 'term_id');
}
```

danielbachhuber:
Ok. We'll keep this in mind for the future, but I don't think we're going to add this filter now.

wkirby:
Is there a reason besides you don't feel like it? Filtering the full response object has a few downsides that a more specific filter avoids, namely:
- Need to validate that an attribute is set before attempting to modify it.
- Need to set up another loop if you're looking to perform the same modifications on multiple taxonomies.
- Requires explicit knowledge of the taxonomy details in more places.

By adding a filter you take away no functionality, while increasing the flexibility of the final output.

danielbachhuber:
> Is there a reason besides you don't feel like it?

While it does take a bit more code, this particular need is addressable by an existing filter. Furthermore, I don't want to set precedent of adding filters throughout response preparation for very specific needs.

wkirby:
Fair enough.

PR #2349, opened 2016-03-07
Use `$comment` from the database with `rest_insert_comment` action

### LINKED ISSUE #2340: Hook rest_insert_comment object does not have comment_ID
Seem the hook `rest_insert_comment` does not work as expected. It is fired after a comment is created or updated, so I expect it has a comment_ID but actually it not.

```
/**
 * Fires after a comment is created or updated via the REST API.
 *
 * @param array           $prepared_comment Inserted comment data.
 * @param WP_REST_Request $request          The request sent to the API.
 * @param boolean         $creating         True when creating a comment, false when updating.
 */
do_action( 'rest_insert_comment', $prepared_comment, $request, true );
```

[The implementation](https://github.com/WP-API/WP-API/blob/7063707a6818652a3e76acbf23e5a3307ce774e2/lib/endpoints/class-wp-rest-comments-controller.php) passes $prepared_comment in but not $comment and missing the ID.
Please correct me if I am wrong.

danielbachhuber:
It's a more helpful representation of the comment for callbacks to
reference.

Fixes #2340

### DIFF
--- a/lib/endpoints/class-wp-rest-comments-controller.php
+++ b/lib/endpoints/class-wp-rest-comments-controller.php
@@ -383,11 +383,11 @@ class WP_REST_Comments_Controller extends WP_REST_Controller {
 		/**
 		 * Fires after a comment is created or updated via the REST API.
 		 *
-		 * @param array           $prepared_comment Inserted comment data.
-		 * @param WP_REST_Request $request          The request sent to the API.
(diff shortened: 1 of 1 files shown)


danielbachhuber:
@WP-API/amigos #reviewmerge

rmccue:
Build failing because of https://core.trac.wordpress.org/ticket/35975; merging!

PR #2356, opened 2016-03-11
#538 add unauthenticated user creation

BE-Webdesign:
For #538. The approach taken here allows for unauthenticated user creation (i.e. a new user registering for a website). Things to consider: how we want to handle multisite user creation. I think what we have right now works well however, I am not sure how multi-site users should be handled.  Should add_user_to_blog() be added somewhere and have a blogs attribute added to the schema, containing a list of blog id's a user is associated with? Also potentially add in validation for a format password to the password arg so that only valid WordPress passwords would be accepted. Thoughts in general would be appreciated.

### DIFF
--- a/lib/endpoints/class-wp-rest-users-controller.php
+++ b/lib/endpoints/class-wp-rest-users-controller.php
@@ -27,7 +27,9 @@ class WP_REST_Users_Controller extends WP_REST_Controller {
 				'permission_callback' => array( $this, 'create_item_permissions_check' ),
 				'args'            => array_merge( $this->get_endpoint_args_for_item_schema( WP_REST_Server::CREATABLE ), array(
 					'password'    => array(
-						'required' => true,
+						'description'       => __( 'Password for user.' ),
+						'type'              => 'string',
+						'validate_callback' => 'rest_validate_request_arg',
 					),
 				) ),
 			),
@@ -248,8 +250,7 @@ class WP_REST_Users_Controller extends WP_REST_Controller {
 	 * @return boolean
 	 */
 	public function create_item_permissions_check( $request ) {
-
-		if ( ! current_user_can( 'create_users' ) ) {
+		if ( ! current_user_can( 'create_users' ) && ( ! empty( $request['password'] ) || ! get_option( 'users_can_register' ) ) ) {
 			return new WP_Error( 'rest_cannot_create_user', __( 'Sorry, you are not allowed to create resource.' ), array( 'status' => rest_authorization_required_code() ) );
 		}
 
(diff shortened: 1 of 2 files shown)


BE-Webdesign:
Going to need to fix to work with multi-site properly.

rmccue:
I remain strongly -1 on adding unauthenticated user creation. I think the potential problems we create by adding this far outweigh the benefits.

(I can't remember the exact discussion from WCSF that lead to #538.)

AlbertGuadan:
00101101 00110001 
onJul1,2015
@danielbachhuber  " https://github.com/WP-API/WP-API/issues/538/#onJul1,2015"



If the Support is needed for this creation:,  because it will become part of the .ORG Foundation tools
then i am beginning to make my opinion as to scale in growth harmoniously.

I may be a jack of all trades master of some.

I am starting to see clearly how fast the API teams growth is [trajectoring](https://twitter.com/lizcoleybooks/status/345662666439798784) , and so, my opinion is that @rmccue  could maybe forward think about the foundations future growth. ( dealing w/the REST API & because of it )   This is an old, but rather wanted enhancement that i suppose doubling the API team staff by end year would allow for more enhancements for the rest of the worlds' end users.  If the enhancement is for proprietary gains don't waist your time.

This will encourage more Open Source.

Open Source is free and if i can donate 5 bucks time to time to the Foundation i will, and i am so sure many others will and could, too.


The Wordpress.org Foundation has all ready modified the way humans view the internet.

Some times we code great,  sometimes we need human resources ( the physical office department )

The point here is blunt- i think the .ORG is moving faster than us slow thinkers can think,  is why i propose  the .org foundation to come up with a donate to the Wordpress.org community Paypal button. ( to make for inter-competitiveness maybe a separate button for each Team )
As the donations come into the Foundation more humans can contribute and solve more enhancements because the team can sometimes be collaborating in human, ( maybe use polldaddy so we can all have a say in voting ) for as long as they wish w/ salary.  The Internet is on the move we shouldn't slower,  it down and compromise it because of stubbornness in only one way to get something done. This enhancement is a year old and @royboy789 talking about he has a solution to getting it done (  on Dec 22, 2014  ) meaning -  use the new funding to fly the guy out to the .org office for a week to crunch it with a bonus if solved, is how to use the funding that comes from the new donate to Wordpress.org .ORG button

The rest of the world like the end users can donate for each initiative.
The popular enhancements can become supported shall they be .ORG Open Source







I am so sure that if  and when some of the teams collaborate together at the office they visit more solutions will be achieved for a better internet over all.

 we all appreciate your work ( the communities lost hours ) and if we are to continue to be the leading platform of open source software eventually we will face scale.  
The Foundation needs financial support and over night we will raise it.
Get me that button.

My arg, being that over 25% CMS are driven with the Foundations overall Team.
It may be time for the Team to eat well is why i propose this contextual  :_)  words.


Build the support Open Source Wordpress.org Paypal button i along with many others can and will place the code into our sites.  I feel confident that many end users may place the button into their sites to too.

if this don't make sense  never mind for now. lol

all i have to say is thanks API team keep up the awesome work

BE-Webdesign:
> I remain strongly -1 on adding unauthenticated user creation. I think the potential problems we create by adding this far outweigh the benefits.

@rmccue In many ways I agree with that, is that how the rest of the team feels as well?  Just looking at #538 there was not much saying it shouldn't be done, so I started to work on it. @rmccue , @joehoyle , @rachelbaker , @danielbachhuber thoughts on this subject!?  Also this is in 2.0 milestone. Should it not be?

danielbachhuber:
> In many ways I agree with that, is that how the rest of the team feels as well?

I can see arguments for and against, and I don't have a strong opinion at this point.

> Also this is in 2.0 milestone. Should it not be?

Given 2.0 aligns with a formal 2.0 release of the plugin, and not introduction into core, I think we need to triage the 2.0 milestone for what we feel should be in the formal 2.0 release. I'd be fine for moving unauthenticated user creation into "maybelater", or someone could just create a separate plugin for it.

BE-Webdesign:
@danielbachhuber Triage for the 2.0 milestone would be awesome, if you have the time.

danielbachhuber:
> Triage for the 2.0 milestone would be awesome, if you have the time.

Let's do it during Monday's meeting

@rmccue it would be good to get a post up about change in direction before then, if you have time.

rachelbaker:
FWIW: I also am -1 on this from being in the "core" plugin.

danielbachhuber:
Let's :football: this from the plugin then.

Issue #2357, opened 2016-03-11
`/posts/?filter[posts_per_page]=-1` returns all posts


miya0001:
`per_page` must be between 1 and 100, but following request returns all posts.

```
/wp-json/wp/v2/posts/?filter[posts_per_page]=-1
```

Is it all right?

https://github.com/WP-API/WP-API/issues/1609#issuecomment-146568663

danielbachhuber:
No, we should enforce a minimum.

miya0001:
OK! I want to send PR, please wait couple of days. :)

technotip:
Is it fixed?

I'm using filter[posts_per_page]=1 followed by &filter[before]=somevaliddate and it returns 1st post published on the blog &filter[after]=somevalidDate returns last/latest post published on the blog.

How can I solve this?

deonvisser:
Have you tried ../posts?per_page=1&before="somedate"
per_page and before are arguments not parameters. You do not need to add filter.
Works for me.

kadamwhite:
Confirmed that `/wp-json/wp/v2/posts?per_page=-1` correctly triggers validation, but `/wp-json/wp/v2/posts?filter[posts_per_page]=-1` still does return all posts.

PR #2359, opened 2016-03-12
Adds validation to filter params

miya0001:
Fixes following.
- `per_page=-1` => Status header should be 400, but 200. It has been ok with `per_page=101`.
- `filter[posts_per_page]=-1` => Validation is not working then API returns all posts.

### DIFF
--- a/lib/endpoints/class-wp-rest-controller.php
+++ b/lib/endpoints/class-wp-rest-controller.php
@@ -250,7 +250,6 @@ abstract class WP_REST_Controller {
 				'default'            => 10,
 				'minimum'            => 1,
 				'maximum'            => 100,
-				'sanitize_callback'  => 'absint',
 				'validate_callback'  => 'rest_validate_request_arg',
 			),
 			'search'                 => array(
--- a/lib/endpoints/class-wp-rest-posts-controller.php
+++ b/lib/endpoints/class-wp-rest-posts-controller.php
@@ -1702,6 +1702,7 @@ class WP_REST_Posts_Controller extends WP_REST_Controller {
 		$params['filter'] = array(
 			'description'       => __( 'Use WP Query arguments to modify the response; private query vars require appropriate authorization.' ),
 		);
+		$params['filter[posts_per_page]'] = $params['per_page'];
 		return $params;
 	}
 
--- a/plugin.php
+++ b/plugin.php
@@ -282,6 +282,13 @@ if ( ! function_exists( 'rest_validate_request_arg' ) ) {
 	 * @return WP_Error|boolean
 	 */
 	function rest_validate_request_arg( $value, $request, $param ) {
+		// for validating filter args
+		if ( preg_match( "/filter\[([a-zA-Z0-9_]+)\]/", $param, $m )) {
+			$filters = $request->get_param( 'filter' );
+			if ( ! empty( $filters[ $m[1] ] ) ) {
+				$value = $filters[ $m[1] ];
+			}
+		}
 
 		$attributes = $request->get_attributes();
 		if ( ! isset( $attributes['args'][ $param ] ) || ! is_array( $attributes['args'][ $param ] ) ) {
--- a/plugin.php
+++ b/plugin.php
@@ -283,10 +283,12 @@ if ( ! function_exists( 'rest_validate_request_arg' ) ) {
 	 */
 	function rest_validate_request_arg( $value, $request, $param ) {
 		// for validating filter args
-		if ( preg_match( "/filter\[([a-zA-Z0-9_]+)\]/", $param, $m )) {
-			$filters = $request->get_param( 'filter' );
-			if ( ! empty( $filters[ $m[1] ] ) ) {
-				$value = $filters[ $m[1] ];
+		if ( isset( $request['filter'] ) ) {
+			if ( preg_match( "/filter\[([a-zA-Z0-9_]+)\]/", $param, $matches ) ) {
+				$filters = $request->get_param( 'filter' );
+				if ( ! empty( $filters[ $matches[1] ] ) ) {
+					$value = $filters[ $matches[1] ];
+				}
 			}
 		}
 
--- a/tests/test-rest-attachments-controller.php
+++ b/tests/test-rest-attachments-controller.php
@@ -112,6 +112,7 @@ class WP_Test_REST_Attachments_Controller extends WP_Test_REST_Post_Type_Control
 			'context',
 			'exclude',
 			'filter',
+			'filter[posts_per_page]',
 			'include',
 			'media_type',
 			'mime_type',
--- a/tests/test-rest-pages-controller.php
+++ b/tests/test-rest-pages-controller.php
@@ -57,6 +57,7 @@ class WP_Test_REST_Pages_Controller extends WP_Test_REST_Post_Type_Controller_Te
 			'context',
 			'exclude',
 			'filter',
+			'filter[posts_per_page]',
 			'include',
 			'menu_order',
 			'offset',
--- a/tests/test-rest-posts-controller.php
+++ b/tests/test-rest-posts-controller.php
@@ -64,6 +64,7 @@ class WP_Test_REST_Posts_Controller extends WP_Test_REST_Post_Type_Controller_Te
 			'context',
 			'exclude',
 			'filter',
+			'filter[posts_per_page]',
 			'include',
 			'offset',
 			'order',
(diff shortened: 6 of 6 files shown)


miya0001:
https://github.com/WP-API/WP-API/issues/2357

lib/endpoints/class-wp-rest-controller.php
```diff
@@ -250,7 +250,6 @@ public function get_collection_params() {
 				'default'            => 10,
 				'minimum'            => 1,
 				'maximum'            => 100,
-				'sanitize_callback'  => 'absint',
```
BE-Webdesign, line 253: Why is absint() being taken out?

lib/endpoints/class-wp-rest-posts-controller.php
```diff
@@ -1702,6 +1702,7 @@ public function get_collection_params() {
 		$params['filter'] = array(
 			'description'       => __( 'Use WP Query arguments to modify the response; private query vars require appropriate authorization.' ),
 		);
+		$params['filter[posts_per_page]'] = $params['per_page'];
```
BE-Webdesign, line 1705: I believe most of the API would write that array like $params['filter']['posts_per_page'].

tests/test-rest-attachments-controller.php
```diff
@@ -112,6 +112,7 @@ public function test_registered_query_params() {
 			'context',
 			'exclude',
 			'filter',
+			'filter[posts_per_page]',
```
BE-Webdesign, line 115: Are these lines added in the controllers necessary?

plugin.php
```diff
@@ -282,6 +282,13 @@ function register_api_field( $object_type, $attributes, $args = array() ) {
 	 * @return WP_Error|boolean
 	 */
 	function rest_validate_request_arg( $value, $request, $param ) {
+		// for validating filter args
+		if ( preg_match( "/filter\[([a-zA-Z0-9_]+)\]/", $param, $m )) {
+			$filters = $request->get_param( 'filter' );
+			if ( ! empty( $filters[ $m[1] ] ) ) {
+				$value = $filters[ $m[1] ];
+			}
+		}
```
BE-Webdesign, line 293: Not sure this is the best approach and I would also recommend wrapping this in an `if ( isset( $request['filter'] ) )`. No reason to do a regex check for every time this function is called.  Also naming $m to $matches would be preferred in my opinion, as it makes the code more readable.  There is also a missing space on line 286 before the last parenthesis.

tests/test-rest-pages-controller.php
```diff
@@ -57,6 +57,7 @@ public function test_registered_query_params() {
 			'context',
 			'exclude',
 			'filter',
+			'filter[posts_per_page]',
```
BE-Webdesign, line 60: same as above ^

tests/test-rest-posts-controller.php
```diff
@@ -64,6 +64,7 @@ public function test_registered_query_params() {
 			'context',
 			'exclude',
 			'filter',
+			'filter[posts_per_page]',
```
BE-Webdesign, line 67: same as above ^

tests/test-rest-posts-controller.php
```diff
@@ -1664,6 +1665,50 @@ public function test_get_additional_field_registration() {
...
+		$request->set_query_params( array(
+			'per_page' => 10,
+			'filter' => array(
+				'posts_per_page' => -1
+			)
+		) );
+		$response = $this->server->dispatch( $request );
+		$this->assertEquals( 400, $response->get_status() );
+	}
+
```
BE-Webdesign, line 1711: I would also verify in the tests that the error response message matches and that you are not getting another 400 error.  Otherwise looks good!  So `$this->assertErrorResponse( 'rest_invalid_param', $response, 400 );` after any of your status checks for an error.

BE-Webdesign:
Looks good other than some minor issues.  I also think some discussing around line 285 - 291 in plugin.php would be great.

lib/endpoints/class-wp-rest-controller.php
miya0001, line 253: `per_page=-1` will be changed to `per_page=1` by `absint()`.
So API will returns a single post and status header will be `200`.
If this line would be removed, API returns `400`.

lib/endpoints/class-wp-rest-posts-controller.php
miya0001, line 1705: Do you mean like following?

```
$params['filter']['posts_per_page'] = $params['per_page'];
```

It will not work, because here is not supporting multidimensional arrays.
https://github.com/WP-API/WP-API/blob/develop/plugin.php#L284-L358

But I am not sure that it is best or not. :smile:

plugin.php
miya0001, line 293: > Not sure this is the best approach

Actually me too. :smile: 

> I would also recommend wrapping this in an if ( isset( $request['filter'] ) ). No reason to do a regex check for every time this function is called. Also naming $m to $matches would be preferred in my opinion, as it makes the code more readable. There is also a missing space on line 286 before the last parenthesis.

Oh, certainly!

tests/test-rest-posts-controller.php
miya0001, line 1711: > that you are not getting another 400 error.

Oh, I get it. :smile:

tests/test-rest-attachments-controller.php
miya0001, line 115: If the line isn't there, `phpunit` will fail.

miya0001:
@BE-Webdesign Thanks for review. I updated!

lib/endpoints/class-wp-rest-posts-controller.php
BE-Webdesign, line 1705: Okay, now I see what you are doing.  I am not sure whether this is the best approach to solving this issue.  It is basically like adding a new top level parameter but as an array which is clever, but somewhat confusing, to me at least.  This is why the lines needed to be added in the tests for query params.  This approach definitely works and adds minimal lines of code but just feels a bit odd to me; definitely clever though! :smile: I think separately validating the filter for various use cases, would be better than using rest_validate_request_arg but that is just my opinion.

lib/endpoints/class-wp-rest-controller.php
BE-Webdesign, line 253: Yup, I probably should have thought about that :smile:.

tests/test-rest-attachments-controller.php
BE-Webdesign, line 115: Okay, now I see what you did.  I don't know if it should be done that way. See my above comment for line 1705 in class-rest-posts-controller.php

lib/endpoints/class-wp-rest-posts-controller.php
miya0001, line 1705: HAHAHA, I know it looks like a cheat. :smile: 

But I found out some problem like following.
- Callback function for validation is called in the WordPress core. I couldn't figure it out that I should make ticket on WordPress Trac.
- `per_page` and `filter[posts_per_page]` are same parameter and `per_page` will over written by `filter[posts_per_page]`. The cause of this bug is the 'per_page' validation is being applied but the `filter[post_per_page]` validation is not being applied. They should both be using the same validation.

I have another idea.

I thought following is more readable?

lib/endpoints/class-wp-rest-posts-controller.php: 1705

```
$params['filter.posts_per_page'] = $params['per_page'];
```

plugin.php: 285

```
        // for validating filter args
        if ( isset( $request['filter'] ) ) {
            if ( $filter = preg_split( "/\./", $params ) && isset( $filter[1] ) && $filter[1] ) {
                $filters = $request->get_param( 'filter' );
                if ( ! empty( $filters[ $filter[1] ] ) ) {
                    $value = $filters[ $filter[1] ];
                }
            }
        }
```

lib/endpoints/class-wp-rest-controller.php
danielbachhuber, line 253: > per_page=-1 will be changed to per_page=1 by absint().

Doesn't validation happen before this, though? If validation happens before this, then the validation callback should trigger the error before it hits the sanitization callback.

miya0001, line 253: I tried that on some sites like following.
https://firegoby.jp/wp-json/wp/v2/posts?per_page=-1

miya0001, line 253: Is this problem should solved on WordPress core?
It looks that validation callback fires after sanitization callback.

danielbachhuber, line 253: > Is this problem should solved on WordPress core?
> It looks that validation callback fires after sanitization callback.

Ugh.

rmccue, line 253: We can use `intval` in any case, then validate that it's > 0.

lib/endpoints/class-wp-rest-posts-controller.php
rmccue, line 1705: I'd actually rather we just remove `posts_per_page` from `filter` and require people to use `per_page`. Let's just `unset( $args['filter']['posts_per_page'] );`

kadamwhite, line 1705: In favor—:+1: for removing `posts_per_page`. My understanding of `filter` is that it's intended as a shim for query modifications not formally supported by first-order query params; since we have `per_page`, omitting the filter param feels appropriate.

lib/endpoints/class-wp-rest-controller.php
danielbachhuber, line 253: We can't use `intval()` right now, because of #1520

BE-Webdesign, line 253: Ya sanitization happens before validation and there is a lot of things that depend on it from what I have seen.  Would there be an argument against doing something like `sanitize_callback => 'rest_intval',` , then somewhere `function rest_intval( $value, $request, $param ) { return intval( $value ); }`.  I like the white list approach but maybe use a sort of stop gap solution before white list is in core??

lib/endpoints/class-wp-rest-posts-controller.php
miya0001, line 1705: Actually my first idea was `unset( $args['filter']['posts_per_page'] );` too.
But I can't figure it out that how do we handle other pagination params.

Following should be removed too?
- nopaging (boolean) - show all posts or use pagination. Default value is 'false', use paging.
- posts_per_page (int) - number of post to show per page (available since Version 2.1, replaced showposts parameter). Use 'posts_per_page'=>-1 to show all posts (the 'offset' parameter is ignored with a -1 value). Set the 'paged' parameter if pagination is off after using this parameter. Note: if the query is in a feed, wordpress overwrites this parameter with the stored 'posts_per_rss' option. To reimpose the limit, try using the 'post_limits' filter, or filter 'pre_option_posts_per_rss' and return -1
- posts_per_archive_page (int) - number of posts to show per page - on archive pages only. Over-rides posts_per_page and showposts on pages where is_archive() or is_search() would be true.
- offset (int) - number of post to displace or pass over. Warning: Setting the offset parameter overrides/ignores the paged parameter and breaks pagination (Click here for a workaround). The 'offset' parameter is ignored when 'posts_per_page'=>-1 (show all posts) is used.
- paged (int) - number of page. Show the posts that would normally show up just on page X when using the "Older Entries" link.
- page (int) - number of page for a static front page. Show the posts that would normally show up just on page X of a Static Front Page.
- ignore_sticky_posts (boolean) - ignore post stickiness (available since Version 3.1, replaced caller_get_posts parameter). false (default): move sticky posts to the start of the set. true: do not move sticky posts to the start of the set.

https://codex.wordpress.org/Class_Reference/WP_Query

miya0001:
Ah, it looks there are two problems on this PR.
1. `per_page=-1` returns `200` and single post. It looks that this problem need to more discuss.
2. `fiter[posts_per_page]=-1` returns all posts. It will be solved by just `unset( $args['filter']['posts_per_page'] );`.

Should I open new two PRs for them and close this?

kadamwhite:
The first point above seems to have been resolved as `per_page=-1` returns an error on the latest plugin/wp. The second point, around unsetting the posts_per_page filter, still needs to be addressed.

kadamwhite:
`filter[posts_per_page]` is no longer supported as of beta 14, removed in #2699; `per_page` is now properly validated.

PR #2360, opened 2016-03-12
Fix phpdoc for _add_extra_api_post_type_arguments()

BE-Webdesign:
Fixes phpdoc for _add_extra_api_post_type_arguments().

Currently Reads: `@global array $wp_taxonomies Registered taxonomies.`
Should Read: `@global array $wp_post_types Registered post types.`


rachelbaker:
Merged #2360

PR #2361, opened 2016-03-12
Fix #2194 validate array type 

### LINKED ISSUE #2194: rest_validate_request_arg() should validate `type=array`

BE-Webdesign:
Fixes #2194.  However, there may be a better implementation but I believe this is a quick solution and works for csv lists as well, as long as they are sanitized via wp_parse_id_list() or another function which returns sanitized arrays.   Originally had a more complex system allowing for an array format to be specified like 'format' => 'id_list', but it quickly descended into madness.


danielbachhuber:
@BE-Webdesign Can you merge master, and add a couple test cases for invalid param specified on a request?

BE-Webdesign:
@danielbachhuber Yes, what do you mean by merge master?

danielbachhuber:
> Yes, what do you mean by merge master?

Sorry, I meant merge the "develop" branch to remove the failing test.

BE-Webdesign:
Yeah I can do that thank you for clarification :smiley:

BE-Webdesign:
PHP 7 failed not sure how to interpret that error message other than it wasn't my fault (maybe it is don't know).

As far as adding test coverage for invalid params, it is actually quite difficult to do this as wp_parse_id_list always returns a valid array, and I alluded to this in last week's Skype call.  So wp/v2/posts?include=ilovesteak will coerce that value into array( 0 );  As it runs absint() on everything.  To make matters worse array( 0 ), (which differs from array()) is necessary for a lot of functionality to work properly in WordPress especially around the parent properties because they accept 0 as a response to show all posts rather than posts allocated to an array of ids.  I can write tests for properties not using wp_parse_id_list but most of the array type arguments use it.  Thoughts would be appreciated.

rmccue:
PHP 7 error was a download error from svn, restarting build.

BE-Webdesign:
So as something else to mention I had an idea to do something like have array formats.  So a format id_list could be specified and then we could also handle csv lists of slugs as well. For format slug_list etc. or hybrid formats.  However this where things descended into madness quickly.  So thoughts would be good as well.

BE-Webdesign:
After reviewing my PR so far it is pretty far from complete.  There are some params with type array that need to be validated and sanitized to consistently return proper arrays.  For example the roles param has to be specified like roles[]=administrator or something to work properly.  I imagine roles=administrator is preferred with csv list support?

BE-Webdesign:
@danielbachhuber What is the desired functionality?  Because all of these sanitizing functions always return a valid array, any sort of invalid data passed in will never trigger an error it will just silently not do anything or return empty sets.  Do we want the ability to error if a bunch of nonsense is thrown into the array?  Also for schema should all of these be type => array( 'string', 'array' ) if they except csv/ssv id or slugs lists?

PR #2363, opened 2016-03-13
Remove infrastructure tests from the plugin

danielbachhuber:
These are all covered by `Tests_REST_API` in core


BE-Webdesign:
Do we want to keep em here for doc purposes or get them out to prep for core merge?

danielbachhuber:
> Do we want to keep em here for doc purposes or get them out to prep for core merge?

We've already removed most other infrastructure tests. These were missed. Infrastructure is the responsibility of WordPress core.

BE-Webdesign:
Sounds good then.

PR #2372, opened 2016-03-16
Fixes #386 roles param get users

### LINKED ISSUE #386: Add `roles` param for `GET /wp/v2/users`
I can get all users for a single role fairly easily, via the filter query parameter:
```
/wp-json/users?filter[role]=contributor
```
However, `filter[role]` doesn't seem to support the comma syntax that I can use in other commands:
```
// Doesn't work:
/wp-json/users?filter[role]=contributor,author
// I'd have expected it to work like this:
/wp-json/posts?filter[author]=17,4,164

```
Is there any way through the API to query for users of multiple roles? At present I'm having to do this by chaining batches of queries, and even with promises that request syntax is nasty. I'd much prefer to get them as a single collection.

Related #924

BE-Webdesign:
Fixes #386 roles param for get user.  Exposes role querying for GET
wp/v2/users endpoint.  This currently supports an OR relation the default
is AND but I thought using the OR operator would match up with what
seems to be how people would want to use it.  AND didn't seem to make
much sense.   if I am correct, out of the box WordPress doesn't support multiple roles
for a user?


BE-Webdesign:
Example request is authenticate as a user with list_users capability would look like this:

**For a list of authors and subscribers:**
GET wp/v2/users?roles=author,subscriber

**Only admins:**
GET wp/v2/users?roles=administrator

#review would be awesome!

tests/test-rest-users-controller.php
```diff
@@ -322,6 +323,45 @@ public function test_get_items_slug_query() {
...
+		$this->assertErrorResponse( 'rest_user_cannot_view', $response, 403 );
+	}
+
+	public function test_get_items_invalid_roles() {
+		wp_set_current_user( $this->user );
+		$lolz = $this->factory->user->create( array( 'display_name' => 'lolz' ) );
+		$request = new WP_REST_Request( 'GET', '/wp/v2/users' );
+		$request->set_param( 'roles', 'ilovesteak' );
+		$response = $this->server->dispatch( $request );
+		$this->assertErrorResponse( 'rest_invalid_param', $response, 400 );
```
danielbachhuber: Can we add a test that the invalid param is the one we inspected?

tests/test-rest-users-controller.php
```diff
@@ -322,6 +323,45 @@ public function test_get_items_slug_query() {
...
+		$request->set_param( 'roles', '' );
+		$response = $this->server->dispatch( $request );
+		$this->assertErrorResponse( 'rest_invalid_param', $response, 400 );
```
danielbachhuber: Why is this still returning an error if `roles` is empty?

lib/endpoints/class-wp-rest-users-controller.php
```diff
@@ -841,6 +858,51 @@ public function get_collection_params() {
...
 		return $query_params;
 	}
+
+	/**
+	 * Return a sanitized array of user roles.
+	 *
+	 * @param  mixed $roles Either an array of roles or an array
+	 * @return array $roles Sanitized user roles.
+	 */
+	public function rest_sanitize_user_roles( $roles ) {
```
danielbachhuber: Instead of introducing a method on this class, I want to create a `wp_parse_slug_list` to handle parsing a list of slugs. See https://core.trac.wordpress.org/ticket/35582 for the patch.

We can include our copy of the function in `core-integration.php`

lib/endpoints/class-wp-rest-users-controller.php
```diff
@@ -841,6 +858,51 @@ public function get_collection_params() {
...
+
+	/**
+	 * Get the query params for collections
+	 *
+	 * @global $wp_roles Used by get_editable_roles.
+	 *
+	 * @param  array $roles User roles.
+	 * @return boolean|WP_Error
+	 */
+	public function rest_validate_user_roles( $roles ) {
```
danielbachhuber: Does the permissions check happen before or after this validation does?

I think we might want to skip validation entirely, as we don't validate any other query params in this way at the moment.

danielbachhuber:
> #review would be awesome!

Few things to clean up

lib/endpoints/class-wp-rest-users-controller.php
BE-Webdesign: Permissions happens after as it does with every other request. I was just trying to keep things consistent.  Is that something that should be addressed in the infrastructure?  It would make more sense to me to check permissions for a request before anything is done, but I am also not an expert on that sort of stuff and if the current flow is better should I just chop the sanitization and validation out?.

lib/endpoints/class-wp-rest-users-controller.php
BE-Webdesign: Okay, sounds good.  I realized some problems with the sanitize and validate functions anyways.

tests/test-rest-users-controller.php
BE-Webdesign: If the param is specified as empty like wp/v2/users?roles= should is just go through with the request as though nothing is specified?  It's an easy fix just wasn't sure what would be appropriate.  If you had a client that was assigning something and it continually assigned empty values it might be hard to trace what was happening if you kept getting a response as though roles was never implied to be set.  If you need me to clarify my thoughts on that let me know but for now I will just remove the empty check and place it around where roles is assigned to prepared args.

tests/test-rest-users-controller.php
BE-Webdesign: Ya, sounds like a good idea.

tests/test-rest-users-controller.php
BE-Webdesign: For now, I will make an empty roles param pass silently and do nothing but if any of the nonsense above makes sense and validating against empty makes sense let me know.

lib/endpoints/class-wp-rest-users-controller.php
```diff
@@ -76,6 +77,21 @@ public function register_routes() {
...
 	/**
+	 * Permissions check for getting all users.
+	 *
+	 * @param WP_REST_Request $request Full details about the request.
+	 * @return WP_Error|boolean
+	 */
+	public function get_items_permissions_check( $request ) {
+		// Check if roles is specified in GET request and if user can list users.
+		if ( ! empty( $request['roles'] ) && ! current_user_can( 'list_users' ) ) {
+			return new WP_Error( 'rest_user_cannot_view', __( 'Sorry, you cannot filter by role.' ), array( 'status' => rest_authorization_required_code() ) );
```
danielbachhuber, line 82: Is `rest_user_cannot_view` the right error code?

tests/test-rest-users-controller.php
```diff
@@ -322,6 +323,44 @@ public function test_get_items_slug_query() {
...
+	}
+
+	public function test_get_items_invalid_roles() {
+		wp_set_current_user( $this->user );
+		$lolz = $this->factory->user->create( array( 'display_name' => 'lolz', 'role' => 'author' ) );
+		$request = new WP_REST_Request( 'GET', '/wp/v2/users' );
+		$request->set_param( 'roles', 'ilovesteak,author' );
+		$response = $this->server->dispatch( $request );
+		$data = $response->get_data();
+		$this->assertEquals( 1, count( $data ) );
```
danielbachhuber, line 360: Do taxonomy queries return the full set when a term doesn't exist, or an empty set?

BE-Webdesign, line 360: If you are referring to role as a taxonomy for users (When I hear the term taxonomy I think of post taxonomies, so just want to double check.) then I am pretty sure it returns an empty set but would have to double check and will add a test.  What is the desired functionality because it can go either way?

lib/endpoints/class-wp-rest-users-controller.php
BE-Webdesign, line 82: I don't know haha, I copied it from some of the other permission checks on get requests. For example, if someone put in context=edit they would get this error code if unauthenticated/unauthorized.

tests/test-rest-users-controller.php
danielbachhuber, line 360: > If you are referring to role as a taxonomy for users (When I hear the term taxonomy I think of post taxonomies, so just want to double check.)

I'm referring to post taxonomies. When you `GET /wp-json/wp/v2/posts?categories=999`, you'll get an empty array, not all of the posts. 

> What is the desired functionality because it can go either way?

If the role doesn't exist, I think I'd expect I'd get an empty set, given we don't error if the role doesn't exist.

BE-Webdesign, line 360: > I'm referring to post taxonomies. When you GET /wp-json/wp/v2/posts?categories=999, you'll get an empty array, not all of the posts.

I am not following at all then how does tax_query and roles intertwine? Or are you asking to match that functionality?

> If the role doesn't exist, I think I'd expect I'd get an empty set, given we don't error if the role doesn't exist.

Should we make it error on role not existing?  It does on update and create requests but wasn't sure if it should be applicable on just retrieving (empty set seems to make sense)?

danielbachhuber, line 360: > I am not following at all then how does tax_query and roles intertwine? Or are you asking to match that functionality?

I'm asking for this feature to match the behavior of the existing feature.

> Should we make it error on role not existing? It does on update and create requests but wasn't sure if it should be applicable on just retrieving (empty set seems to make sense)?

Providing an invalid role should return an empty set on `GET` requests, not return an error.

BE-Webdesign, line 360: > Providing an invalid role should return an empty set on GET requests, not return an error.

This is what it does and I added a test for it yesterday for returning empty sets.

danielbachhuber, line 360: > This is what it does and I added a test for it yesterday for returning empty sets.

Oh, I see. Sorry, I misread this test we're commenting on (`test_get_items_invalid_roles`)

BE-Webdesign:
@danielbachhuber Is this ready to go or what else needs to be done?

danielbachhuber:
> Is this ready to go or what else needs to be done?

Waiting for the final build to pass.

BE-Webdesign:
@danielbachhuber Awesome! Thanks for the feedback, @WP-API/amigos are the best! :smiley:

danielbachhuber:
:blush:

technotip:
I'm not able to fetch users based on their roles :-(
Does that require authentication?

I just want to fetch contributors list and display it on my homepage, well without using any kind of authentication - my case.

danielbachhuber:
> I'm not able to fetch users based on their roles :-(
> Does that require authentication?

Yes, it does require authentication.

> I just want to fetch contributors list and display it on my homepage, well without using any kind of authentication - my case.

You'll need to write a bit of custom code for this.

technotip:
> You'll need to write a bit of custom code for this.

Could you please help me do that? I'm not an expert in php :-(

BE-Webdesign:
@technotip http://v2.wp-api.org/guide/problems/ This has good information on how to authenticate with the API.

Issue #2368, opened 2016-03-15
Add rest_pre_insert filter for terms


danielbachhuber:
We have equivalent filters for Posts, Commands and Users, but we're missing it for terms.

kjbenk:
Are you referring to this filter within the User endpoint? 

https://github.com/kjbenk/WP-API/blob/develop/lib/endpoints/class-wp-rest-users-controller.php#L546-L605

If so, then we will need to implement the `prepare_item_for_database` function as well and call it within the `create_item` and `update_item` functions...correct?

I just want to make sure before I create a PR.

danielbachhuber:
> Are you referring to this filter within the User endpoint?

Yes, that's the existing filter on the User endpoint.

> If so, then we will need to implement the prepare_item_for_database function as well and call it within the create_item and update_item functions...correct?

I don't see creating a `prepare_item_for_database` method as _necessary_ per say, but we could do that as well.

kjbenk:
I think we should add the new `prepare_item_for_database` function since all other endpoints that have a `create_item` action have this function.  My initial thought is to refactor the code in these two places into the new `prepare_item_for_database` function:
1. https://github.com/kjbenk/WP-API/blob/develop/lib/endpoints/class-wp-rest-terms-controller.php#L335-L358
2. https://github.com/kjbenk/WP-API/blob/develop/lib/endpoints/class-wp-rest-terms-controller.php#L427-L450

@danielbachhuber Does this sound like the right direction?

BE-Webdesign:
@kjbenk I would be :+1: on that so there is consistency across the controller classes.  Those two areas look like the right place to do it as well, but I will defer the decision to @WP-API/amigos.

danielbachhuber:
> Does this sound like the right direction?

Works for me.

kjbenk:
Is there any reason why a `create_item` call within invalid term would throw a 404 when an `update_item` call with an invalid term throws a 400?

Update Item: https://github.com/WP-API/WP-API/blob/develop/lib/endpoints/class-wp-rest-terms-controller.php#L446
Create Item: https://github.com/WP-API/WP-API/blob/develop/lib/endpoints/class-wp-rest-terms-controller.php#L354

Should they both be a 400 status? @danielbachhuber

BE-Webdesign:
[...]

PR #2377, opened 2016-03-16
Add/rest pre insert filter terms

kjbenk:
Fixes this issue https://github.com/WP-API/WP-API/issues/2368


lib/endpoints/class-wp-rest-terms-controller.php
```diff
@@ -528,6 +489,52 @@ public function delete_item( $request ) {
...
+			$prepared_term->args['description'] = $request['description'];
+		}
+
+		if ( isset( $request['slug'] ) ) {
+			$prepared_term->args['slug'] = $request['slug'];
+		}
+
+		if ( isset( $request['parent'] ) ) {
+			if ( ! is_taxonomy_hierarchical( $this->taxonomy ) ) {
+				return new WP_Error( 'rest_taxonomy_not_hierarchical', __( 'Can not set resource parent, taxonomy is not hierarchical.' ), array( 'status' => 400 ) );
```
danielbachhuber: I don't think `prepare_item_for_database()` should ever return a `WP_Error`. You can silently ignore the `parent` param if it's not set in the schema.

lib/endpoints/class-wp-rest-terms-controller.php
```diff
@@ -528,6 +489,52 @@ public function delete_item( $request ) {
...
+	 * @param WP_REST_Request $request Request object.
+	 * @return object $prepared_term Term object.
+	 */
+	public function prepare_item_for_database( $request ) {
+		$prepared_term = new stdClass;
+		$prepared_term->args = array();
+
+		if ( isset( $request['name'] ) ) {
+			$prepared_term->name = $request['name'];
+			$prepared_term->args['name'] = $request['name'];
```
danielbachhuber: Er, the `$prepared_term` object should behave similarly to a WP_Term object. Its attributes shouldn't be nested under `$args`.

kjbenk: I see what you mean, but if I return a single term object then would that still work with the `wp_insert_term` function?  I see that you need to supply both the name and the args separately which is why I nested them that way.

https://github.com/WP-API/WP-API/blob/develop/lib/endpoints/class-wp-rest-terms-controller.php#L360

Do you think a better way would be to return a WP_Term object and then supply the entire object as the args and then just use the name attribute for the first parameter of `wp_insert_term`?

Maybe like this `$term = wp_insert_term( $prepared_term->name, $this->taxonomy, $prepared_term );`

danielbachhuber: > but if I return a single term object then would that still work with the wp_insert_term function? I

No, you'll need to break them apart.

> Do you think a better way would be to return a WP_Term object and then supply the entire object as the args and then just use the name attribute for the first parameter of wp_insert_term?

Yes, which is why I left the comment.

lib/endpoints/class-wp-rest-terms-controller.php
kjbenk: Would silently ignoring the `parent` parma be setting it to `-1` instead of throwing an error, or not setting it at all within the $prepared_term object?

Also, if I remove the returned WP_Error objects then should I remove these checks as well?
- https://github.com/WP-API/WP-API/pull/2377/files#diff-6a5f19de0f37a384ee52877de312ee67R409
- https://github.com/WP-API/WP-API/pull/2377/files#diff-6a5f19de0f37a384ee52877de312ee67R337

danielbachhuber: > Would silently ignoring the parent parma be setting it to -1 instead of throwing an error, or not setting it at all within the $prepared_term object?

Probably setting it to 0, as that's the closest representation to a WP Term object

> Also, if I remove the returned WP_Error objects then should I remove these checks as well?

Yes.

kjbenk:
@danielbachhuber let if me if that new commit https://github.com/WP-API/WP-API/pull/2377/commits/5bc83097db06f53c07db5bd8de4136f2904d0db7 is good for a merge.

lib/endpoints/class-wp-rest-terms-controller.php
```diff
@@ -528,6 +505,63 @@ public function delete_item( $request ) {
...
+
+		if ( isset( $request['slug'] ) ) {
+			$prepared_term->slug = $request['slug'];
+		}
+
+		if ( isset( $request['term_group'] ) ) {
+			$prepared_term->term_group = (int) $request['term_group'];
+		}
+
+		if ( isset( $request['term_taxonomy_id'] ) ) {
```
danielbachhuber: We don't support setting `term_taxonomy_id` through the REST API. `term_taxonomy_id` and `term_id` are the same in our eyes, and `term_id` is always exposed as `id`.

lib/endpoints/class-wp-rest-terms-controller.php
```diff
@@ -528,6 +505,63 @@ public function delete_item( $request ) {
...
+			$parent_term = get_term( (int) $request['parent'], $this->taxonomy );
+
+			if ( $parent_term ) {
+				$parent_term_id = $parent_term->term_id;
+			}
+
+			$prepared_term->parent = $parent_term_id;
+		}
+
+		if ( isset( $request['count'] ) ) {
```
danielbachhuber: `count` is a readonly field, and can't be set.

lib/endpoints/class-wp-rest-terms-controller.php
```diff
@@ -528,6 +505,63 @@ public function delete_item( $request ) {
...
+			$prepared_term->count = (int) $request['count'];
+		}
+
+		/**
+		 * Filter term data before inserting term via the REST API.
+		 *
+		 * @param object          $prepared_term Term object.
+		 * @param WP_REST_Request $request       Request object.
+		 */
+		return apply_filters( 'rest_pre_insert_term', $prepared_term, $request );
```
danielbachhuber: This should be `rest_pre_insert_{$this->taxonomy}`

lib/endpoints/class-wp-rest-terms-controller.php
```diff
@@ -528,6 +505,63 @@ public function delete_item( $request ) {
...
+
+		if ( isset( $request['name'] ) ) {
+			$prepared_term->name = $request['name'];
+		}
+
+		if ( isset( $request['slug'] ) ) {
+			$prepared_term->slug = $request['slug'];
+		}
+
+		if ( isset( $request['term_group'] ) ) {
```
danielbachhuber: We don't permit setting the `term_group` through the API.

danielbachhuber:
@kjbenk Better. Left comments on a few more things needing to be changed.

kjbenk:
@danielbachhuber thanks for the feedback, I am learning a lot :) The next commit has passed so you are free to review it.

danielbachhuber:
:ship: :ship: :ship:

PR #2380, opened 2016-03-17
Ensure taxonomy query args are filterable

danielbachhuber:


danielbachhuber:
@WP-API/amigos #reviewmerge

danielbachhuber:
@joehoyle Thoughts on the build failure?

kjbenk:
@danielbachhuber you need to add `tax_query` to the array of valid rest query parameters for the tests to work :)

https://github.com/WP-API/WP-API/blob/develop/lib/endpoints/class-wp-rest-posts-controller.php#L615-L628

danielbachhuber:
> you need to add tax_query to the array of valid rest query parameters for the tests to work :)

I know — wondering why we aren't doing that right now.

BE-Webdesign:
> I know — wondering why we aren't doing that right now.

That is in #2273 which could use some of your feedback!!

rmccue:
> I know — wondering why we aren't doing that right now.

`tax_query` wasn't allowed because you could have a taxonomy that's non-public and non-REST-enabled, and then query by that. We need to do validation on `tax_query` to ensure all queried taxonomies are allowed.

That's not an issue if we force the parameter to something we're building; i.e. set `$tax_query = []` before we pass it through.

rmccue:
> That's not an issue if we force the parameter to something we're building; i.e. set `$tax_query = []` before we pass it through.

(i.e. it shouldn't be an issue with the way the code is written at the moment, so it should be allowed, I think)

kadamwhite:
PR triage discussion link: https://wordpress.slack.com/archives/core-restapi/p1473802292001233

kadamwhite:
I don't know what to do to suppress that error, could use some help on this

kadamwhite:
After discussion in slack, this PR needs a clear description communicating what it is doing (seems to be permitting tax_query to be used in plugins); and we need to decide whether it is still desired after resolving #2799

PR #2381, opened 2016-03-17
Fix incorrect context mention in WP_Error

danielbachhuber:


danielbachhuber:
@WP-API/amigos #reviewmerge

BE-Webdesign:
I would say that is good to merge!

PR #2383, opened 2016-03-17
Permit `edit` context for requests which can edit the user

danielbachhuber:
If a user doesn't have `list_users`, but does have `edit_user`, they should be able to get all of the user details.


danielbachhuber:
@WP-API/amigos #reviewmerge

PR #2155, opened 2016-01-30
Permit viewing of user who published any Public posts

danielbachhuber:
Fixes #2128

PR #2384, opened 2016-03-17
Use `show_in_rest` to determine "public" post types to check

danielbachhuber:
Whether or not a user should appear in the REST API should be determined by whether they have any posts that appear in the API, not public posts generally.

From 71800a95cc722a85025aeb1fe671665b369c3de6 and #2155

### DIFF
--- a/lib/endpoints/class-wp-rest-users-controller.php
+++ b/lib/endpoints/class-wp-rest-users-controller.php
@@ -181,7 +181,7 @@ class WP_REST_Users_Controller extends WP_REST_Controller {
 
 		$id = (int) $request['id'];
 		$user = get_userdata( $id );
-		$types = get_post_types( array( 'public' => true ), 'names' );
+		$types = get_post_types( array( 'show_in_rest' => true ), 'names' );
 
 		if ( empty( $id ) || empty( $user->ID ) ) {
 			return new WP_Error( 'rest_user_invalid_id', __( 'Invalid resource id.' ), array( 'status' => 404 ) );


danielbachhuber:
@joehoyle Why did you change this originally?

BE-Webdesign:
Would this mean that someone with a published public post of a type with show_in_rest => false does not pop up?  Meaning other people can see it but rest api can't?

danielbachhuber:
> Would this mean that someone with a published public post of a type with show_in_rest => false does not pop up? Meaning other people can see it but rest api can't?

More or less, yes.

joehoyle:
@danielbachhuber I think this was intentional (though maybe incorrect) because I was trying to not disclose any more users than are already exposed via the public site (author of published posts).

rmccue:
Yeah, I think technically you could have non-public posts that have show_in_rest, and vice-versa. Maybe we need _both_?

danielbachhuber:
> Yeah, I think technically you could have non-public posts that have show_in_rest, and vice-versa.

Yep.

> Maybe we need both?

Personally, I don't think so. I think we should use `show_in_rest` across the board as the indicator of whether something should be exposed in the API.

BE-Webdesign:
The more I thought about this, the more I agree with the idea of only show_in_rest => true.  This is the REST API so if something is set to not show up in it, show_in_rest => false, why should it, just because it is marked public etc?  Same for the reverse case as well and non-public posts would just be handled as non-public posts as a separate concern.

PR #2386, opened 2016-03-17
Declare `password` in user schema, but never display it

danielbachhuber:
Doing so makes it possible to more easily negate the `required` attribute when subclassing the controller and modifying the schema.

The password is a part of the user data model, it's just never displayed.


danielbachhuber:
@WP-API/amigos #reviewmerge

PR #2286, opened 2016-02-17
Only add the alternate link header for publicly viewable CPTs.

bradyvercher:

PR #2387, opened 2016-03-18
Only add alternate link header for publicly viewable CPTs

danielbachhuber:
Props @bradyvercher

From #2286


lib/endpoints/class-wp-rest-posts-controller.php
```diff
@@ -242,7 +242,7 @@ public function get_item( $request ) {
 		$data = $this->prepare_item_for_response( $post, $request );
 		$response = rest_ensure_response( $data );
 
-		if ( is_post_type_viewable( $post->post_type ) ) {
+		if ( is_post_type_viewable( get_post_type_object( $post->post_type ) ) ) {
```
bradyvercher, line 245: Now I'm questioning everything. I dunno why I didn't catch that before. Thanks!

Issue #2342, opened 2016-03-04
Tags paging not working when fetching tags for specific post


osrecki:
**Version: 2.0-beta12**

Although `tags?per_page=PER_PAGE&page=PAGE` works as expected, when fetching tags for a specific post (`tags?post=POST_ID&per_page=PER_PAGE&page=PAGE`), paging is not working at all. 

Instead, it returns all the tags for the given post, regardless of provided `PER_PAGE` and `PAGE` arguments.

rachelbaker:
@osrecki I wasn't able to reproduce this reported bug. Looks like @kjbenk could not reproduce it either.

Closing this for now. 

If you can provide more details on how we can reproduce the bug, please re-open.

PR #2389, opened 2016-03-18
Adds a Unit Test for paginating post specific tags 

kjbenk:
This test is to test this issue https://github.com/WP-API/WP-API/issues/2342.  The test does pass so I am not 100% sure if the test is not correct or just not properly testing this scenario.

Any help with building out this test case to replicate the error would be welcomed :)

### DIFF
--- a/tests/test-rest-tags-controller.php
+++ b/tests/test-rest-tags-controller.php
@@ -224,6 +224,45 @@ class WP_Test_REST_Tags_Controller extends WP_Test_REST_Controller_Testcase {
 		$this->assertEquals( 'DC', $data[0]['name'] );
 	}
 
+	public function test_get_terms_post_args_paging() {
+		$post_id = $this->factory->post->create();
+		$tag_ids = array();
+
+		for ( $i = 0; $i < 30; $i++ ) {
+			$tag_ids[] = $this->factory->tag->create( array(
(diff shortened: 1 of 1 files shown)


rachelbaker:
Thanks @kjbenk :cake:

tests/test-rest-tags-controller.php
```diff
@@ -224,6 +224,45 @@ public function test_get_items_post_args() {
 		$this->assertEquals( 'DC', $data[0]['name'] );
 	}
 
+	public function test_get_terms_post_args_paging() {
+		$post_id = $this->factory->post->create();
+		$tag_ids = array();
+
+		for ( $i = 0; $i < 30; $i++ ) {
```
danielbachhuber, line 231: @kjbenk In the future, you might want to do this paging test with a smaller set of sample data. Generally, with unit tests, you should use as little data as you actually need for the test, because generating new data incurs database writes, and those database writes add a not-inconsequential amount of time to a full run of the test suite.

kjbenk, line 231: I totally agree and I will keep that in mind for the future, but I was just follow the other [test](https://github.com/WP-API/WP-API/blob/develop/tests/test-rest-tags-controller.php#L344) in this same file that created 50 post tags ;)

PR #2391, opened 2016-03-18
Tests for links

joehoyle:
We currently don't have good coverage for links in all the responses, this is making it difficult to change areas
of code that affect links.

### DIFF
--- a/tests/class-wp-test-rest-post-type-controller-testcase.php
+++ b/tests/class-wp-test-rest-post-type-controller-testcase.php
@@ -198,9 +198,12 @@ abstract class WP_Test_REST_Post_Type_Controller_Testcase extends WP_Test_REST_C
 		$this->assertArrayHasKey( 'X-WP-TotalPages', $headers );
 
 		$all_data = $response->get_data();
-		$data = $all_data[0];
-		$post = get_post( $data['id'] );
-		$this->check_post_data( $post, $data, $context, $response->get_links() );
+		foreach ( $all_data as $data ) {
+			$post = get_post( $data['id'] );
+			// as the links for the post are in the data array we have to pull them out and parse them
+			$links = $data['_links'];
+			$this->check_post_data( $post, $data, $context, $links );
+		}
 	}
 
 	protected function check_get_post_response( $response, $context = 'view' ) {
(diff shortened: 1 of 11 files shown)


joehoyle:
@WP-API/amigos #reviewmerge

tests/class-wp-test-rest-post-type-controller-testcase.php
```diff
@@ -153,6 +153,48 @@ protected function check_post_data( $post, $data, $context ) {
 			sort( $data[ $taxonomy->rest_base ] );
 			$this->assertEquals( $terms, $data[ $taxonomy->rest_base ] );
 		}
+
+		// test links
+		if ( $links ) {
```
danielbachhuber, line 158: Are `$links` always passed?

danielbachhuber, line 158: ... meaning, is there some way we can always test them? I'm not a huge fan of conditionals inside of unit tests because I think it can be easy to mess up some of your logic and not notice that your tests are no longer running.

tests/test-rest-comments-controller.php
```diff
@@ -1288,6 +1288,11 @@ protected function check_comment_data( $data, $context ) {
 		$this->assertEquals( mysql_to_rfc3339( $comment->comment_date_gmt ), $data['date_gmt'] );
 		$this->assertEquals( get_comment_link( $comment ), $data['link'] );
 		$this->assertContains( 'author_avatar_urls', $data );
+		$this->assertEqualSets( array(
+			'self',
+			'collection',
+			'up',
+		), array_keys( $links ) );
```
danielbachhuber, line 1295: Do we need to replicate this test in more controllers?

tests/class-wp-test-rest-post-type-controller-testcase.php
joehoyle, line 158: @danielbachhuber this is due to the fact that items in collections don't have the same links in them as getting the single item, that's to say, getting a single item via `dispatch` will return a REST_Response that you call `get_links()` on, however that's a slightly different format to an collection item's `_links` property.

tests/test-rest-comments-controller.php
joehoyle, line 1295: I'd rather get these merged, as this PR is only adding tests, these are better than none

PR #2393, opened 2016-03-19
Post endpoint - remove empty/null values from the return data.

adamsilverstein:
Fixes an issues with the JS CLient that prevented saving a created post. Receiving and then sending the null/empty values back to the server fails a validation test, and the save fails. This change strips empty values from the return.


adamsilverstein:
For more details about the reasoning behind this PR, see https://github.com/WP-API/WP-API/pull/2303.

danielbachhuber:
I don't think we should be stripping data from the response.

What if we set any date field to `null` if it equals 0000 in the database, and accept a create or update request with a null date to mean to set the date value to 000 in the database?

danielbachhuber:
For slugs, the problem needs to be addressed with #1585

kadamwhite:
Depends on resolution of #2657, there are some differences of opinion there

kadamwhite:
Per #2657 we should _not_ omit values from the data returned to the client; missing data should be exposed as `null`. Closing this PR and opening a new issue to track [making the change Daniel suggested above](https://github.com/WP-API/WP-API/pull/2393#issuecomment-198708735).

PR #2396, opened 2016-03-20
Explicitly distinguish between create and update in prepare_item_for_database()

### LINKED ISSUE #2370: Explicitly distinguish between create and update in Posts::prepare_item_for_database()
Currently, I have to infer based on whether `id` is in the `$prepared_post`, or do some deep inspection of the `$request`. It would be better if we explicitly distinguished between create and update by passing a third `$update` boolean argument to the filter.

JPry:
Fixes #2370

This updates the various versions of `prepare_item_for_database()`. There is now a `$creating` parameter that can be passed in to indicate a new item. This new parameter is also passed to the filters associated with this method.

### DIFF
--- a/lib/endpoints/class-wp-rest-controller.php
+++ b/lib/endpoints/class-wp-rest-controller.php
@@ -127,10 +127,12 @@ abstract class WP_REST_Controller {
 	/**
 	 * Prepare the item for create or update operation.
 	 *
-	 * @param WP_REST_Request $request Request object.
-	 * @return WP_Error|object $prepared_item
+	 * @param WP_REST_Request $request  Request object.
+	 * @param bool            $creating Whether the item is being created.
+	 *
+	 * @return WP_Error|stdClass $prepared_item
 	 */
-	protected function prepare_item_for_database( $request ) {
+	protected function prepare_item_for_database( $request, $creating ) {
 		return new WP_Error( 'invalid-method', sprintf( __( "Method '%s' not implemented. Must be over-ridden in subclass." ), __METHOD__ ), array( 'status' => 405 ) );
 	}
 
--- a/lib/endpoints/class-wp-rest-attachments-controller.php
+++ b/lib/endpoints/class-wp-rest-attachments-controller.php
@@ -105,7 +105,7 @@ class WP_REST_Attachments_Controller extends WP_REST_Posts_Controller {
 			}
 		}
 
-		$attachment = $this->prepare_item_for_database( $request );
+		$attachment = $this->prepare_item_for_database( $request, true );
 		$attachment->file = $file;
 		$attachment->post_mime_type = $type;
 		$attachment->guid = $url;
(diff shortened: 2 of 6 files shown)


lib/endpoints/class-wp-rest-controller.php
```diff
@@ -127,10 +127,12 @@ public function delete_item( $request ) {
 	/**
 	 * Prepare the item for create or update operation.
 	 *
-	 * @param WP_REST_Request $request Request object.
-	 * @return WP_Error|object $prepared_item
+	 * @param WP_REST_Request $request  Request object.
+	 * @param bool            $creating Whether the item is being created.
```
danielbachhuber, line 131: `wp_insert_post` uses `$updating` instead of `$creating`:

```
/**
 * Fires once a post has been saved.
 *
 * @since 2.0.0
 *
 * @param int     $post_ID Post ID.
 * @param WP_Post $post    Post object.
 * @param bool    $update  Whether this is an existing post being updated or not.
 */
do_action( 'wp_insert_post', $post_ID, $post, $update );
```

Can we follow this pre-existing pattern?

Also, do we need to necessarily pass a variable to the method, or is there a way of determining whether it's a create or update already from within the method? I think it might be safe to key off whether or not the ID is present on the prepared database object.

JPry, line 131: > Can we follow this pre-existing pattern?

Good point, I'll make those changes.

> Also, do we need to necessarily pass a variable to the method, or is there a way of determining whether it's a create or update already from within the method? I think it might be safe to key off whether or not the ID is present on the prepared database object.

This looks like it will work for all of the subclasses of `WP_REST_Controller` except for `WP_REST_Comments_Controller` and `WP_REST_Terms_Controller`. The `prepare_item_for_database()` method for these classes does not handle the ID field. For comments, it's possible this was done to facilitate an easy way to determine if only the comment `status` is changing in `update_item()`. Perhaps the logic could be updated to allow for an ID being set as well as the comment status?

As for the terms, maybe there's room for an adjustment there as well? I'm seeing this line twice:

``` php
$term = get_term( (int) $request['id'], $this->taxonomy );
```

danielbachhuber, line 131: > As for the terms, maybe there's room for an adjustment there as well? I'm seeing this line twice:

I believe that's intentional:
1. Fetch the item from the database.
2. Fetch the item again after it's been updated in the database.

BE-Webdesign, line 131: > wp_insert_post uses $updating instead of $creating:

This is unrelated to this PR in some ways but every hook in the api is using $creating should we change it to $updating and reverse the booleans to follow with the rest of wp?

joehoyle:
I'm not 100% sure that this is needed. If the caller hooking into this action has to check for `creating`, why not just check for the ID instead? 

Also, I'm not really a fan of the one-off flag `true / false` on the `prepare_item_for_database`. Can the hooking function not get everything it already needs from the `$request` object (that certainly has the ID for one).

kadamwhite:
[Discussed this in PR scrub today](https://wordpress.slack.com/archives/core-restapi/p1473800252001139), @rmccue & I agree with @joehoyle that #2370 feels like a premature optimization. The presence of an ID is by definition a good indicator that a value is being updated, and the repetition of `$request['id']` is a reasonable price to pay to avoid an over-abstraction we're not certain will be needed.

PR #2397, opened 2016-03-21
Allow `rest_validate_request_arg()` to validate arrays

JPry:
Fixex #2194

This adds array validation to `rest_validate_request_arg()` as described in #2194. Unit tests have been adeed for the new type.


plugin.php
```diff
@@ -295,6 +295,18 @@ function rest_validate_request_arg( $value, $request, $param ) {
 			}
 		}
 
+		if ( 'array' === $args['type'] ) {
+			// A comma-separated array of IDs could be acceptable instead of an array.
```
danielbachhuber, line 464: This isn't the right abstraction.

JSON Schema permits us to specify two different types for a given property. In the case of anything using `wp_parse_id_list()`, we need to specify `type => [ 'string', 'array' ]`, and then properly validate an array of types.

danielbachhuber:
@JPry: @BE-Webdesign had already started a PR for this issue in #2361. Which of you wants to take this all of the way through?

BE-Webdesign:
Doesn't matter to me, I'm happy to refine my PR or have somebody do a much better one :smiley:!

JPry:
I don't mind if @BE-Webdesign's PR goes forward instead of mine. I just happened to have a couple of hours available and wanted to help out, so I picked a couple of issues in the Beta 13 milestone.

plugin.php
JPry, line 464: > JSON Schema permits us to specify two different types for a given property. In the case of anything using `wp_parse_id_list()`, we need to specify `type => [ 'string', 'array' ]`, and then properly validate an array of types.

I just want to make sure I'm understanding you correctly. In the case of `type => [ 'string', 'array' ]`, we should ideally be checking for an array, and then ensuring that each element is a string? Or as another example, `type => [ 'integer', 'array' ]` would mean checking for an array of integers?

JPry:
I do have an hour or 2 this morning to work on this, so maybe I'll see how far I can get :smile:

plugin.php
joehoyle, line 464: In JSON Schema, I believe the way to define an arrays type is like so 

``` json
    "options": {
            "type": "array",
            "minItems": 1,
            "items": { "type": "string" },
            "uniqueItems": true
        },
```

See http://json-schema.org/example2.html

joehoyle:
@JPry is there any update that needs to be added here, or is this changed direction and needs to be closed out?

JPry:
@joehoyle Give me a day or two to review this, and then we can close it out if nothing has changed.

JPry:
I made some updates here based on [this comment](https://github.com/WP-API/WP-API/pull/2397#r75177757). If we're following the JSON schema as @joehoyle mentioned, then the changes I've made seem to be the right way forward.

At this point I'm looking for some feedback on the approach I'm using. If what I've done here _isn't_ the right way to go about this, then I would appreciate some input about what a better method might be.

I'm also curious about the failed tests from when I merged `develop` into my branch. I didn't have a chance to examine the tests in-depth yet, but a cursory examination makes it seem like it _may_ not be my changes that caused the failures.

joehoyle:
@JPry thanks for expanding on this, so I think this works, however i think it could be more elegant. If we have a function `validate_value_by_schema( $object, $schema )`, we could have this function recursively call it's self in the event of finding an array. This would also allow us to deep check json schema objects too, I think.

I have an idea of how this could work, I'm happy to adapt this to try that direction.

JPry:
Thanks for the idea @joehoyle. I'd be willing to make some updates to this based on your idea and what @BE-Webdesign mentioned in slack today. For my own reference, here are the relevant links from slack:

https://wordpress.slack.com/archives/core-restapi/p1476455590007085
https://wordpress.slack.com/archives/core-restapi/p1476455898007096

plugin.php
```diff
@@ -552,6 +587,21 @@ function rest_sanitize_request_arg( $value, $request, $param ) {
...
+			if ( is_string( $value ) && isset( $args['allow_csv'] ) && $args['allow_csv'] ) {
+				$value = wp_parse_id_list( $value );
+			}
+
+			$value = (array) $value;
+
+			// Map values to the correct type, defaulting to string as the type.
+			$type = isset( $args['items']['type'] ) ? $args['items']['type'] : 'string';
+			foreach ( $value as &$_value ) {
+				settype( $_value, $type );
```
BE-Webdesign, line 601: Rather than using `settype` we should use our own validation systems because `settype` can lead to the wrong information.  We can get away with `settype` though as most of the endpoints only deal with arrays of strings or integers.  If this PR plans to handle booleans then this will cause a problem when values like `'false'` are passed into an array booleans.

tests/test-rest-controller.php
```diff
@@ -15,6 +15,14 @@ public function setUp() {
...
 					'type'        => 'string',
 				),
+				'somearray'       => array(
+					'type'        => 'array',
+					'allow_csv'   => false,
+				),
+				'somecsv'         => array(
+					'type'        => 'array',
+					'allow_csv'   => true,
+				),
```
BE-Webdesign, line 25: We need to test against the actual JSON Schema here.  So we need to have items added.  In a PR I did that I ultimately did not like I had schema like this.

``` php
someIdList => array(
    'type' => 'array',
    'items' => array(
        'type' => 'integer'
    ),
   'minItems' => 1,
   'uniqueItems' => true,
),
```

Then the same thing for string lists.  Currently in the endpoints most of the arrays are id lists and only a few are string lists.

plugin.php
```diff
@@ -460,6 +460,41 @@ function rest_validate_request_arg( $value, $request, $param ) {
...
+			$function = '__return_false';
+			switch ( $type ) {
+				case 'integer':
+					$function = 'is_numeric';
+					break;
+
+				case 'boolean':
+					$function = 'rest_is_boolean';
+					break;
+
```
BE-Webdesign, line 485: I wouldn't worry about supporting boolean because there are no endpoint parameters that use an array of booleans. This PR won't handle booleans 100% properly so let's not check for them.

plugin.php
```diff
@@ -460,6 +460,41 @@ function rest_validate_request_arg( $value, $request, $param ) {
...
+				case 'string':
+					$function = 'is_string';
+					break;
+			}
```
BE-Webdesign, line 489: This should have a default case so there is no fall through, currently in the API we use super loose validation and most things will pass for true by default.  I believe this PR is returning false for all of these values, that fall through.

lib/endpoints/class-wp-rest-users-controller.php
```diff
@@ -905,16 +905,18 @@ public function get_collection_params() {
...
-			'description'        => __( 'Ensure result set excludes specific ids.' ),
-			'type'               => 'array',
-			'default'            => array(),
-			'sanitize_callback'  => 'wp_parse_id_list',
+			'description'       => __( 'Ensure result set excludes specific ids.' ),
+			'type'              => 'array',
+			'default'           => array(),
+			'sanitize_callback' => 'wp_parse_id_list',
+			'validate_callback' => 'rest_validate_request_arg',
 		);
```
BE-Webdesign, line 913: The `items` schema needs to be added into all of the array parameters so that your validation is actually working.  This is why all of the tests are failing because the schema is not set yet.  Every include/exclude author_include/exclude parameter needs it and I believe user roles.  There are some other ones that I can't remember.

PR #2403, opened 2016-03-23
Bring JavaScript client up to date

adamsilverstein:
Changes in this commit: https://github.com/WP-API/client-js/compare/0d01ff4b9712ef05f9329946e95ff5d602f0ac4e...master
- Support lodash, plus older and newer underscore: add an alias for `_.contains`
- Add args and options on the model/collection prototypes
- Rework category/tag mixins to support new API structure
- Add workaround for the null/empty values returned by the API when creating a new post - these values are not accepted for subsequent updates/saves, so explicitly excluding them. See https://github.com/WP-API/WP-API/pull/2393
- Better handling of the (special) `me` endpoint
- Schema parsing cleanup
- Introduce `wp.api.loadPromise` so developers can ensure api load complete before using


danielbachhuber:
@adamsilverstein Do we need to version bump the enqueued script?

adamsilverstein:
@danielbachhuber good point. bumped in latest commit.

PR #2407, opened 2016-03-25
Accept empty string as orderby

redundans:
First, I want to thank you all for a great work with the WP REST API. We use it a lot. But this other day we found a problem in using the API to fetch live search result. So I wanted to propose a very simple solution that we ended up using in our latest project.
## The problem

When using the WP-API to achive live search with AJAX you will not be able to get the result ordered by relevance.
## The solution

As stated in WP core query.php (https://core.trac.wordpress.org/browser/tags/4.4.2/src/wp-includes/query.php#L3028) search results is ordered by relevance only when another "orderby" is not specified in the query. The problem is that there is no way not to specify "orderby" in WP REST API v2.

WordPress seams to handle this by setting the query variable of "orderby" to an empty string. By letting the user to pass an empty string through the WP-APIs function rest_validate_request_arg. This is easiest achived by returning $params['orderby']['em'] = array('', ... ) in the function get_collection_params.

### DIFF
--- a/lib/endpoints/class-wp-rest-posts-controller.php
+++ b/lib/endpoints/class-wp-rest-posts-controller.php
@@ -1668,7 +1668,7 @@ class WP_REST_Posts_Controller extends WP_REST_Controller {
 			'type'               => 'string',
 			'default'            => 'date',
 			'enum'               => array(
-				'',
+				'relevance',
 				'date',
 				'id',
 				'include',
--- a/lib/endpoints/class-wp-rest-posts-controller.php
+++ b/lib/endpoints/class-wp-rest-posts-controller.php
@@ -1668,6 +1668,7 @@ class WP_REST_Posts_Controller extends WP_REST_Controller {
 			'type'               => 'string',
 			'default'            => 'date',
 			'enum'               => array(
+				'',
 				'date',
 				'id',
 				'include',
--- a/tests/test-rest-posts-controller.php
+++ b/tests/test-rest-posts-controller.php
@@ -272,6 +272,20 @@ class WP_Test_REST_Posts_Controller extends WP_Test_REST_Post_Type_Controller_Te
 		$this->assertEquals( 'Apple Cobbler', $data[0]['title']['rendered'] );
 	}
 
+	public function test_get_items_orderby_relevance() {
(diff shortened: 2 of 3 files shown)


BE-Webdesign:
@redundans Thanks for using the WP REST API! :smile:

We need unit tests in this PR to cover your use case.  Any new functionality for the API requires testing coverage.

Also this will probably need the blessing of someone who was more involved with the decision to have a default =>  'name' and more familiar with orderby intricacies in general.

redundans:
@BE-Webdesign I ran som local tests for this before we used it in production. But there is another solution for search ordering. We could pass an empty string or "relevance" as orderby to the the filter parameter. Like this:

`/wp-json/wp/v2/post?search=XXX&filter[orderby]=`
or
`/wp-json/wp/v2/post?search=XXX&filter[orderby]=relevance`

Ordering by relevance is quite badly documented in the WP_Query's query function. Maybe this pull request is redundant?

BE-Webdesign:
Yes, that would be another way to do it by avoiding the top level param.  There are a lot of those redundancies in the API currently.  However, adding this functionality to the top level orderby param, I think would be useful. We just need unit tests to cover it (WordPress uses phpunit and you can find our tests in the `tests` directory in the plugin).

The only argument I would see against it is that the sorting should be done in the application using the API as opposed to using the filter `posts_search_orderby`, which you still can currently use, as you have pointed out, with `?filter[orderby]=`.

joehoyle:
I'd rather go with a `relevance` flag for this, or maybe `_relevance` to indicate `relevance` is not a field of the posts. IMO we shouldn't expose the fact that WP_Query happens to order by relevance on an ampty value.

danielbachhuber:
> I'd rather go with a relevance flag for this, or maybe _relevance to indicate relevance is not a field of the posts.

Me too.

> IMO we shouldn't expose the fact that WP_Query happens to order by relevance on an ampty value.

And it's only sometimes too, as it can be disabled with a filter.

joehoyle:
Cool, so I take my `_relevance` idea back, as we use normal string to indicate things like this, that are not object property names. 

We should probably run said filter to know whether to add it to the enum?

BE-Webdesign:
Ordering by relevance landed in #2579.

PR #2412, opened 2016-03-28
Use Compact links for embedded responses if they are available.

joehoyle:
This fixes curies compatibility with 4.5, and doesn't fatal under 4.4

### DIFF
--- a/core-integration.php
+++ b/core-integration.php
@@ -35,3 +35,50 @@ if ( ! function_exists( 'wp_parse_slug_list' ) ) {
 		return array_unique( $list );
 	}
 }
+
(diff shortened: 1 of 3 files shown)


joehoyle:
@WP-API/amigos #reviewmerge

danielbachhuber:
@joehoyle Why don't any of the tests need changing?

joehoyle:
@danielbachhuber all the tests are set up to use the full-uris, not compact URIs, we have a helper `rest_extand_curies` function or something similar.

danielbachhuber:
Ok.

Issue #2392, opened 2016-03-18
Update documentation on contributing to WP-API


danielbachhuber:
Specifically:
- [x] How to run tests and the linter locally.
- [x] Link to CONTRIBUTING.md from README.md

kjbenk:
Here are some other things worth noting:
1. Explanation of our continuously deployment system, [Travis CI](https://travis-ci.org/WP-API/WP-API).
2. Explain and document what `grunt` is and how to install and run it.  Maybe go over the `Gruntfile.js`.
3. Explanation of our coding standards, aka a slimmed down version of WordPress VIP. Maybe go over the `phpcs.ruleset.xml`.

Once contributors understand these topics and tools they should commit pull requests with a higher pass rate :)

PR #2413, opened 2016-03-28
Make it easier for others to get started contributing

danielbachhuber:
See #2392


danielbachhuber:
@BE-Webdesign @kjbenk Could I get your thoughts and feedback on these changes?

BE-Webdesign:
Yeah, this covers the process.  Maybe put in links to npm/node and composer so if they are unfamiliar with the two they can get more information there.  https://nodejs.org/ and https://getcomposer.org/

PR #2415, opened 2016-03-30
Fix #2398 Update return codes to 404 when object not found

### LINKED ISSUE #2398: Return consistent status code on update / delete requests to invalid resource id
Sometimes we're returning a 404:

```
if ( ! $term ) {
    return new WP_Error( 'rest_term_invalid', __( "Resource doesn't exist." ), array( 'status' => 404 ) );
}
```

Other times, we're returning a 400:

```
$user = get_userdata( $id );
if ( ! $user ) {
    return new WP_Error( 'rest_user_invalid_id', __( 'Invalid resource id.' ), array( 'status' => 404 ) );
}
```

We should consistently return a 404 when the resource doesn't exist. 400 is less precise, and better for reporting invalid request arguments.

antisilent:
Fixes #2398 to consistently return 404 when objects aren't found in the database. I may have missed a couple, but I'm pretty sure I got everything. :)

### DIFF
--- a/lib/endpoints/class-wp-rest-posts-controller.php
+++ b/lib/endpoints/class-wp-rest-posts-controller.php
@@ -395,7 +395,7 @@ class WP_REST_Posts_Controller extends WP_REST_Controller {
 		$post = get_post( $id );
 
 		if ( empty( $id ) || empty( $post->ID ) || $this->post_type !== $post->post_type ) {
-			return new WP_Error( 'rest_post_invalid_id', __( 'Post id is invalid.' ), array( 'status' => 400 ) );
+			return new WP_Error( 'rest_post_invalid_id', __( 'Post id is invalid.' ), array( 'status' => 404 ) );
 		}
 
 		$post = $this->prepare_item_for_database( $request );
@@ -795,7 +795,7 @@ class WP_REST_Posts_Controller extends WP_REST_Controller {
 			if ( get_current_user_id() !== $post_author ) {
 				$user_obj = get_userdata( $post_author );
 				if ( ! $user_obj ) {
-					return new WP_Error( 'rest_invalid_author', __( 'Invalid author id.' ), array( 'status' => 400 ) );
+					return new WP_Error( 'rest_invalid_author', __( 'Invalid author id.' ), array( 'status' => 404 ) );
 				}
 			}
 			$prepared_post->post_author = $post_author;
@@ -906,7 +906,7 @@ class WP_REST_Posts_Controller extends WP_REST_Controller {
 			if ( $result ) {
 				return true;
 			} else {
-				return new WP_Error( 'rest_invalid_featured_media', __( 'Invalid featured media id.' ), array( 'status' => 400 ) );
+				return new WP_Error( 'rest_invalid_featured_media', __( 'Invalid featured media id.' ), array( 'status' => 404 ) );
 			}
 		} else {
 			return delete_post_thumbnail( $post_id );
--- a/lib/endpoints/class-wp-rest-posts-controller.php
+++ b/lib/endpoints/class-wp-rest-posts-controller.php
@@ -795,7 +795,7 @@ class WP_REST_Posts_Controller extends WP_REST_Controller {
 			if ( get_current_user_id() !== $post_author ) {
 				$user_obj = get_userdata( $post_author );
 				if ( ! $user_obj ) {
-					return new WP_Error( 'rest_invalid_author', __( 'Invalid author id.' ), array( 'status' => 404 ) );
+					return new WP_Error( 'rest_invalid_author', __( 'Invalid author id.' ), array( 'status' => 400 ) );
 				}
 			}
 			$prepared_post->post_author = $post_author;
--- a/lib/endpoints/class-wp-rest-terms-controller.php
+++ b/lib/endpoints/class-wp-rest-terms-controller.php
@@ -340,7 +340,7 @@ class WP_REST_Terms_Controller extends WP_REST_Controller {
 			$parent = get_term( (int) $request['parent'], $this->taxonomy );
 
 			if ( ! $parent ) {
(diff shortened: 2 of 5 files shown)


BE-Webdesign:
Thanks @antisilent!

The unit tests failed because they are still expecting 400 instead of 404.  Please go through the test files and change them to match the new error statuses.  Here is our current guide on contribution to the API https://github.com/WP-API/WP-API/blob/develop/CONTRIBUTING.md.

BE-Webdesign:
Also I believe, you need to keep the ones for creating on 400 not 404.  My interpretation of how 404 would be used is if you are querying for a resource that doesn't exist i.e. updating a non existent post, deleting non existent user.  For query params that don't exist I would probably lean towards using 400.  For instance if you wanted to update a page's parent and the page exists but the parent doesn't you should get a 400 because it is a bad request as opposed to the page resource not existing.  That is just my interpretation of how it should be handled in the API and I would be glad to be corrected.

lib/endpoints/class-wp-rest-terms-controller.php
```diff
@@ -340,7 +340,7 @@ public function create_item( $request ) {
 			$parent = get_term( (int) $request['parent'], $this->taxonomy );
 
 			if ( ! $parent ) {
-				return new WP_Error( 'rest_term_invalid', __( "Parent resource doesn't exist." ), array( 'status' => 400 ) );
+				return new WP_Error( 'rest_term_invalid', __( "Parent resource doesn't exist." ), array( 'status' => 404 ) );
```
BE-Webdesign: I believe this should still be 400.

lib/endpoints/class-wp-rest-users-controller.php
```diff
@@ -628,7 +628,7 @@ protected function check_role_update( $user_id, $roles ) {
 		foreach ( $roles as $role ) {
 
 			if ( ! isset( $wp_roles->role_objects[ $role ] ) ) {
-				return new WP_Error( 'rest_user_invalid_role', sprintf( __( 'The role %s does not exist.' ), $role ), array( 'status' => 400 ) );
+				return new WP_Error( 'rest_user_invalid_role', sprintf( __( 'The role %s does not exist.' ), $role ), array( 'status' => 404 ) );
```
BE-Webdesign: This would stay as 400 as well, I believe.

lib/endpoints/class-wp-rest-users-controller.php
```diff
@@ -368,11 +368,11 @@ public function update_item( $request ) {
...
 		$user = get_userdata( $id );
 		if ( ! $user ) {
-			return new WP_Error( 'rest_user_invalid_id', __( 'Invalid resource id.' ), array( 'status' => 400 ) );
+			return new WP_Error( 'rest_user_invalid_id', __( 'Invalid resource id.' ), array( 'status' => 404 ) );
 		}
 
 		if ( email_exists( $request['email'] ) && $request['email'] !== $user->user_email ) {
-			return new WP_Error( 'rest_user_invalid_email', __( 'Email address is invalid.' ), array( 'status' => 400 ) );
+			return new WP_Error( 'rest_user_invalid_email', __( 'Email address is invalid.' ), array( 'status' => 404 ) );
 		}
```
BE-Webdesign, line 376: ^ This should definitely be 400.

lib/endpoints/class-wp-rest-terms-controller.php
```diff
@@ -420,7 +420,7 @@ public function update_item( $request ) {
...
```
BE-Webdesign: I think this should be 400 as well.

lib/endpoints/class-wp-rest-posts-controller.php
```diff
@@ -795,7 +795,7 @@ protected function prepare_item_for_database( $request ) {
 			if ( get_current_user_id() !== $post_author ) {
 				$user_obj = get_userdata( $post_author );
 				if ( ! $user_obj ) {
-					return new WP_Error( 'rest_invalid_author', __( 'Invalid author id.' ), array( 'status' => 400 ) );
+					return new WP_Error( 'rest_invalid_author', __( 'Invalid author id.' ), array( 'status' => 404 ) );
```
BE-Webdesign: I believe this would be 400.

antisilent:
Test cases have been updated and codes have been reverted to 400 in cases where parameters or prerequisites are missing. :)

lib/endpoints/class-wp-rest-posts-controller.php
```diff
@@ -906,7 +906,7 @@ protected function handle_featured_media( $featured_media, $post_id ) {
 			if ( $result ) {
 				return true;
 			} else {
-				return new WP_Error( 'rest_invalid_featured_media', __( 'Invalid featured media id.' ), array( 'status' => 400 ) );
+				return new WP_Error( 'rest_invalid_featured_media', __( 'Invalid featured media id.' ), array( 'status' => 404 ) );
```
rachelbaker, line 909: This should remain a `400` error because the provided featured_image id doesn't exist and the request needs to be modified.

lib/endpoints/class-wp-rest-users-controller.php
```diff
@@ -380,7 +380,7 @@ public function update_item( $request ) {
...
 
 		if ( ! empty( $request['slug'] ) && $request['slug'] !== $user->user_nicename && get_user_by( 'slug', $request['slug'] ) ) {
-			return new WP_Error( 'rest_user_invalid_slug', __( 'Slug is invalid.' ), array( 'status' => 400 ) );
+			return new WP_Error( 'rest_user_invalid_slug', __( 'Slug is invalid.' ), array( 'status' => 404 ) );
```
rachelbaker, line 383: This should remain a `400` error because the provided `slug` is not allowed and the request needs to be modified.

rachelbaker:
@antisilent Thank you for the PR.  Two of these error codes should remain `400`.  This needs a refresh to remove the two error code changes noted (and their related test changes) and then it will be ready for merge.  

Would you be able to update the PR?

Issue #1784, opened 2015-12-03
Comments: 409 Conflict and HTML response when posting a duplicate comment


sdellow:
If a duplicate comment gets posted via the API you get an error like this:

    <!DOCTYPE html>
	
	<html xmlns="http://www.w3.org/1999/xhtml" lang="en-GB">
	<head>
	    <meta http-equiv="Content-Type" content="text/html; charset=utf-8" />
	    <meta name="viewport" content="width=device-width">
	        <title>WordPress &rsaquo; Error</title>
	        <style type="text/css">
		html {
			background: #f1f1f1;
		}
		body {
			background: #fff;
			color: #444;
			font-family: "Open Sans", sans-serif;
			margin: 2em auto;
			padding: 1em 2em;
			max-width: 700px;
			-webkit-box-shadow: 0 1px 3px rgba(0,0,0,0.13);
			box-shadow: 0 1px 3px rgba(0,0,0,0.13);
		}
		h1 {
			border-bottom: 1px solid #dadada;
			clear: both;
			color: #666;
			font: 24px "Open Sans", sans-serif;
			margin: 30px 0 0 0;
			padding: 0;
			padding-bottom: 7px;
		}
		#error-page {
			margin-top: 50px;
		}
		#error-page p {
			font-size: 14px;
			line-height: 1.5;
			margin: 25px 0 20px;
		}
		#error-page code {
			font-family: Consolas, Monaco, monospace;
		}
		ul li {
			margin-bottom: 10px;
			font-size: 14px ;
		}
		a {
			color: #21759B;
			text-decoration: none;
		}
		a:hover {
			color: #D54E21;
		}
		.button {
			background: #f7f7f7;
			border: 1px solid #cccccc;
			color: #555;
			display: inline-block;
			text-decoration: none;
			font-size: 13px;
			line-height: 26px;
			height: 28px;
			margin: 0;
			padding: 0 10px 1px;
			cursor: pointer;
			-webkit-border-radius: 3px;
			-webkit-appearance: none;
			border-radius: 3px;
			white-space: nowrap;
			-webkit-box-sizing: border-box;
			-moz-box-sizing:    border-box;
			box-sizing:         border-box;

			-webkit-box-shadow: inset 0 1px 0 #fff, 0 1px 0 rgba(0,0,0,.08);
			box-shadow: inset 0 1px 0 #fff, 0 1px 0 rgba(0,0,0,.08);
		 	vertical-align: top;
		}

		.button.button-large {
[...]

Issue #1660, opened 2015-10-20
Consecutive `POST /comments` get 200 instead of 201 in a period of time.


southp:
## Environment
WordPress: 4.3.1
WP REST API: 2.0-beta 4
Machine: [Chassis](https://github.com/Chassis/Chassis) with VirtualBox 4.3.28 as its VM backend

## How To
Here is the snippet of the shell script I used to experiment. It uses [httpie](httpie.org). Simply replace the url and the post id should do the trick:

```
#!/bin/sh
date_str=`date`
http POST http://<site-url>/wp-json/wp/v2/comments author_name='Shell' \
    author_email='shell@xd.cc' \
    author_url='http://goog.le' \
    content="from shell $date_str" \
    post=<pid> \
    parent=0
```
In my case:

1. Run the script the first time. It should return `201`, and the anonymous comment is created successfully.
1. Run the script again. It returned 200 without response body.
1. Wait for a period of time ( about 30sec ~ 1 min. ). Now you should return to the step 1.

Toddses:
This is probably default WordPress behavior. Anonymous comments in WordPress are limited to one every X seconds. I think the default is 15 seconds.

danielbachhuber:
Core calls `wp_die()` for this check ([ref](https://core.trac.wordpress.org/browser/tags/4.3/src/wp-includes/comment.php#L1501)), so we'll need to address it when we decide how to handle `wp_die()` #790

For now, I'll add a skipped test #1662

rmccue:
Per https://github.com/WP-API/WP-API/issues/790#issuecomment-73872576 can we avoid this function ever getting called? We should be able to run the `comment_flood_filter` filter ourselves before `wp_allow_comment` (which eventually triggers the action) is run.

danielbachhuber:
> can we avoid this function ever getting called?

Avoid calling `check_comment_flood_db()` entirely? I should think we need to persist that behavior.

websupporter:
We could utilize `wp_allow_comment()` in `create_item_permissions_check()`, if the risk of running into `wp_die()` wouldn't be there. Since 4.4 we have `wp_handle_comment_submission()` which handles the exit process in `wp-comments-post.php`. So, if we could transform `wp_die()` for flood and duplicate comments into an WP Error object, which travels all the way to `wp-comments-post.php`, this would be the step to go, right?

I've tried to get this started here: https://core.trac.wordpress.org/ticket/36901

rachelbaker:
[...]

PR #2417, opened 2016-04-03
#790 wp die handler

### LINKED ISSUE #790: Add special handling for wp_die
We need to add a `wp_die_handler` filter so we can return a response when WordPress evokes `wp_die`.

Reference link to how Jetpack handles this: https://github.com/Automattic/jetpack/blob/a8f890c21db29f0d4ec22734fbb19b74cfa45bb1/class.json-api.php#L554

BE-Webdesign:
Fixes #790, #1660, #1784.  Adds a wp_die() handler for the WP REST API.  Not sure if this is the best implementation.  Would love some feedback and will probably mull over it a bit before it is ready for merge.


BE-Webdesign:
Need to check update handler for duplicate comments as well so a comment can't be edited to a duplicate.  Also there was a skipped test that I did not unskip.

BE-Webdesign:
Just need an O.K. on the implementation.  Not sure if I am sold on it entirely myself though.

danielbachhuber:
> Just need an O.K. on the implementation.

I think the prior consensus was that we needed to abstract away the use of `wp_die()` in core for the functions we care about. If you wanted to prepare patches for those instances, we might be able to get them in 4.6-early

BE-Webdesign:
Okay, I can't do anything today but can probably whip something up tomorrow.  Is there already a trac ticket?

danielbachhuber:
> Is there already a trac ticket?

I don't know, actually. Probably worth discussing at Monday's meeting.

BE-Webdesign:
Closing out as this was decided to be handled in a core patch of comments.

Issue #1612, opened 2015-09-29
How to implement comment threading using `GET /comments` ?


southp:
I'm working on a plugin featuring viewport-based infinite scrolling of comments using WP API v2.
As the endpoint `GET /comments` responds paged data in descending order of date (which is reasonable), I am wondering how could I display comment threads properly before everything is loaded? For example, like the hierarchical looks of the theme twenty-fifteen.

For example, suppose there are 5 comments in descending order: c4, c3, c2, c1, c0, and the parent of c4 is c0. If I set `page=1, per_page=4`, there is no way I can show c4 as a reply of c0 before c0 is actually loaded.

Currently I only have thought of two possible ways:

* Do not use viewport-based infinite scrolling. Load new pages automatically instead, and refresh the comment hierarchy as needed.
    * I think this can solve the problem, but what if users' prefer viewport-based way, like I do? 
* Add a `child` field in the response data.
    * I think this might be the weirdest, because there is no such a data in WP database. Also, the thread setting might be changed.

I'm wondering, is there any other way? Do I miss any helpful feature of WP API for this?

Thanks.

joehoyle:
Hmm interesting, so this is how I see it working:

`GET http://demo.wp-api.org/wp-json/wp/v2/comments?post=49&parent=0?_embed` and I think we could then add a `replies` link to each comment which would basically link to  `http://demo.wp-api.org/wp-json/wp/v2/comments?parent={comment_id}`. Passing `_embed` would then allow all the replies to load in (recursively). The only issue I see with this is it would be potentially a lot of data to load at once (if say a given comment has 200 replies) but that may be an edge case.

You'd probably want to also only embed the replies, not the rest of the embeddable links (see `up`), which we don't currently support in the API. This kind of hierarchical stuff isn't super simple in REST, but I think the use case is a valid one. I can see that we are 90% of the way there to have a good way to do this. I'd say the immediate todos:

1. Add a `replies` link to each comment that is `embeddable`
1. Discuss how we can to recursive whitelisted embedding

southp:
Hi @joehoyle ,
Thanks for taking this further.

I think adding an `parent` query parameter is a good way. I was thinking about how to do thread collapsing like Facebook does and it was not an easy thing with the current API. With it, I think both use cases can be fulfilled.

As I am quite new to this, I would like to ask for some clarifications:

> Passing _embed would then allow all the replies to load in (recursively)  
>

[...]

PR #2418, opened 2016-04-04
#1612 threaded comment links

BE-Webdesign:
Solves #1612.  Adds links for children of the comment when they have permission to see that comment.  Weird thing happening in the unit tests in `test_get_items_parent_exclude_arg()`.  I will most likely need some help with that before merge.


BE-Webdesign:
I have a much, much better solution to this and when I get around to it, I will revise this.  Instead of this approach, I plan to return a link to a collection of comments with the matching parent.  This way pagination can happen among other things.

danielbachhuber:
> I have a much, much better solution to this and when I get around to it, I will revise this. Instead of this approach, I plan to return a link to a collection of comments with the matching parent. This way pagination can happen among other things

👍

Issue #2416, opened 2016-03-30
Only include Post data on the response object when declared in schema


danielbachhuber:
This is what our code path looks like currently:

```
setup_postdata( $post );

// Base fields for every post.
$data = array(
    'id'           => $post->ID,
    'date'         => $this->prepare_date_response( $post->post_date_gmt, $post->post_date ),
    'date_gmt'     => $this->prepare_date_response( $post->post_date_gmt ),
    'guid'         => array(
        /** This filter is documented in wp-includes/post-template.php */
        'rendered' => apply_filters( 'get_the_guid', $post->guid ),
        'raw'      => $post->guid,
    ),
    'modified'     => $this->prepare_date_response( $post->post_modified_gmt, $post->post_modified ),
    'modified_gmt' => $this->prepare_date_response( $post->post_modified_gmt ),
    'password'     => $post->post_password,
    'slug'         => $post->post_name,
    'status'       => $post->post_status,
    'type'         => $post->post_type,
    'link'         => get_permalink( $post->ID ),
);

$schema = $this->get_item_schema();

if ( ! empty( $schema['properties']['title'] ) ) {
    $data['title'] = array(
        'raw'      => $post->post_title,
        'rendered' => get_the_title( $post->ID ),
    );
}
```

If I remove `title` as a property from the schema, it's not included in the response. However, if I remove `password` from schema properties, it's still included on the response object.

I'd expect the response object would only have fields declared in the schema, or at least that WP REST API wouldn't blindly add database data to the response if was purposefully removed from the schema.

websupporter:
I wonder about `categories` and `tags`. They are not included in `$schema` and not in the initial `$data`. Shouldn't they be part of the scheme? But I guess, this would make thinks difficult since all taxonomies with `"show_in_rest" => true` are showing up as a key.

Otherwise, I just thought of something like this:

```
foreach( $data as $key => $property )
    if( ! isset( $schema['properties'][ $key ] ) || empty( $schema['properties'][ $key ] ) )
        unset( $data[ $key ] );
```

websupporter:
sorry, my mistake its included. perfect!

danielbachhuber:
We also need to make sure create / update requests don't process data that's not included in the schema.

PR #2423, opened 2016-04-06
 Include Post data on the response object when declared  #2416

websupporter:
See https://github.com/WP-API/WP-API/issues/2416


danielbachhuber:
Sorry, this isn't quite the right approach. We should permit arbitrary data to be included in the response — however the core fields shouldn't be included by default unless they're a part of the schema. See how the `title` field is included as an example of how the others should follow.

websupporter:
thanks @danielbachhuber, will rework my approach.

danielbachhuber:
@websupporter Sorry, I missed your updates. Feel free to ping me next time so the PR isn't left hanging.

PR #2428, opened 2016-04-07
#2424 mixed usage of rest url

### LINKED ISSUE #2424: Mixed Usage of path formatting when calling rest_url for add_link
We are utilizing the `rest_url` filter for WordPress.com and encountered an issue with our logic, specifically around link results for the users endpoint.

It seems there is varying usage of the `path` param passed to `rest_url()` in the endpoint logic.  In most cases, there is a leading slash used in the path parameter [#](https://github.com/WP-API/WP-API/blob/develop/lib/endpoints/class-wp-rest-users-controller.php#L547), but in other cases there is not [#](https://github.com/WP-API/WP-API/blob/03d3708d06f6ac9f9a6ff8e2be9af62123fdbc8b/lib/endpoints/class-wp-rest-taxonomies-controller.php#L152).

So in our filter for `rest_url` we have to sniff out if a path has a leading slash or not - not the end of the world, but it might be nice to have a consistent standard in the core endpoints when calling `rest_url()` for links.

BE-Webdesign:
Should fix #2424, however, after writing a unit test it doesn't appear to matter whether the '/' is there or not as get_rest_url adds a leading slash regardless as far as I can tell.  Also it is very difficult to write this kind of test so any help would be appreciated.


tests/class-wp-test-rest-controller-testcase.php
```diff
@@ -6,7 +6,7 @@
 
 	public function setUp() {
 		parent::setUp();
-
+		add_filter( 'rest_url', array( $this, 'test_rest_url_for_leading_slash' ), 10, 2 );
```
danielbachhuber: I think you need to remove this filter on `tearDown` too

tests/class-wp-test-rest-controller-testcase.php
```diff
@@ -39,4 +39,8 @@ public function tearDown() {
 
 	abstract public function test_get_item_schema();
 
+	public function test_rest_url_for_leading_slash( $url, $path ) {
+		// Make sure path for rest_url has a leading slash for proper resolution.
+		$this->assertTrue( 0 === strpos( $path, '/' ) );
```
danielbachhuber, line 44: Because you're running this test on a filter, you need to make sure to return the `$url` :) Otherwise, the sky will fall upon your head, as it has done.

BE-Webdesign, line 44: Yeah, I thought of that not sure why I never put it in.  The last thing is that this function is called on every subclass as though it is a test so you get an exception thrown because no parameters are passed into it.  I tried to make it protected but then the subclasses can't access it.  Should I change this to parent in the filter?

danielbachhuber, line 44: You can rename it to `filter_rest_url_for_leading_slash`

BE-Webdesign, line 44: Thanks, that solved everything.  One last thing though, the test itself doesn't really help at all for what we are trying to do as `get_rest_url()` always adds the leading slash before it hits the filter `rest_url`.  How would we test the input into the filter to see if there is a slash or not?  Currently the test itself just checks to make sure there is a leading slash at the start of $path and asserts whether that is true which it is regardless of whether there is a leading slash or not being put into the `rest_url()`.

danielbachhuber, line 44: > One last thing though, the test itself doesn't really help at all for what we are trying to do as get_rest_url() always adds the leading slash before it hits the filter rest_url.

Oh. @timmyc How are you running into your problem then?

BE-Webdesign, line 44: I think they are rewriting `rest_url()`, from my understanding.

joehoyle:
This looks good to me, I think we can roll this in as-is. Thanks @BE-Webdesign!

PR #2429, opened 2016-04-08
#2426 inconsistent type for user caps

### LINKED ISSUE #2426: Inconsistent type returned for user capabilities.
When a WP user has capabilities and a call is made to `/users/:id?context=edit`, a JSON object/map is returned (`"capabilities": {"a":"foo", "b":"bar"}`).

However, when a user has no capabilities (`- No role for this site -` selected in wp-admin), an empty array is returned (`"capabilities": []`).

BE-Webdesign:
Fixes #2426, returns an empty object for users who do not have capabilities.


lib/endpoints/class-wp-rest-users-controller.php
```diff
@@ -517,6 +517,17 @@ public function prepare_item_for_response( $user, $request ) {
 		}
 
 		$context = ! empty( $request['context'] ) ? $request['context'] : 'embed';
+
+		if ( 'edit' === $context ) {
```
danielbachhuber: Let's check like this:

```
if ( isset( $data['capabilities'] ) && empty( $data['capabilities'] ) ) {
    $data['capabilities'] = new stdClass();
}
```

This way, we'll always fix our data regardless of `context`.

lib/endpoints/class-wp-rest-users-controller.php
```diff
@@ -517,6 +517,17 @@ public function prepare_item_for_response( $user, $request ) {
...
+			if ( empty( $data['capabilities'] ) ) {
+				$data['capabilities'] = new stdClass();
```
danielbachhuber: I'm not sure whether this should actually default to an empty object, or `null`. Can you take a look to see if there's prior art?

lib/endpoints/class-wp-rest-users-controller.php
BE-Webdesign: Okay, that's how I originally had it. I thought that since it only ever shows up in edit context, it should be in that conditional.  Will revert it.

lib/endpoints/class-wp-rest-users-controller.php
BE-Webdesign: I would lean towards empty object as the schema itself is object, or we can change the schema do be array( 'object', 'null' );  I will check to see if there is any info regarding what is better.

tests/test-rest-users-controller.php
```diff
@@ -1079,8 +1096,16 @@ protected function check_user_data( $user, $data, $context, $links ) {
 			$this->assertEquals( $user->last_name, $data['last_name'] );
 			$this->assertEquals( $user->nickname, $data['nickname'] );
 			$this->assertEquals( $user->user_email, $data['email'] );
-			$this->assertEquals( $user->allcaps, $data['capabilities'] );
-			$this->assertEquals( $user->caps, $data['extra_capabilities'] );
+			if ( 'object' === gettype( $data['capabilities'] ) ) {
+				$this->assertEquals( new stdClass(), $data['capabilities'] );
+			} else {
+				$this->assertEquals( $user->allcaps, $data['capabilities'] );
+			}
```
danielbachhuber: This doesn't seem quite right — our tests shouldn't be expecting two different types.

BE-Webdesign: It's because associative arrays equal objects in json.  That is why there is the difference.  The weird part is that before I put these changes in, this test only flagged for failing on running multisite.xml.  I agree that this should not be there though, but I am not quite sure how to get around having to include it.

lib/endpoints/class-wp-rest-users-controller.php
danielbachhuber: > Can you take a look to see if there's prior art?

I took a look at the codebase. For the date fields, we use `null` instead of an empty string if there's no date stored. For `parent`, we use `null` if `0` is stored in the database. I think we should probably use `null` instead of an empty object.

BE-Webdesign:
Forgot the tests for multisite don't work the same on check_user_data.

BE-Webdesign:
#review-merge  Not sure why check_user_data only fails on multisite. I think we should probably investigate that some.

lib/endpoints/class-wp-rest-users-controller.php
```diff
@@ -517,6 +517,15 @@ public function prepare_item_for_response( $user, $request ) {
 		}
 
 		$context = ! empty( $request['context'] ) ? $request['context'] : 'embed';
+
+		if ( isset( $data['capabilities'] ) && empty( $data['capabilities'] ) ) {
```
rmccue: These checks are redundant; `empty` checks `isset()` internally already.

rmccue:
(Sorry for the late comment.)

I kinda like an empty object here, since it's an object rather than a scalar value already. It indicates that the user has 0 capabilities rather than not having capabilities at all (if that makes sense?)

Re: multisite failures, that's because on multisite, users don't have any capabilities on a site until they're added to it. On single sites, I believe they always have at least one capability.

BE-Webdesign:
> Re: multisite failures, that's because on multisite, users don't have any capabilities on a site until they're added to it. On single sites, I believe they always have at least one capability.

Yeah, I eventually figured that out hehe.

> I kinda like an empty object here, since it's an object rather than a scalar value already.

I would be pro empty object, but I also agree with danielbachuber that we should keep things consistent and most empty values we send through the API come out as null.  If you guys can give me a strong vote either way, I would be glad to revise.

lib/endpoints/class-wp-rest-users-controller.php
BE-Webdesign: Not sure what I was thinking exactly, thanks for catching that and pointing it out.

BE-Webdesign:
#reviewmerge - Went back to empty object and fixed conditional checks.

lib/endpoints/class-wp-rest-users-controller.php
```diff
@@ -566,6 +566,15 @@ public function prepare_item_for_response( $user, $request ) {
 		}
 
 		$context = ! empty( $request['context'] ) ? $request['context'] : 'embed';
+
+		if ( empty( $data['capabilities'] ) ) {
+			$data['capabilities'] = new stdClass();
```
rmccue: Rather than using `new stdClass`, can we instead just cast `$user->caps` to an `object` above?

lib/endpoints/class-wp-rest-users-controller.php
```diff
@@ -817,13 +826,13 @@ public function get_item_schema() {
 				),
 				'capabilities'    => array(
 					'description' => __( 'All capabilities assigned to the resource.' ),
-					'type'        => 'object',
+					'type'        => array( 'object', 'null' ),
```
rmccue: Seems like this patch changes this to actually be just `object`?

tests/test-rest-users-controller.php
```diff
@@ -446,6 +446,22 @@ public function test_get_user_invalid_id() {
 		$this->assertErrorResponse( 'rest_user_invalid_id', $response, 404 );
 	}
 
+	public function test_get_user_empty_capabilities() {
+		wp_set_current_user( $this->user );
+		$this->allow_user_to_manage_multisite();
+
+		$lolz = $this->factory->user->create( array( 'display_name' => 'lolz', 'roles' => '' ) );
+		delete_user_option( $lolz, 'capabilities' );
+		delete_user_option( $lolz, 'user_level' );
```
rmccue, line 455: Nice variable naming. ;)

lib/endpoints/class-wp-rest-users-controller.php
BE-Webdesign: I knew I missed something.

tests/test-rest-users-controller.php
BE-Webdesign, line 455: Do you need me to change it?

lib/endpoints/class-wp-rest-users-controller.php
BE-Webdesign: Sure, I like that idea.

BE-Webdesign:
Is this good to go now?

tests/test-rest-users-controller.php
rmccue, line 455: Given our tests are full of Simpsons references, not an issue. :)

BE-Webdesign, line 455: :)

PR #2435, opened 2016-04-13
Avoid unnecessary SQL query by passing `$user_nicename`

danielbachhuber:
If no `$user_nicename` is supplied, `get_author_posts_url()` calls `get_userdata()` to get the `$user_nicename`. Because we already have it, we don't need to incur an unnecessary database hit.


danielbachhuber:
@WP-API/amigos #reviewmerge

PR #2436, opened 2016-04-13
Avoid unnecessary SQL SELECT against user meta

danielbachhuber:
Because we're fetching all fields, `WP_User_Query` primes our user objects with their role / capabilities. This data is stored in user meta. But, core [isn't yet smart enough](https://core.trac.wordpress.org/ticket/36508) to precache user meta when `fields=>all`. We can be smarter than core by using `fields=>all_with_meta`


danielbachhuber:
@WP-API/amigos #reviewmerge

PR #2438, opened 2016-04-13
Mark more post properties as readonly

danielbachhuber:
These properties are truly readonly; if you try to update them, nothing
will update.


danielbachhuber:
@WP-API/amigos #reviewmerge

PR #2439, opened 2016-04-13
Only return or update user properties when defined in schema

danielbachhuber:
If a WordPress has removed the field from the schema, it shouldn't be
included in the response, or be unexpectedly updated in the database.


danielbachhuber:
@WP-API/amigos #reviewmerge

PR #2440, opened 2016-04-13
Mark Users' `capabilities` property as readonly

danielbachhuber:
It's not actually editable through the API


danielbachhuber:
@WP-API/amigos #reviewmerge

PR #2441, opened 2016-04-13
Revert "Avoid unnecessary SQL SELECT against user meta"

danielbachhuber:
Reverts WP-API/WP-API#2436

@boonebgorges made a wise comment about the negative performance impact of loading user meta when a given site has abused the amount of data they put in user meta. For now, I think it's better for developers to opt-in to this change as needed.


dleeward:
Is there code that we non-developers can add to our custom-functions file that will allow us to GET and POST to non-protected usermeta?

Issue #2382, opened 2016-03-17
Add get_items_permissions_check() method to Users controller


danielbachhuber:
To be consistent with our other controllers, we should have a `get_items_permission_check()` method on the Users controller. If `context=edit` and the user can't edit users, it should return an error, similar to our existing controllers.

BE-Webdesign:
I'll add this into #2372?  Or is it better to have a separation of concerns and create a separate PR?

danielbachhuber:
> Or is it better to have a separation of concerns and create a separate PR?

Separate PR, please :)

PR #2463, opened 2016-04-24
return error, if user can't list users & content=edit

websupporter:
Solves https://github.com/WP-API/WP-API/issues/2382 

Instead of the capability `edit_user` I took 'list_users' like it is also done in the `get_item_permissions_check()`. Did make sense to me.


danielbachhuber:
Thanks for the pull request! Can you include a test case for this change?

websupporter:
I hope this is what's needed :)

danielbachhuber:
> I hope this is what's needed :)

Perfect!

PR #2470, opened 2016-04-27
Conditionally model the term response based on its schema

danielbachhuber:
Similarly, only permit editing fields present on the schema


danielbachhuber:
@WP-API/amigos #reviewmerge

joehoyle:
Do we have the same issue on the other objects?

danielbachhuber:
Yes. I've just been running into the issue piecemeal, as I extend the controllers.

danielbachhuber:
Related #2416

PR #2479, opened 2016-05-04
create/update requests don't process data not included in the schema

websupporter:
> We also need to make sure create / update requests don't process data that's not included in the schema.
> https://github.com/WP-API/WP-API/issues/2416

This commit updates `prepare_item_for_database()` which is used by `update_item()` and `create_item()` to prepare the post. Here it also checks for the schema.


websupporter:
@danielbachhuber ping :)
I think this is what's needed. 
Please have a look. 

Thank you.

danielbachhuber:
👍

PR #2495, opened 2016-05-12
*Improve regex detection of me

pcfreak30:
Current regex will match /wp/v2/site due to the `e`. It doesn't do an alternative list.


rmccue:
Thanks for the PR! Our JS client is maintained on a [separate repo](https://github.com/WP-API/client-js), can you file the PR there instead please? :)

pcfreak30:
Done

PR #2500, opened 2016-05-15
Remove Unused Parameter in lib/endpoints/class-wp-rest-controller.php

hideokamoto:
I run [phpmd](https://phpmd.org/), and find following notification.

```
/var/www/wordpress/wp-content/plugins/rest-api/lib/endpoints/class-wp-rest-controller.php:292Avoid unused local variables such as ''.
```

So, I removed  parameter


danielbachhuber:
👍

PR #2535, opened 2016-06-09
Introduce WP_REST_Controller::get_post() for allowing plugins to mutate get_post()'s return value

westonruter:
There is no `get_post` filter as there is when calling `get_comment()` or `get_term()` or `get_{$type}_meta()` and so on. This can make it very difficult particularly in the Customizer to preview changes to posts. Currently the Customizer relies on queries to be made using `suppress_filters => false` and/or for posts to be iterated in The Loop where `setup_postdata` is called and thus `the_post` is done, both of which provide opportunities for a plugin (here Customize Posts) to mutate the post before it is returned. 

In the REST API, we hit a wall recently in Customize Posts with accessing a post that is saved in the DB as a draft with a Customizer state that makes the status `publish`. We are working to allow such posts to be viewed via the REST API publicly for unauthenticated users if a snapshot (aka transaction) UUID is supplied to recall that Customizer state in which the post has the status of `publish`. The problem is that `\WP_REST_Posts_Controller::get_item_permissions_check()` uses `get_post()` to obtain the post object and then passes this into `\WP_REST_Posts_Controller::check_read_permission()` without any opportunity for a plugin to override the `post_status` of this object to have the desired behavior of forcing it to be publicly accessible given the Customizer state.

In lieu of there being a `get_post` filter as proposed in [#12955](https://core.trac.wordpress.org/ticket/12955#comment:33), I'm suggesting that the REST API introduce `\WP_REST_Controller::get_post()` wrapper for `get_post()` which includes a `the_post` action for plugins to extend the object to be returned.

See https://github.com/xwp/wp-customize-snapshots/issues/32

### DIFF
--- a/lib/endpoints/class-wp-rest-attachments-controller.php
+++ b/lib/endpoints/class-wp-rest-attachments-controller.php
@@ -48,7 +48,7 @@ class WP_REST_Attachments_Controller extends WP_REST_Posts_Controller {
 
 		// Attaching media to a post requires ability to edit said post
 		if ( ! empty( $request['post'] ) ) {
-			$parent = get_post( (int) $request['post'] );
+			$parent = $this->get_post( (int) $request['post'] );
 			$post_parent_type = get_post_type_object( $parent->post_type );
 			if ( ! current_user_can( $post_parent_type->cap->edit_post, $request['post'] ) ) {
 				return new WP_Error( 'rest_cannot_edit', __( 'Sorry, you are not allowed to upload media to this resource.' ), array( 'status' => rest_authorization_required_code() ) );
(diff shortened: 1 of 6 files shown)


rachelbaker:
@westonruter Thank you for the PR.  

I have two concerns:
1. Applying `the_post` action to the REST API endpoint Post interactions would mean any themes or plugins that have been using this hook for a given use-case would now unexpectedly impact the sites REST API endpoints permissions and data availability. 
2. Only using the current Core `the_post` action, without also providing a `rest_` specific filter/action would make it difficult for themes and plugins to limit targeting ONLY REST API data OR ONLY Core data.

I see two possible solutions:
1. Use a new `rest_the_post` action **INSTEAD** of `the_post`.
2. Add a `rest_the_post` filter to the return value of the `the_post` method.

What do you think?

westonruter:
@rachelbaker thanks for the feedback! I've implemented a `rest_the_post` filter as you suggested in f3fc6d5.

Issue #2284, opened 2016-02-16
comments can be send with empty values


bobsilon:
Hi there,
Is it good the wp-api save the comments with empty values?
If I submit a comment form with empty fields, empty `author_name`, empty  `author_email`, empty `author_url` and others except `post`, It will be submitted in database. Is it right behavior?

danielbachhuber:
> Is it good the wp-api save the comments with empty values?

I think we're changing this behavior with #2049

I'll flag for @rachelbaker to confirm the expected behavior.

kadamwhite:
This should be resolved now that #2741 has been merged

PR #2540, opened 2016-06-11
Check for comment duplicates and flood

websupporter:
This PR tries to solve https://github.com/WP-API/WP-API/issues/1660

As long as `wp_allow_comment()` can die, we need a workaround. Hopefully, this will change in a future release (https://core.trac.wordpress.org/ticket/36901)

I guess, the patch needs some more work and suggestions, but to get this started with.

`grunt phpcs` returns two problems. I need to do direct database queries to reproduce the functionality and I think, we shouldn't cache those queries if I am not mistaken. 

```
----------------------------------------------------------------------
FOUND 2 ERRORS AND 2 WARNINGS AFFECTING 2 LINES
----------------------------------------------------------------------
 366 | WARNING | Usage of a direct database call is discouraged.
 366 | ERROR   | Usage of a direct database call without caching is
     |         | prohibited. Use wp_cache_get / wp_cache_set or
     |         | wp_cache_delete.
 396 | WARNING | Usage of a direct database call is discouraged.
 396 | ERROR   | Usage of a direct database call without caching is
     |         | prohibited. Use wp_cache_get / wp_cache_set or
     |         | wp_cache_delete.
----------------------------------------------------------------------
```

I was thinking to outsource this functionality in a `rest_allow_comments()` to keep the `create_item_permissions_check()` clean and readable?

Any suggestions are welcome :)

### DIFF
--- a/lib/endpoints/class-wp-rest-comments-controller.php
+++ b/lib/endpoints/class-wp-rest-comments-controller.php
@@ -304,107 +304,10 @@ class WP_REST_Comments_Controller extends WP_REST_Controller {
 			}
 		}
 
-		global $wpdb;
 		$commentdata = $this->prepare_item_for_database( $request );
-
-		/* Make sure, we have all the necessary data */
-		$commentdata['comment_author']       = ! isset( $commentdata['comment_author'] )       ? '' : $commentdata['comment_author'];
-		$commentdata['comment_author_email'] = ! isset( $commentdata['comment_author_email'] ) ? '' : $commentdata['comment_author_email'];
-		$commentdata['comment_author_url']   = ! isset( $commentdata['comment_author_url'] )   ? '' : $commentdata['comment_author_url'];
-		$commentdata['comment_author_IP']    = ! isset( $commentdata['comment_author_IP'] )    ? '127.0.0.1' : $commentdata['comment_author_IP'];
-
-		$commentdata['comment_date']     = ! isset( $commentdata['comment_date'] )     ? current_time( 'mysql' )                           : $commentdata['comment_date'];
-		$commentdata['comment_date_gmt'] = ! isset( $commentdata['comment_date_gmt'] ) ? get_gmt_from_date( $commentdata['comment_date'] ) : $commentdata['comment_date_gmt'];
-
-		$commentdata['comment_post_ID']  = ! isset( $commentdata['comment_post_ID'] )  ? 0  : $commentdata['comment_post_ID'];
-		$commentdata['comment_content']  = ! isset( $commentdata['comment_content'] )  ? '' : $commentdata['comment_content'];
-		$commentdata['comment_karma']    = ! isset( $commentdata['comment_karma'] )    ? 0  : $commentdata['comment_karma'];
-		$commentdata['comment_approved'] = ! isset( $commentdata['comment_approved'] ) ? 1  : $commentdata['comment_approved'];
-		$commentdata['comment_agent']    = ! isset( $commentdata['comment_agent'] )    ? '' : $commentdata['comment_agent'];
-		$commentdata['comment_type']     = ! isset( $commentdata['comment_type'] )     ? '' : $commentdata['comment_type'];
-		$commentdata['comment_parent']   = ! isset( $commentdata['comment_parent'] )   ? 0  : $commentdata['comment_parent'];
-
-		/* This filter is documented in wp-includes/comment.php */
-		$commentdata = apply_filters( 'preprocess_comment', $commentdata );
-		$commentdata = wp_filter_comment( $commentdata );
-
-		// Set author data if the user's logged in
-		$missing_author = empty( $commentdata['user_id'] )
-			&& empty( $commentdata['comment_author'] )
-			&& empty( $commentdata['comment_author_email'] )
-			&& empty( $commentdata['comment_author_url'] );
-
-		if ( is_user_logged_in() && $missing_author ) {
-			$user = wp_get_current_user();
-			$commentdata['user_id'] = $user->ID;
-			$commentdata['comment_author'] = $user->display_name;
-			$commentdata['comment_author_email'] = $user->user_email;
-			$commentdata['comment_author_url'] = $user->user_url;
-		}
-
-		//Test for duplicate comment
-		//catches the functionality of wp_allow_comment() in order to
-		//prevent a wp_die and return a WP_Error
-		$dupe = $wpdb->prepare(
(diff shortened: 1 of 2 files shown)


lib/endpoints/class-wp-rest-comments-controller.php
```diff
@@ -720,7 +823,7 @@ protected function prepare_item_for_database( $request ) {
 		}
 
 		if ( isset( $request['parent'] ) ) {
-			$prepared_comment['comment_parent'] = $request['parent'];
+			$prepared_comment['comment_parent'] = (int) $request['parent'];
```
websupporter, line 729: I've added `(int)` because somehow `$request['parent'']` was an empty array in some tests (e.g. `test_get_items_no_post()`. I haven't searched yet, why (it is not indicated by the test itself though) and I am not sure if there is a bug behind, which should be fixed. I will have a look, but maybe someone knows already why it is sometimes an array.

BE-Webdesign, line 729: Not sure why it would be an array but definitely should be type cast to int.  I think maybe putting the new logic into a protected method somewhere in the controller would be nice, to keep `create_permissions_check()` clean. When your patch lands in core we can just cut the lines back out.  Not a super big deal though.  Looks good to me.

BE-Webdesign:
Thanks for all of the work on this, big props!  Not sure what to do about code sniffer failing though. Maybe try `// @codingStandardsIgnoreStart` and `// @codingStandardsIgnoreEnd` to wrap around the code.  Not an expert on caching yet so I would defer that judgement to someone else hehe.

tests/test-rest-comments-controller.php
```diff
@@ -929,7 +926,7 @@ public function test_create_comment_two_times() {
 		$request->set_body( wp_json_encode( $params ) );
 
 		$response = $this->server->dispatch( $request );
-		$this->assertEquals( 400, $response->get_status() );
+		$this->assertEquals( 429, $response->get_status() );
```
BE-Webdesign: Can we do a `$this->assertErrorResponse()` instead?  Same thing for the 409 response in `test_create_item_duplicate()`.

lib/endpoints/class-wp-rest-comments-controller.php
websupporter, line 729: The `get_collection_params()` defines it as an array, so you can use `parent__in` of the `WP_Comment_Query` in `get_items()`. So I guess, the best would be to use `(int)` and make sure its an integer when it comes to writing the database, as I've did.

websupporter:
Hi @BE-Webdesign,
thanks for your support. Tried it so far locally and this would work without errors. I have a couple more questions, where you might be able to guide me.
1. If we go through the usual comment process, we first run into `wp_handle_comment_submission()`. This function does a lot of validation and for example returns a WP_Error if there is no name or no email-address and so on. Right now, we do not check for those things in the REST API. Should we?
2. For my check, I do not simply reproduce `wp_allow_comment()` but I try to grab all the filters, which might alter the commentdata before, this is `preprocess_comment` and `wp_filter_comment()` which are triggered in `wp_new_comment()` before this function calls `wp_allow_comment()`. I thought this might be useful. But in `create_item()` we skip these filters since we completely skip `wp_new_comment()` and do use `wp_insert_comment()` directly. Shouldn't we have those filters also in the `create_item()`? I guess, there is a reason, why we skip `wp_new_comment()` and go more directly, but I think it would be useful to run those filters during our creation process?
3. Regarding `(int)` for the parent parameter. I just had a second look into the output of `http://local.wordpress-trunk.dev/wp-json/wp/v2/` for the `wp/v2/comments/`. In the args of the POST method it tells me to use an integer and the default is `0`. I compared the `prepare_item_for_database()` between post and comments. post always checks back with the schema

```
if ( ! empty( $schema['properties']['content'] ) && isset( $request['content'] ) )
```

while comments just checks for 

```
if ( isset( $request['content'] ) ) {
```

shouldn't this be brought inline? Should I do a seperate PR for this or just fix it here together with this PR? Related: https://github.com/WP-API/WP-API/issues/2416

websupporter:
Ah, for point no. 1: Related https://github.com/WP-API/WP-API/issues/2284

BE-Webdesign:
> If we go through the usual comment process, we first run into wp_handle_comment_submission(). This function does a lot of validation and for example returns a WP_Error if there is no name or no email-address and so on. Right now, we do not check for those things in the REST API. Should we?

I think eventually yes, but it would also probably be a separate PR.  I believe someone has already begun work on that, but feel free to patch it as there hasn't been much activity on that front.

> For my check, I do not simply reproduce wp_allow_comment() but I try to grab all the filters, which might alter the commentdata before, this is preprocess_comment and wp_filter_comment() which are triggered in wp_new_comment() before this function calls wp_allow_comment(). I thought this might be useful. But in create_item() we skip these filters since we completely skip wp_new_comment() and do use wp_insert_comment() directly. Shouldn't we have those filters also in the create_item()? I guess, there is a reason, why we skip wp_new_comment() and go more directly, but I think it would be useful to run those filters during our creation process?

Not totally sure why `wp_new_comment()` is not used and `wp_insert_comment()` is, so can't really comment on that.  I think there was a patch being done somewhere that was also trying to make use of `wp_new_comment()`.  I would honestly have to look into more.

> Regarding (int) for the parent parameter. I just had a second look into the output of http://local.wordpress-trunk.dev/wp-json/wp/v2/ for the wp/v2/comments/. In the args of the POST method it tells me to use an integer and the default is 0. I compared the prepare_item_for_database() between post and comments. post always checks back with the schema
> 
> if ( ! empty( $schema['properties']['content'] ) && isset( $request['content'] ) )
> while comments just checks for
> 
> if ( isset( $request['content'] ) ) {
> shouldn't this be brought inline? Should I do a seperate PR for this or just fix it here together with this PR?

Number three I think would be a separate PR, unless it directly affects the new comment handling for duplicates and flood.  I think there are a couple more endpoints, that do not follow the check for schema and maybe one PR could be done for all of them.

websupporter:
@BE-Webdesign Thanks for your notes!

Regarding `wp_new_comment()`

> Turns out I can't use wp_new_comment instead of wp_insert_comment since it means that you won't be able to create a comment without a post_id associated with it which one of the tests requires.
> (https://github.com/WP-API/WP-API/pull/2049)

Well, I guess for now, I leave this PR as it is and wait if there is more feedback on the issue worked on here.

websupporter:
Just quickly @rmccue & @danielbachhuber, we will not push this here if I understood you correctly on WCEU? Just because I was stupid enough not to create a new branch for this and I would restart my whole fork now to continue on other stuff than?

danielbachhuber:
Right, we should fix the core functions instead of copying them into there.

PR #2553, opened 2016-06-21
The get_the_excerpt filter expects the post object as of WP 4.5.

lgedeon:
Pass it into prepare_excerpt_response() so that it can pass it on.


rachelbaker:
@lgedeon Thank you!  Merged.

PR #2556, opened 2016-06-26
Define type as a string, not an array

danielbachhuber:
From #2429


danielbachhuber:
@BE-Webdesign ^^ FYI

BE-Webdesign:
Thank you for fixing that.  I left the array() container because it used to be two different values woops.

PR #2571, opened 2016-06-30
#2516 change permissions check to match core

### LINKED ISSUE #2516: Uploading an attachment should not require the edit_posts capability
Currently WP_REST_Attachments_Controller::create_item_permissions_check() checks $post_type_obj->cap->create_posts which returns the attachment-specifc "upload_files", but also $post_type_obj->cap->edit_posts which returns the general "edit_posts".

I suggest that edit_posts should not be required if the user can upload_files, as there is then no way via map_meta_cap to modify the API to allow a Subscriber (for example) to be able to submit attachments with the intention of allowing them to use a second API call to attach the attachment to a custom post type that is not 'post' (where that post type is registered using capability_type).

Unless there's a way I'm missing of submitting the attachment upload in such a way that satisfies this constraint?

nullvariable:
update permissions check to match core. Removes checks for create/edit_post permissions in favor of upload_files. Should resolve #2516

### DIFF
--- a/tests/test-rest-attachments-controller.php
+++ b/tests/test-rest-attachments-controller.php
@@ -20,6 +20,17 @@ class WP_Test_REST_Attachments_Controller extends WP_Test_REST_Post_Type_Control
 		$this->contributor_id = $this->factory->user->create( array(
 			'role' => 'contributor',
 		) );
+		//matches core capabilities for subscriber role and adds upload_files
+		_x('File upload role', 'User role');
+		add_role('uploader', 'File upload role');
+		$role = get_role('uploader');
+		$role->add_cap('upload_files');
+		$role->add_cap('read');
+		$role->add_cap('level_0');
+		//print_r(array('role'=>$role));
+		$this->uploader_id = $this->factory->user->create( array(
+			'role' => 'uploader',
+		) );
 
 		$orig_file = dirname( __FILE__ ) . '/data/canola.jpg';
 		$this->test_file = '/tmp/canola.jpg';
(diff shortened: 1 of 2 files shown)


danielbachhuber:
> Removes checks for create/edit_post permissions in favor of upload_files.

Can you make sure there's a test case included for the scenario outlined on the original issue?

nullvariable:
@danielbachhuber sure, the existing test case seemed to work well, it tests if a subscriber can upload (a role which does not have upload_files), there is also a test case for uploads working for a user with edit_posts, which in a default role setup overlaps with admin/editor/author being the only roles with upload_files.

So the test I would write would add a new role with only the upload files capability and then test if that role can upload a file, does that sound like what you're looking for?

danielbachhuber:
> So the test I would write would add a new role with only the upload files capability and then test if that role can upload a file, does that sound like what you're looking for?

Yep

nullvariable:
@danielbachhuber that should do the trick

joehoyle:
This looks good to me, presuming this is core's behaviour which I haven't actually verified. Given updating the attachment probably does require `edit_posts`. I'd be interested in how cores handles this if the user only has `upload_file`. In the case of uploading with the REST API, you don't get to send any attachment data (such as title, description etc) in the initial `create` request as the body is used for the raw image data, thus `edit_post` is virtually always required to create an attachment.

nullvariable:
@joehoyle I reviewed all the places that I could find in core here: https://github.com/WP-API/WP-API/issues/2516#issuecomment-229113511

as best I could tell with core you can create an attachment without `edit_posts`

tests/test-rest-attachments-controller.php
```diff
@@ -20,6 +20,17 @@ public function setUp() {
 		$this->contributor_id = $this->factory->user->create( array(
 			'role' => 'contributor',
 		) );
+		//matches core capabilities for subscriber role and adds upload_files
+		_x( 'File upload role', 'User role' );
```
joehoyle, line 24: What's this thing doing here? :) Don't think it's needed?

nullvariable, line 24: well 'File upload role' in line 25 should be translated but doesn't need to be for tests I guess

joehoyle, line 24: Right, but you are not assigning it to anything, no?

nullvariable, line 24: Correct. We'd either need to add a variable and pass it to the add_role function or just remove it since our goal here isn't really to test translations. I can modify it either way, but I'd lean towards removing it since it's not the goal of the test and won't really have an impact either way.

kadamwhite:
From PR review today, these are the outstanding TODOs here:
- [ ] Verify this is how Core actually works
- [ ] Clean up the test cases

@nullvariable we're moving this to needs-refresh ([discussion](https://wordpress.slack.com/archives/core-restapi/p1473799219001090))

adamsilverstein:
@nullvariable Thanks for your contribution here. I verified this is actually how core works (as well as that is possible given the variance) and created a refreshed patch (cleaning up the unit test a bit) in https://github.com/WP-API/WP-API/pull/2743.

PR #2572, opened 2016-06-30
Fixed typo on line 120

hyldmo:


BE-Webdesign:
Thanks for catching that @eivhyl. This is part of the client-js library so you will want to do the PR on that repo.  Here is a link to the file you would need to edit https://github.com/WP-API/client-js/blob/master/js/utils.js

hyldmo:
My bad! I'll close this and re-open it there. Thanks for notifying me

Issue #2554, opened 2016-06-23
Posts results order


agustina-biscayart-movile:
Hi there!

I'm usign the parameter "search" to get those posts that matchs with specific term. I've seen that there are this options to order the results by:

![image](https://cloud.githubusercontent.com/assets/12632582/16305990/8d35329e-3932-11e6-89b2-8f22f5e99dbb.png)

Do you know if there is a way to order the result by "**best match**"?. For example, if I look in my WordPress site by "pizza" and a I have 3 documents whit that word: one with _pizza_ in the tittle, other with _pizza_ in the subtitle and other with the word _pizza_ just as part of the text, the site returns the results exactly in that way. But using this plugin, I get the results not always in that order, it uses the default that is "date".

Do you know if there is something that I can do to change this?

Thanks!

almirbi:
Hello there,

can you show us how are your posts being rendered on your site, without the REST API?

Cheers

websupporter:
Hi,
sorry for answering so late. I spoke with @rmccue while WCEU. Basically, we are missing the WP build in relevance search for the orderby parameter. I tried it with such a query, and I think this is what you are looking for.

I will do a PR and hope it will be part of the WordPress REST API soon.

If it would be part of the API you could perform a request like this:
`http://local.wordpress-trunk.dev/wp-json/wp/v2/posts/?search=pizza&orderby=relevance`

websupporter:
I was digging a bit around in the discussions around orderby, since I was not to sure, maybe I would find something which is a stopper for adding the relevance as a orderby value. So far I haven't found anything, but something else. If you do not want to wait, or if this PR will not be part of the plugin, there is a workaround, how to add `relevance` to the allowed orderby values. See here: https://github.com/WP-API/WP-API/issues/2265

rmccue:
This was added in #2579.

Issue #2265, opened 2016-02-11
Custom orderby param value


Youdaman:
Hi, I've got some logic in my pre_user_query and pre_get_posts that lets me use a custom value with the orderby query param.

In the latest update there's a breaking change that limits the valid request args to an enum set, and I understand I have to use rest_validate_request_arg somehow to allow my custom orderby value.

I don't know where/how to do this however, and if you could point me to the right filter or snippet that lets me merge my value with the defaults that would be great!

Youdaman:
I managed to work out how to slot in my custom orderby param ("honey") as follows:

``` php
add_filter('rest_endpoints', 'my_modify_rest_routes');
function my_modify_rest_routes( $routes ) {
  array_push( $routes['/wp/v2/topic'][0]['args']['orderby']['enum'], 'honey' );
  array_push( $routes['/wp/v2/users'][0]['args']['orderby']['enum'], 'honey' );
  //error_log('routes = ' . print_r($routes, true));
  return $routes;
}
```

Hope this helps anyone else having the same issue since the breaking change.

danielbachhuber:
> In the latest update there's a breaking change that limits the valid request args to an enum set, and I understand I have to use rest_validate_request_arg somehow to allow my custom orderby value.

Sorry about that :( It skipped my mind that anyone would be using a custom orderby value.

I'll make sure we get the "extending routes" aspect documented https://github.com/WP-API/docs-v2/issues/114

PR #2579, opened 2016-07-02
Adding relevance orderby

websupporter:
Solves https://github.com/WP-API/WP-API/issues/2554

I add `relevance` as an orderby value, so we can use the relevance orderby build in the WP_Query.
I've extended also the unit test to test the relevance orderby.

### DIFF
--- a/lib/endpoints/class-wp-rest-posts-controller.php
+++ b/lib/endpoints/class-wp-rest-posts-controller.php
@@ -71,10 +71,6 @@ class WP_REST_Posts_Controller extends WP_REST_Controller {
 
 		$post_type = get_post_type_object( $this->post_type );
 
-		if ( ! empty( $request['orderby'] ) && 'relevance' === $request['orderby'] && empty( $request['search'] ) ) {
-			return new WP_Error( 'rest_no_search_term_defined', __( 'You need to define a search term in order to use the relevance search.' ), array( 'status' => 400 ) );
-		}
-
 		if ( 'edit' === $request['context'] && ! current_user_can( $post_type->cap->edit_posts ) ) {
 			return new WP_Error( 'rest_forbidden_context', __( 'Sorry, you are not allowed to edit these posts in this post type' ), array( 'status' => rest_authorization_required_code() ) );
 		}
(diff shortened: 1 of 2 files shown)


rachelbaker:
Isn't use of the `relevance` orderby limited to collections with a search parameter?  What is the expected result if I use `orderby=releveance` without a search query/

websupporter:
Hi @rachelbaker,
ah right, here is the possible stopper. Yes, `relevance` is a bit out of line. It works only together with the `search`-parameter (https://core.trac.wordpress.org/browser/tags/4.5.3/src/wp-includes/query.php#L3004) and would be ignored otherwise.

I don't know. I understand if this is a stopper. On the other side, I think to be able to use the WP relevance search without a workaround (you could do something like this: https://github.com/WP-API/WP-API/issues/2265) would be a good enhancement of the REST API.

But maybe we could enhance the `get_items_permissions_check()` and check if its a relevance search without a search query to throw an error? Would this be a solution? Or should we drop it entirely?

Thanks for having a look.

rachelbaker:
@websupporter I think returning an error if `relevance` is used as the `orderby` param without a `search` value.  It would probably also make sense to default to `relevance` as the `orderby` param if there is a search query.

rachelbaker:
Related #2407

websupporter:
Hi @rachelbaker,
sorry for the late reply. I've added a permission check.

But I am not sure how I could set `orderby` to `relevance` by default in case a search term is given. The only way I could see would be in the `get_items()`. But it would need something like `if ( empty( $_REQUEST['orderby'] ) )` and it just looks wrong to me.

I think, the default should be `date` as we tell the users in the collection params, but maybe you had something else in mind?

websupporter:
Will check this error these days. Smashed my Virtualbox today somehow. Something related to Kernel or whatever. Try to get it work as soon as possible.

lib/endpoints/class-wp-rest-posts-controller.php
```diff
@@ -71,6 +71,10 @@ public function get_items_permissions_check( $request ) {
 
 		$post_type = get_post_type_object( $this->post_type );
 
+		if ( ! empty( $request['orderby'] ) && $request['orderby'] === 'relevance' && empty( $request['search'] ) ) {
```
joehoyle: @websupporter I think this should be in `get_items` as this is a logic check, not to do with permissions.

joehoyle:
I'm not sure if I like defaulting to a different value for orderby is `search` is specific. It will break the `default` we are saying it is in the schema.

websupporter:
@joehoyle moved the check in the `get_items()` "as this is a logic check, not to do with permissions."

I would agree on the schema problem and wouldn't like to break this either.

joehoyle:
Thanks!

PR #2594, opened 2016-07-13
Add initial Sites controller

dancrumb:
Adds the initial Sites controller for /sites REST endpoints. Lots of further development needed, but this gets the ball rolling

### DIFF
--- a/lib/endpoints/class-wp-rest-sites-controller.php
+++ b/lib/endpoints/class-wp-rest-sites-controller.php
@@ -1,0 +1,713 @@
+<?php
+
+/**
+ * Access users
+ */
+class WP_REST_Sites_Controller extends WP_REST_Controller {
+
+	public function __construct() {
+		$this->namespace = 'wp/v2';
+		$this->rest_base = 'sites';
+	}
+
+	/**
+	 * Register the routes for the objects of the controller.
+	 */
+	public function register_routes() {
+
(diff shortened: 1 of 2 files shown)


BE-Webdesign:
Hi @dancrumb. Thank you for the Pull Request.

I am not sure if you are aware of the [site endpoint](https://github.com/WP-API/wp-api-site-endpoints).  It is somewhat hidden and easily missable.  If you would like to work on the site endpoint for the WP REST API, that is where you would want to add your Pull Requests.  Looking forward to your contributions!

dancrumb:
Happy to do that, but can someone help me understand why it's its own repo and not a branch off of this one?

Also, how is one supposed to use it, since it _feels_ like it is actually a subset of this plugin and not a plugin in its own right.

BE-Webdesign:
The site endpoint is in a separate plugin mostly because the "core endpoints" (post, comments, users, taxonomies etc.) are in the process of being considered for core at some point hopefully in the near future.  Other endpoints which are not as far along in development have been split out into separate plugins in the repository.  It is unclear at this time whether they will be merged into the WP REST API plugin before its merge with WordPress core.  As the additional endpoints are deemed to be more stable and usable they will either be merged into core after WP REST API plugin is merged or they will be merged into the WP REST API plugin before it is merged into core.

PR #2605, opened 2016-07-19
Use VIP cached function to get_term_link

jasonbahl:
During a VIP GO code review, we were asked to update our codebase to ensure we're using cached functions. We've updated our codebase and are working on submitting PR's to codebases we don't directly control.


rmccue:
The API is intended to be merged into WP core, so this isn't the right place for this change to go. Either the core `get_term_link()` needs to be patched to allow shortcircuiting (which would work for all of VIP without modifications) or VIP needs to approach this in a different way.

Thanks for the PR!

PR #2615, opened 2016-07-25
Fix broken `is_bool()` check on boolean request args

rachelbaker:
In `rest_validate_request_arg()` we cannot use the `! is_bool()` conditional to validate if a request argument is either `true` or `false`.  The query string in the url (example: https://demo.wp-api.org/wp-json/wp/v2/categories?hide_empty=true) will be parsed as a string.  Which means you will always get the error: 
`{
code: "rest_invalid_param",
message: "Invalid parameter(s): hide_empty",
data: {
status: 400,
params: {
hide_empty: "hide_empty is not of type boolean"
}
}
}`

This PR changes the conditional in `rest_validate_request_args()` to use a conditional that is much less strict and accepts query string parameters correctly.


rachelbaker:
@WP-API/amigos #reviewmerge

rachelbaker:
Ugghh.. going to close this PR and try a different approach.

PR #2630, opened 2016-07-31
Fix broken `is_bool()` check on boolean request args

### LINKED ISSUE #2616: Boolean request args fail validation
In rest_validate_request_arg() we cannot use the `! is_bool()` conditional to validate if a request argument is either `true` or `false`. The query string in the url (example: https://demo.wp-api.org/wp-json/wp/v2/categories?hide_empty=true) will be parsed as a string. Which means you will always get the error: 
{ code: "rest_invalid_param", message: "Invalid parameter(s): hide_empty", data: { status: 400, params: { hide_empty: "hide_empty is not of type boolean" } } }

rachelbaker:
Fixes #2616 

In `rest_validate_request_arg()` we cannot use the `! is_bool()`conditional to validate if a request argument is either true or false. The query string in the url (example: https://demo.wp-api.org/wp-json/wp/v2/categories?hide_empty=true) will be parsed as a string. Which means you will always get the error: 

```
{ code: "rest_invalid_param", message: "Invalid parameter(s): hide_empty", data: { status: 400, params: { hide_empty: "hide_empty is not of type boolean" } } }
```

This PR introduces the new function `rest_is_boolean` that will allow us to validate boolean strings in API request parameters.  It also changes the conditional in rest_validate_request_args() to use `rest_is_boolean` for a much less strict (and accurate) validation of boolean-type parameters.

### DIFF
--- a/plugin.php
+++ b/plugin.php
@@ -299,8 +299,8 @@ if ( ! function_exists( 'rest_validate_request_arg' ) ) {
 			return new WP_Error( 'rest_invalid_param', sprintf( __( '%s is not of type %s' ), $param, 'integer' ) );
 		}
 
-		if ( 'boolean' === $args['type'] && ! is_bool( $value ) ) {
-			return new WP_Error( 'rest_invalid_param', sprintf( __( '%s is not of type %s' ), $param, 'boolean' ) );
+		if ( 'boolean' === $args['type'] && ! rest_is_boolean( $value ) ) {
+				return new WP_Error( 'rest_invalid_param', sprintf( __( '%s is not of type %s' ), $value, 'boolean' ) );
 		}
 
 		if ( 'string' === $args['type'] && ! is_string( $value ) ) {
@@ -428,5 +428,36 @@ if ( ! function_exists( 'rest_is_ip_address' ) ) {
 		}
 
 		return $ipv4;
+	}
+}
+
+/**
+ * Determines if a given value is boolean-like.
+ *
+ * @param bool|string $maybe_bool The value being evaluated.
+ * @return boolean True if a boolean, otherwise false.
+ */
+if ( ! function_exists( 'rest_is_boolean' ) ) {
+	function rest_is_boolean( $maybe_bool ) {
+		if ( is_bool( $maybe_bool ) ) {
+			return true;
+		}
+
+		if ( is_string( $maybe_bool ) ) {
+			$maybe_bool = strtolower( $maybe_bool );
+
+			$valid_boolean_values = array(
+				'false',
+				'true',
+				'0',
+				'1',
+			);
+
+			if ( in_array( $maybe_bool, $valid_boolean_values, true ) ) {
+				return true;
+			}
+		}
+
+		return false;
 	}
 }
--- a/tests/test-rest-controller.php
+++ b/tests/test-rest-controller.php
@@ -52,9 +52,22 @@ class WP_Test_REST_Controller extends WP_Test_REST_TestCase {
 			rest_validate_request_arg( false, $this->request, 'someboolean' )
 		);
 
+		$this->assertTrue(
+			rest_validate_request_arg( 'true', $this->request, 'someboolean' )
+		);
+		$this->assertTrue(
+			rest_validate_request_arg( 'TRUE', $this->request, 'someboolean' )
+		);
+		$this->assertTrue(
+			rest_validate_request_arg( 'false', $this->request, 'someboolean' )
+		);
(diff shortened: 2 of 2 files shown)


rachelbaker:
@WP-API/amigos ##reviewmerge

websupporter:
I think, we just moved the problem a bit further.

If we do `?hide_empty=false`

We will get 
`if ( $prepared_args['hide_empty'] ) echo "hello"; //echos hello although set to false`

what is not intended and finally in `get_terms()` it is only checked
`
    if ( $args['hide_empty'] && !$hierarchical ) {
        $where .= ' AND tt.count > 0';
    }
`

so basically, it would add the `AND tt.count > 0` because the string `"false"` is `true`.

I think, if we loosen the `is_bool()` check we need to sanitize afterwards in `get_items()`.This was my initial idea, why I thought a function with a "tripod"-return (`true`|`false`|`WP_Error`) might be helpful, so we wouldn't need a function to validate and a function to sanitize.

I want to go these days through the plugin and see, what other booleans we have in the schemas, because we would need to go through all of them, right?

BE-Webdesign:
I like this PR.  Could we add test cases for each of the four strings, or is that not necessary?

When using `WP_REST_Controller::get_endpoint_args_for_item_schema()`, is there a potential case where accepting such a wide array of values would be a bad thing?

I would suggest strictly sanitizing the value to a true boolean, which can then be validated with `is_bool()`.  That way a lot of the logic in the endpoint can act on strict comparison instead of loose.  You can put the logic in `rest_sanitize_request_arg()` and from there on out the rest of the code can act as though the value is indeed a real boolean.

> so basically, it would add the AND tt.count > 0 because the string "false" is true.

This is the main reason why I believe we would want a sanitization function to take loose values and conform them to a strict boolean value.

> I want to go these days through the plugin and see, what other booleans we have in the schemas, because we would need to go through all of them, right?

I think we would definitely want to do this.  We also need to do this with arrays at some point.

joehoyle:
I _think_ should be a sanitization transformation on the property rather than validity. Having it so would also address @websupporter concern about 'false' still being in the request object's data. (string) "false" is a quirk of URL params, so the sooner we can make it a "real" boolean the better. So the flow would be: 
1. Send `hide_emtpy=false`
2. Sanitize_callback `rest_boolval` => "false" => (bool) false
3. Validate callback `is_bool` (as usual) => true

rachelbaker:
@joehoyle 

> I think should be a sanitization transformation on the property rather than validity.

Right now we don't have any sanitization, and when we do add it the sanitization would run after the validation.  I agree, we should correctly set the type in the sanitization function.  However, if we ONLY sanitize the value folks can end up with unexpected results due to how booleans are cast: http://php.net/manual/en/function.boolval.php

I think we should do both.

rachelbaker:
@websupporter @BE-Webdesign I agree we need to sanitize the value as well, filed in #2633. I originally was hoping to keep that a separate (while still VERY related) issue.  That does not sound like it is possible due to the inconsistencies of our conditional checks within the different endpoints.

websupporter:
@joehoyle 
validation is done earlier than sanitization. I was wondering myself, was introduced some weeks ago:
https://core.trac.wordpress.org/changeset/37943
https://core.trac.wordpress.org/ticket/37192

joehoyle:
@websupporter ahh I hadn't noticed that - I added a comment to https://core.trac.wordpress.org/ticket/37192 with my thoughts. Personally I think that was a mistake, but further discussion on that should probably go on https://core.trac.wordpress.org/ticket/37192

BE-Webdesign:
Should we merge this when sanitize PR is ready?

joehoyle:
@rachelbaker do you think we should support `(int) 0` / `(int) 1` here? URL params are not the only way to get data in, a JSON blob could be passed in, and I don't think the `rest_is_bool` function can handle ints?

plugin.php
```diff
@@ -430,3 +430,34 @@ function rest_is_ip_address( $ipv4 ) {
...
+			$valid_boolean_values = array(
+				'false',
+				'true',
+				'0',
+				'1',
+			);
+
+			if ( in_array( $maybe_bool, $valid_boolean_values, true ) ) {
+				return true;
+			}
```
rmccue, line 458: This should just `return in_array( ...` since we don't want to fall-through.

kadamwhite:
Labeled `needs refresh` to account for integers, @BE-Webdesign is going to work on updating the patch.

BE-Webdesign:
This was added in #2704.

PR #2641, opened 2016-08-05
Improve WP_REST_Controller::filter_response_by_context().

tfrommen:
This pull request improves the `WP_REST_Controller::filter_response_by_context()` method by adding a `continue;` after which the only thing that _could_ be done is to delete stuff that we just deleted. ;) No need for all the unnecessary stuff.


joehoyle:
Makes sense, thanks @tfrommen!

Issue #2642, opened 2016-08-05
Show Private / Draft Posts via Javascript


wblaircox:
I'm unable to show private or draft posts via javascript. Using Postman has not worked, best I can determine it may have something to do with my server running PHP as CGI. But I can run my script from a WP page, which is fine for my usage. I've read the examples for filtering, but at this point I'm just trying to get filtering to work at all. Any suggestions?

function getpost() {
        var postsCollection = new wp.api.collections.Posts({
            data: {
                'filter': {
                    'category-slug': 'ChartData',
                    'posts_per_page': -1
                }
            }
        });
        postsCollection.fetch();
        console.log(postsCollection);
    };

BE-Webdesign:
Hi again,

You would need to use the status param to view private and draft posts.  Check out the [API Reference for the status param](http://v2.wp-api.org/reference/posts/).  You will need to be authenticated with the API and have proper permissions to view.

Going to close but we can continue discussion here.

wblaircox:
Thanks for the assistance. I've read this reference and tried a few times / variations. I must be doing something wrong. Some references online say it needs to be post_status. I started with the javascript example on the documentation page, then applied the parameter. This is run on a page within Wordpress, the user is admin. I just bring back the results to console.log for now. 

function getpost() {
        // Load an existing post
        var postsCollection = new wp.api.collections.Posts({
            status: 'draft'
        });
            postsCollection.fetch();

wblaircox:
You know, I'm not sure it will matter. The JSON this pulls back for the published posts is a pain in the ass. So many fields I don't need and trying to pull back variables in the structure is proving challenging. Methods I use continually elsewhere don't seem to work as expected with this. A bit of code I use to get the index array and pull everything back doesn't even work. Love the idea, but a week on this is enough for now. I need to have a working function, not a make work project. Maybe when I get more handy with Javascript.

kadamwhite:
@wblaircox If you want a lighter set of JSON, it's probably best to create your own [custom endpoint](http://v2.wp-api.org/extending/adding/) — but in the case of your specific question, we did _not_ document the "any" status, or even the full list of statuses available, so sincere apologies for that. I've opened a PR to add an enum of supported status types to the API, and for the time being using `status: 'any` ought to give you the full set you desire.

kadamwhite:
[...]

PR #2645, opened 2016-08-07
Document options of the "status" parameter for Post collection GETs

kadamwhite:
In #2642 we refer API users to the documentation for the `status` parameter, but that parameter does not actually document "any" or the other valid WP_Query Status options that can be passed to the API. Since retrieving a full list of posts regardless of status is a critical action within the context of an admin interface, we should explicitly list out the options accepted for `status`.

I believe this will do the trick but I am on the road and haven't been able to attempt regenerating the docs to validate that this goes through alright.


joehoyle:
I noticed this the other day too, thanks!

PR #2652, opened 2016-08-12
fixing invalid regex testing for model vs collection route

sbussetti:
The current regex actually tests for "any route ending with ), +, |, m, or e" which will incorrectly match any collection route as a model route if it ends with a literal "m" or "e".  This fixes that regex.

Side note:  I'm unclear from the repo how you prefer we rebuild the .min and .map but I'm happy to also do that if needed.

### DIFF
--- a/wp-api.js
+++ b/wp-api.js
@@ -1080,7 +1080,7 @@
 				) {
 
 					// Single items end with a regex (or the special case 'me').
-					if ( /.*[+)|me]$/.test( index ) ) {
+					if ( /.*([+)]|me)$/.test( index ) ) {
 						modelRoutes.push( { index: index, route: route } );
 					} else {
 


BE-Webdesign:
Hi @sbussetti,

Thank you for using the WP REST API and providing this PR.  I believe you will need to open it up on the [client-js repo](https://github.com/WP-API/client-js), as that is built and loaded into this; the main repo.

sbussetti:
Thanks! Will do.

~ Steve

On August 12, 2016 at 4:53:41 PM, ChopinBach (notifications@github.com) wrote:

Hi @sbussetti,

Thank you for using the WP REST API and providing this PR. I believe you will need to open it up on the client-js repo, as that is built and loaded into this; the main repo.

sbussetti:
Just checked.  This is already fixed on master in https://github.com/WP-API/client-js WP-API/client-js@f9d9a15 -- so I assume it just needs to be updated here?

joehoyle:
Yup we just need to sync it back to this repo, that is done when we release a new version of this plugin. Thanks for the PR, sorry it was redundant!

PR #2656, opened 2016-08-17
fix forum url and installer-name

torounit:
fix plugin slug for v2.


BE-Webdesign:
Hi @torounit,

Thank you for the PR.  I am not sure if someone has scripts attached to that data, which could potentially do bad things if we change it, so I will be leaving this PR untouched, until there is verification.  Most likely it is fine, and thank you for catching this.

joehoyle:
I think this is fine, thanks @torounit.

PR #2662, opened 2016-08-20
#1612 comment children links

### LINKED ISSUE #1612: How to implement comment threading using `GET /comments` ?

BE-Webdesign:
Fixes #1612,

This patch adds a `children` link to the response when the comment has children.  The link is to a collection of comments who share the same parent as the original comment resource.  This enables a client to then recursively move through the links, with pagination parameters to boot, to performantly collect all threaded comments of a single comment.

### DIFF
--- a/lib/endpoints/class-wp-rest-comments-controller.php
+++ b/lib/endpoints/class-wp-rest-comments-controller.php
@@ -640,6 +640,18 @@ class WP_REST_Comments_Controller extends WP_REST_Controller {
 			);
 		}
 
+		// Only grab one comment to verify the comment has children.
+		$comment_children = $comment->get_children( array( 'number' => 1 ) );
+		if ( ! empty( $comment_children ) ) {
+			$args = array( 'parent' => $comment->comment_ID );
+			$rest_url = add_query_arg( $args, rest_url( $this->namespace . $this->rest_base ) );
+
+			$links['children'] = array(
+				'href'       => $rest_url,
+				'embeddable' => true,
+			);
+		}
+
 		return $links;
 	}
 
--- a/lib/endpoints/class-wp-rest-comments-controller.php
+++ b/lib/endpoints/class-wp-rest-comments-controller.php
@@ -653,7 +653,7 @@ class WP_REST_Comments_Controller extends WP_REST_Controller {
 			$rest_url = add_query_arg( $args, rest_url( $this->namespace . '/' . $this->rest_base ) );
 
 			$links['children'] = array(
-				'href'       => $rest_url,
+				'href' => $rest_url,
 			);
 		}
 
--- a/lib/endpoints/class-wp-rest-comments-controller.php
+++ b/lib/endpoints/class-wp-rest-comments-controller.php
@@ -644,7 +644,7 @@ class WP_REST_Comments_Controller extends WP_REST_Controller {
 		$comment_children = $comment->get_children( array( 'number' => 1 ) );
 		if ( ! empty( $comment_children ) ) {
 			$args = array( 'parent' => $comment->comment_ID );
-			$rest_url = add_query_arg( $args, rest_url( $this->namespace . $this->rest_base ) );
+			$rest_url = add_query_arg( $args, rest_url( $this->namespace . '/' . $this->rest_base ) );
 
 			$links['children'] = array(
 				'href'       => $rest_url,
--- a/lib/endpoints/class-wp-rest-comments-controller.php
+++ b/lib/endpoints/class-wp-rest-comments-controller.php
@@ -654,7 +654,6 @@ class WP_REST_Comments_Controller extends WP_REST_Controller {
 
 			$links['children'] = array(
 				'href'       => $rest_url,
-				'embeddable' => true,
 			);
 		}
 
--- a/lib/endpoints/class-wp-rest-comments-controller.php
+++ b/lib/endpoints/class-wp-rest-comments-controller.php
@@ -647,7 +647,7 @@ class WP_REST_Comments_Controller extends WP_REST_Controller {
 		}
 
 		// Only grab one comment to verify the comment has children.
-		$comment_children = $comment->get_children( array( 'number' => 1 ) );
+		$comment_children = $comment->get_children( array( 'number' => 1, 'count' => true ) );
 		if ( ! empty( $comment_children ) ) {
 			$args = array( 'parent' => $comment->comment_ID );
 			$rest_url = add_query_arg( $args, rest_url( $this->namespace . '/' . $this->rest_base ) );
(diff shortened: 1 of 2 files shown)


BE-Webdesign:
The test cases are not ideal, but it would involve refactoring existing test cases.  I decided to just keep this PR clean and to the point of the issue.  I will open up a separate PR at some point to improve the test suite.  I believe an issue is already open for that and I will try to hunt it down.

lib/endpoints/class-wp-rest-comments-controller.php
```diff
@@ -640,6 +640,18 @@ protected function prepare_links( $comment ) {
 			);
 		}
 
+		// Only grab one comment to verify the comment has children.
+		$comment_children = $comment->get_children( array( 'number' => 1 ) );
```
kadamwhite: In code review @rmccue identified this to be a performance concern, as `get_children` runs a query every time that is not object-cached. Performance impact needs to be assessed on a site with a high volume of comments before this can be merged.

BE-Webdesign:
As noted in a Slack conversation today. Performance benchmarks will need to occur to ensure this is fast enough on higher scale installs.  Testing at 10k comments and 100k comments and looking at the differences.  Will be posted at some point soonish.

lib/endpoints/class-wp-rest-comments-controller.php
BE-Webdesign: I tested this on 1,100,000 comments, using Redis it dropped the response time back to normal.  The query is object cached, so to my understanding it should not be a big problem.  There is very little difference in response time for that magnitude of comments when the children are grabbed or not grabbed. On average over 100 requests each the difference was 2ms.

lib/endpoints/class-wp-rest-comments-controller.php
```diff
@@ -640,6 +640,18 @@ protected function prepare_links( $comment ) {
...
+		if ( ! empty( $comment_children ) ) {
+			$args = array( 'parent' => $comment->comment_ID );
+			$rest_url = add_query_arg( $args, rest_url( $this->namespace . $this->rest_base ) );
```
BE-Webdesign: This is not the right link it needs a `/` before `$this->rest_base`.

BE-Webdesign:
I tested this on 1,100,000 comments, using Redis it dropped the response time back to normal. The query is object cached, so to my understanding it should not be a big problem. There is very little difference in response time for that magnitude of comments, when the children are grabbed or not grabbed. On average over 100 requests for each one with comment children one without, the difference was 2ms.

lib/endpoints/class-wp-rest-comments-controller.php
rmccue: Since we only care about the count, we should pass `count => true`, which queries for `COUNT(*)` instead of the fields. This is less expensive when we actually need to hit the database, and should be equivalent performance for the cached case.

rmccue (changes requested):

lib/endpoints/class-wp-rest-comments-controller.php
BE-Webdesign: Sounds good, I will check that out.

BE-Webdesign: That looks like it will save even more execution time and the cache will be a lot smaller 👍! Thank you for the tip.

BE-Webdesign:
I added in the count parameter so the object cache will be a lot smaller.

lib/endpoints/class-wp-rest-comments-controller.php
```diff
@@ -646,6 +646,18 @@ protected function prepare_links( $comment ) {
...
 
+		// Only grab one comment to verify the comment has children.
+		$comment_children = $comment->get_children( array( 'number' => 1, 'count' => true ) );
+		if ( ! empty( $comment_children ) ) {
+			$args = array( 'parent' => $comment->comment_ID );
+			$rest_url = add_query_arg( $args, rest_url( $this->namespace . '/' . $this->rest_base ) );
+
+			$links['children'] = array(
+				'href'       => $rest_url,
+				'embeddable' => true,
```
joehoyle: Hmm I'm not sure `embeddable =>true` is a good idea, typically we don't have potentially large lists as `embeddable => true` for example, posts / users lists.

joehoyle (changes requested):

lib/endpoints/class-wp-rest-comments-controller.php
BE-Webdesign: Sounds good, I will change it. The original ticket I believe wanted it to be embeddable but it doesn't make much sense to embed a large collection, especially if we make a batch endpoint soon.

lib/endpoints/class-wp-rest-comments-controller.php
```diff
@@ -646,6 +646,17 @@ protected function prepare_links( $comment ) {
...
 		}
 
+		// Only grab one comment to verify the comment has children.
+		$comment_children = $comment->get_children( array( 'number' => 1, 'count' => true ) );
+		if ( ! empty( $comment_children ) ) {
+			$args = array( 'parent' => $comment->comment_ID );
+			$rest_url = add_query_arg( $args, rest_url( $this->namespace . '/' . $this->rest_base ) );
+
+			$links['children'] = array(
+				'href'       => $rest_url,
```
rachelbaker: @BE-Webdesign indentation is off after https://github.com/WP-API/WP-API/pull/2662/commits/b10fb736dff1ecb7addad4c91f24fccb6986c8a6

rachelbaker (changes requested):
Minor indentation fix needed.

joehoyle (approved):

joehoyle:
@BE-Webdesign any update here? would be nice to get this out today!

BE-Webdesign:
I fixed the indentation, removed its embedability, and changed the query to only run the count for performance reasons.  It should be good to go. The PR still needs the thumbs up from @rmccue and @rachelbaker.  If it is decided that this feature is not needed, then feel free to close the PR.

rmccue (approved):

rachelbaker (approved):

Issue #2664, opened 2016-08-21
Allow relevance search with filter[s]


websupporter:
In a discussion here in the forum I've realized that the check for relevance only takes `search` into account and not the possibility to send the search string via `filter[s]`. See https://github.com/WP-API/WP-API/pull/2579/commits/417b29066aaccc2017a1616cb8d931be2c090ef2

I've just creating a PR fixing this.

PR #2665, opened 2016-08-21
add filter[s] to allow for relevance search

websupporter:
Fixes https://github.com/WP-API/WP-API/issues/2664

### DIFF
--- a/lib/endpoints/class-wp-rest-posts-controller.php
+++ b/lib/endpoints/class-wp-rest-posts-controller.php
@@ -88,7 +88,7 @@ class WP_REST_Posts_Controller extends WP_REST_Controller {
 
 		//Make sure a search string is set in case the orderby is set to 'relevace'
 		if ( ! empty( $request['orderby'] ) && 'relevance' === $request['orderby'] && empty( $request['search'] ) && empty( $request['filter']['s'] ) ) {
-			return new WP_Error( 'rest_no_search_term_defined', __( 'You need to define a search term in order to use the relevance search.' ), array( 'status' => 400 ) );
+			return new WP_Error( 'rest_no_search_term_defined', __( 'You need to define a search term to order by relevance.' ), array( 'status' => 400 ) );
 		}
 
 		$args                         = array();
--- a/lib/endpoints/class-wp-rest-posts-controller.php
+++ b/lib/endpoints/class-wp-rest-posts-controller.php
@@ -87,7 +87,7 @@ class WP_REST_Posts_Controller extends WP_REST_Controller {
 	public function get_items( $request ) {
 
 		//Make sure a search string is set in case the orderby is set to 'relevace'
-		if ( ! empty( $request['orderby'] ) && 'relevance' === $request['orderby'] && empty( $request['search'] ) ) {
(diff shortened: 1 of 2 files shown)


BE-Webdesign:
I think with that conditional it will error if you were to do `?search=happy&orderby=relevance`.  I believe you would want `&& ( empty( $request['search'] ) || empty( $request['filter']['s'] ) )`

Can you add test cases to verify the error response works as intended?

websupporter:
Hi @BE-Webdesign,
will do. I thought it wouldn't be necessary, but its better I guess.

But basically I want both to be empty in order to say, you can't do a relevance search without a search string.

websupporter:
Added the test case. When I first wrote a test case I thought, I just add it to `test_get_items_order_and_orderby()`, but now it becomes maybe a bit messy. I could split the tests and open up a new `test_get_items_with_orderby_relevance()`?

---

Edit: There we might want also to add a failing search?

BE-Webdesign:
> But basically I want both to be empty in order to say, you can't do a relevance search without a search string.

Okay never mind misread the code then, the conditional is good.

Separate tests would be preferred.  I think all you need for a test case is to test whether the error response returns, when you have `orderby = relevance` and no `search` parameter specified.

lib/endpoints/class-wp-rest-posts-controller.php
```diff
@@ -87,7 +87,7 @@ public function get_items_permissions_check( $request ) {
 	public function get_items( $request ) {
 
 		//Make sure a search string is set in case the orderby is set to 'relevace'
-		if ( ! empty( $request['orderby'] ) && 'relevance' === $request['orderby'] && empty( $request['search'] ) ) {
+		if ( ! empty( $request['orderby'] ) && 'relevance' === $request['orderby'] && empty( $request['search'] ) && empty( $request['filter']['s'] ) ) {
 			return new WP_Error( 'rest_no_search_term_defined', __( 'You need to define a search term in order to use the relevance search.' ), array( 'status' => 400 ) );
```
BE-Webdesign: Could we change the error message to `'You need to define a search term to order by relevance.'`? Not a huge deal, but that would be a bit more clear to me. If you feel the current message is better, just leave it as is.

tests/test-rest-posts-controller.php
```diff
@@ -279,6 +279,13 @@ public function test_get_items_order_and_orderby() {
...
 		$data = $response->get_data();
 		$this->assertEquals( 'Coffee Cake is relevant', $data[0]['title']['rendered'] );
+
+		//Test relevance with filter[s] parameter
+		$request->set_param( 'filter', array( 's' => 'relevant' ) );
+		$request->set_param( 'orderby', 'relevance' );
+		$response = $this->server->dispatch( $request );
+		$data = $response->get_data();
+		$this->assertEquals( 'Coffee Cake is relevant', $data[0]['title']['rendered'] );
 	}
```
BE-Webdesign, line 283: All we need is for a test case is to trigger the error response not the other way around, unless that is already in a test somewhere from your previous PR.  So `request->set_param( 'orderby', 'relevance' );`, then dispatch and use `$this->assertErrorResponse()`.  Also if we could put this into its own test function, rather than the existing one, it might be better because it is testing a different concept.

BE-Webdesign:
Looks good to me.   I can't reset the Scrutinizer checks but going to merge it anyways as the unit tests are good.

PR #2684, opened 2016-09-06
Alter default comment sort order to be "desc"

### LINKED ISSUE #2588: comments collection ordering: Why is default "asc"?
for `/posts` collections, the most recent articles are the first ones returned: you step back through pages to step back through time. for `/comments`, the oldest comments are shown first, as the default ordering for the comments collection is `asc`. This feels inconsistent, and I wanted to raise the question to see whether all collections should follow the same default ordering by date.

kadamwhite:
This follows discussion in #2588 and [in slack](https://wordpress.slack.com/archives/core-restapi/p1473088762000227) around standardizing the sort order of the comments endpoint to bring it in line with both the other endpoints in the API, and with `WP_Comment_Query`, by switching back to "desc" as the default sort order. Note that is a breaking change if API consumers are relying on comment order to be ascending by default.

Closes #2588

### DIFF
--- a/lib/endpoints/class-wp-rest-comments-controller.php
+++ b/lib/endpoints/class-wp-rest-comments-controller.php
@@ -1006,7 +1006,7 @@ class WP_REST_Comments_Controller extends WP_REST_Controller {
 			'type'                  => 'string',
 			'sanitize_callback'     => 'sanitize_key',
 			'validate_callback'     => 'rest_validate_request_arg',
-			'default'               => 'asc',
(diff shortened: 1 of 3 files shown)


kadamwhite:
@WP-API/amigos #reviewmerge

BE-Webdesign:
I would hit merge, but I will defer to someone else in case they have a last vote against this.  The PR looks good to me!

danielbachhuber:
🚢

Issue #2670, opened 2016-08-24
Revision API content field is not rendered


pdufour:
Hoping to get some clarification on this, the revisions API returns "raw" content, and does not apply `the_content` filter to it. Is this a bug? The rest of the API like the posts api returns rendered content. So for instance `<p>` tags will get added.

joehoyle:
Hmm good question, I'm not totally sure why this is the case, will look into it.

kadamwhite:
@pdufour Can you provide an example of the sort of object you're getting back? Most objects, such as posts, will provide the content field as an object with both "raw" and "rendered" strings. "raw" contains the content as-authored, before filters; "rendered" will be post-the_content. Raw content is usually only accessible when in the "edit" context (`?context=edit`) and when authenticated.

Are you seeing "raw" data come back without the corresponding "rendered" key?

adamsilverstein:
I think this is similar to what we have in core currently: revisions are generally retrievable only in their 'raw' form - without <p> tags or shortcode processing. This may be due in part to potential pitfalls of attempting to render a revision where the render might depend on some (non-revisioned) meta data. 

Still this does seem like it would be super useful as a feature and worth considering adding.

pdufour:
There is no rendered field for the revisions API. The object looks like this:

```
"author": "1",
"date": null,
"date_gmt": null,
"guid": "http://localhost/1-autosave-v1/",
"id": 1,
"modified": "2015-09-06T15:01:42",
"modified_gmt": "2015-09-06T22:01:42",
"parent": 1,
"slug": "1-autosave-v1",
"title": "",
"content": "unrendered content"
```

I believe v1 provided the rendered content so is there a reason it changed in v2? Not contesting the decision, just trying to figure out if it's a bug or intended behavior.

websupporter:
I can confirm this behavior. Right now we have `title` (string), `content` (string), `excerpt` (string). 

Will align this to the usual output

```
title:{
raw:(string),
rendered:(string)
}
```

etc.

BE-Webdesign:
This was added in #2693.

PR #2693, opened 2016-09-13
Add raw and rendered to schema

websupporter:
Solves https://github.com/WP-API/WP-API/issues/2670

For usual posts we give the title, the excerpt, the guid and the content in two formats: `raw` and `rendered`. In case of revisions, we just gave the `raw` content. This PR extends the schema, which acts now like we know it from usual posts.

So, instead of `content` being a string it would now be an object:

```
content : {
     raw : "",
     rendered: "",
}
```

The `raw` property would only be send when `context` is set to `edit`.

### DIFF
--- a/lib/endpoints/class-wp-rest-revisions-controller.php
+++ b/lib/endpoints/class-wp-rest-revisions-controller.php
@@ -180,31 +180,80 @@ class WP_REST_Revisions_Controller extends WP_REST_Controller {
 	 */
 	public function prepare_item_for_response( $post, $request ) {
 
-		// Base fields for every post
-		$data = array(
-			'author'       => $post->post_author,
-			'date'         => $this->prepare_date_response( $post->post_date_gmt, $post->post_date ),
-			'date_gmt'     => $this->prepare_date_response( $post->post_date_gmt ),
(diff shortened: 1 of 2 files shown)


websupporter:
I'll check the password protected. Not quite sure if its really needed. If its needed, the code is missing the function.

websupporter:
I've made some tests. When I have a password protected post, the password is not written in the database for the revisions. I've removed the password checks.

BE-Webdesign:
Looks good to me! Going to do one last look through, I am not sure the consensus on whether to add this feature but +1 from me.  It makes for better potential authoring experiences.

websupporter:
While I was working on it I was thinking of a small slider for loggedin authors in the front end where they would be able to slide back in history and see their revisions rendered in the frontend. Small toy, but hey. Wouldn't work without the `rendered` properties.

kadamwhite:
:+1: from @rmccue, myself & @BE-Webdesign. There is an open question [voiced by Ryan](https://wordpress.slack.com/archives/core-restapi/p1473797676000986) around whether this change moves Revisions sufficiently close to Posts that they should just extend the posts controller. That will be addressed in a separate issue.

PR #2699, opened 2016-09-14
Force per_page to override the filter variable

### LINKED ISSUE #2357: `/posts/?filter[posts_per_page]=-1` returns all posts

rmccue:
Fixes #2357.


kadamwhite (approved):

kadamwhite:
Noting some overlap with #2359 that I'd missed previously; shouldn't block this going in, and might simplify #2359

rachelbaker (approved):

PR #2700, opened 2016-09-14
Update and add unit tests to account for ignoring filter[posts_per_page]

kadamwhite:
@joehoyle @rmccue fixed the test broken in #2699 (the test presumes that filter properties work consistently, which that PR breaks by ignoring posts_per_page entirely), but I am not sure how to handle the test for `filter[posts_per_page]=-1`: should I be testing that it is ignored, as seems to be the case, or should I be testing that it is an error?

The solution in #2699 is the quick fix but seems to introduce confusion around which `filter` properties are valid, and which aren't.

### DIFF
--- a/tests/test-rest-posts-controller.php
+++ b/tests/test-rest-posts-controller.php
@@ -464,11 +464,16 @@ class WP_Test_REST_Posts_Controller extends WP_Test_REST_Post_Type_Controller_Te
 		$this->assertErrorResponse( 'rest_invalid_param', $response, 400 );
 	}
 
-	public function test_get_items_invalid_filter_posts_per_page() {
+	public function test_get_items_invalid_posts_per_page_ignored() {
+		// This test ensures that filter[posts_per_page] is ignored, and that -1
+		// cannot be used to sidestep per_page's valid range to retrieve all posts
+		for ( $i = 0; $i < 20; $i++ ) {
+			$this->factory->post->create( array( 'post_status' => 'publish' ) );
+		}
 		$request = new WP_REST_Request( 'GET', '/wp/v2/posts' );
 		$request->set_query_params( array( 'filter' => array( 'posts_per_page' => -1 ) ) );
 		$response = $this->server->dispatch( $request );
-		$this->assertErrorResponse( 'rest_invalid_param', $response, 400 );
+		$this->assertCount( 10, $response->get_data() );
 	}
 
 	public function test_get_items_invalid_context() {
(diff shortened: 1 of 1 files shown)


tests/test-rest-posts-controller.php
```diff
@@ -420,19 +420,21 @@ public function test_get_items_pagination_headers() {
 		$this->assertContains( '<' . $prev_link . '>; rel="prev"', $headers['Link'] );
 		$this->assertFalse( stripos( $headers['Link'], 'rel="next"' ) );
 
-		// With filter params.
+		// With query params.
 		$request = new WP_REST_Request( 'GET', '/wp/v2/posts' );
-		$request->set_query_params( array( 'filter' => array( 'posts_per_page' => 5, 'paged' => 2 ) ) );
+		$request->set_query_params( array( 'per_page' => 5, 'page' => 2 ) );
```
kadamwhite, line 425: This was done because #2699 suppresses `filter[posts_per_page]` entirely, and we have first-party query parameters for pagination so it seems most future conscious to use them.

tests/test-rest-posts-controller.php
```diff
@@ -455,6 +457,20 @@ public function test_get_items_private_filter_query_var() {
...
+		$response = $this->server->dispatch( $request );
+		$this->assertErrorResponse( 'rest_invalid_param', $response, 400 );
+	}
+
+	public function test_get_items_invalid_filter_posts_per_page() {
+		$request = new WP_REST_Request( 'GET', '/wp/v2/posts' );
+		$request->set_query_params( array( 'filter' => array( 'posts_per_page' => -1 ) ) );
+		$response = $this->server->dispatch( $request );
+		$this->assertErrorResponse( 'rest_invalid_param', $response, 400 );
+	}
```
kadamwhite, line 477: This test is failing because it either represents an error condition we don't account for, or because it's the wrong test; hence the Q in the description

BE-Webdesign: I don't know a good way to error for that outside of validating the filter param, which has been decided against in past conversations.  When #2699 is merged, I would be +1 on changing the test to check if the response has greater than 1 and less than 10 posts, as that is the default `per_page amount`. Alternatively you could check the handler for the response to see what the request variables were and see if `per_page` was `10`.

kadamwhite: This is actually a PR against #2699 so I will make that change and (with Ryan's blessing) merge these in to that

BE-Webdesign: Oh yeah, I see now, didn't notice which branch this was going to, I will have to pay better attention to that :).

kadamwhite: @BE-Webdesign 

> Alternatively you could check the handler for the response to see what the request variables were and see if per_page was 10.

How can this be accomplished?

BE-Webdesign: @kadamwhite After reviewing how it would be done it is too convoluted to be useful, it would make one confusing test and that is bad.

PR #2702, opened 2016-09-14
Allow returning an error from field update callbacks

rmccue:
Right now, it's not possible to indicate an update failure in a custom callback. This adds the ability to return an error instance and have that passed back to the client.


joehoyle (approved):
Looks good to me, I guess we have no unit tests covering this case though?

rmccue:
> I guess we have no unit tests covering this case though?

Correct; you're welcome to write some 😉

nazari11:
[themeforest-5991392-edu-educational-courses-college-wp-theme-wordpress-theme.zip](https://github.com/WP-API/WP-API/files/473831/themeforest-5991392-edu-educational-courses-college-wp-theme-wordpress-theme.zip)

BE-Webdesign:
Tests would have been nice but YOLO.

Issue #2633, opened 2016-08-01
Sanitize boolean types in `rest_sanitize_request_arg()`


rachelbaker:
Related to discussions in #2630

websupporter:
Hi @rachelbaker,
since `boolval( 'false' );` would be `true` I think we would need here also a function, this time to sanitize. This function could then be placed in the `rest_sanitize_request_arg()`, which is executed after the validation.

Although we know already the value will be a boolean, I thought, it would be useful to also return a `WP_Error` in case its not, so the function could also be used in cases, where you can't be sure. So I smashed the function I've suggested in https://github.com/WP-API/WP-API/issues/2616 with your function `rest_is_boolean()`.

I searched a bit and finally I found, yes, to use strict mode for `in_array()` can be done for php 5.2, since it was introduced in 4.x (http://php.net/ChangeLog-4.php).

```
    if ( ! function_exists( 'rest_sanitize_boolean' ) ) {
        function rest_sanitize_boolean( $maybe_bool ) {
            if ( ! rest_is_boolean( $maybe_bool ) ) {
                return new WP_Error( 'no-bool', __( 'The parameter was no boolean.' ) );
            }

            if ( is_string( $maybe_bool ) ) {
                $maybe_bool = strtolower( $maybe_bool );
            }

            $true = array(
                true,
                '1',
                'true',
            );

            if ( in_array( $maybe_bool, $true, true ) ) {
                return true;
            }

            return false;           
        }
    }
```

Sorry for more or less reintroducing my suggestion, but let me quickly do some advertisement for it :)

For validation, you expect `true` or `false` to know if the value is valid. To sanitize a boolean, `true` or `false` are the values, if sanitized. But what to return if you do not enter a valid boolean? It can't be `false` and I do not think it should be nothing. So, I think `WP_Error` is a good solution.

What do you think?

BE-Webdesign:
I would make the PR without the `WP_Error` for now, as we will have to wait a while until the check for sanitizing the request args is in core.  Then we can open up an issue to keep track of wanting to put that in.  Alternatively you could leave it in but commented out with a note saying enable once core can handle this.

Otherwise, I think that function is good and would need to be called within `rest_sanitize_request_arg()` in the PR.

websupporter:
[...]

PR #2704, opened 2016-09-15
#2616 boolean validation

### LINKED ISSUE #2616: Boolean request args fail validation

BE-Webdesign:
Fixes #2616, and #2633, In addition #2630 can be closed. Adds boolean validation/sanitization, to handle integers, strings, and booleans.  `'false'` and `'0'` and integer `0` will evaluate to `false`.  The opposite of those values is `true`.

### DIFF
--- a/plugin.php
+++ b/plugin.php
@@ -299,8 +299,8 @@ if ( ! function_exists( 'rest_validate_request_arg' ) ) {
 			return new WP_Error( 'rest_invalid_param', sprintf( __( '%s is not of type %s' ), $param, 'integer' ) );
 		}
 
-		if ( 'boolean' === $args['type'] && ! is_bool( $value ) ) {
-			return new WP_Error( 'rest_invalid_param', sprintf( __( '%s is not of type %s' ), $param, 'boolean' ) );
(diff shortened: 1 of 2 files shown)


kadamwhite (approved):
I think this looks good, all cases I can think of for the boolean parameter validation look to be covered and the `rest_is_bool` method seems to be doing its job. Provisionally approving.

joehoyle (approved):

BE-Webdesign:
I merged it because it was approved by both of you and somehow you couldn't merge it, the button worked for me so I just clicked it, hopefully that is not bad form.

kadamwhite:
@BE-Webdesign No that's great; thanks for taking the initiative. I hope that "approve PR" doesn't mean "only the submitter can merge"; that would be counterintuitive.

Issue #2705, opened 2016-09-15
/wp/v2/posts?sticky=true should show only sticky posts


joehoyle:
Currently you can't query for just sticky (or just not sticky) posts, this shouldn't be too difficult to add.

BE-Webdesign:
Fixed in #2708

PR #2708, opened 2016-09-18
Add sticky parameter to the posts endpoint.

joehoyle:
This allows getting posts that are either "only sticky" for "not sticky"
via `sticky=true` or `sticky=false`. This can also be used in conjunction
with `include` and `exclude`.


joehoyle:
Fixes #2705

tests/test-rest-posts-controller.php
```diff
@@ -1441,6 +1520,19 @@ public function test_update_post_empty_content() {
...
+
+		$request = new WP_REST_Request( 'PUT', sprintf( '/wp/v2/posts/%d', $this->post_id ) );
+		$request->set_body_params( array(
+			'content' => 'Hello\\s',
+		) );
+
+		$response = $this->server->dispatch( $request );
+		$new_data = $response->get_data();
+		$this->assertEquals( 'Hello\\s', $new_data['content']['raw'] );
+	}
```
rmccue: This test appears to be unrelated to the rest of the PR, can you split it into a new one? :)

rmccue (changes requested):

tests/test-rest-posts-controller.php
joehoyle: Ohhhhhh that's why this is failing, doh! Didn't mean to commit this.

lib/endpoints/class-wp-rest-posts-controller.php
```diff
@@ -1776,6 +1798,14 @@ public function get_collection_params() {
 				'default'           => array(),
 			);
 		}
+
+		if ( 'post' === $this->post_type ) {
+			$params['sticky'] = array(
+				'description'       => __( 'Limit result set to items that are sticky.' ),
+				'type'              => 'boolean',
```
rmccue, line 1805: Indentation is off.

rmccue (changes requested):

rmccue:
Over to KAdam for review.

rmccue (approved):

lib/endpoints/class-wp-rest-posts-controller.php
```diff
@@ -124,6 +124,28 @@ public function get_items( $request ) {
 			unset( $args['filter'] );
 		}
 
+		if ( isset( $request['sticky'] ) ) {
```
kadamwhite, line 127: As noted in slack, `sticky=false` does not work (it's interpreted as `true` due to string coercion -- suggest using `rest_is_boolean` introduced in #2630

kadamwhite (changes requested):

tests/test-rest-posts-controller.php
```diff
@@ -357,6 +358,88 @@ public function test_get_items_tags_and_categories_query() {
...
+		$request = new WP_REST_Request( 'GET', '/wp/v2/posts' );
+		$request->set_param( 'sticky', false );
+
+		$response = $this->server->dispatch( $request );
+		$this->assertCount( 1, $response->get_data() );
+
+		$posts = $response->get_data();
+		$post = $posts[0];
+		$this->assertEquals( $id1, $post['id'] );
+	}
```
kadamwhite, line 422: Not sure if it's worth it but I think this test (variant of the one above) will demonstrate whether `sticky=false` in query string will work:

``` php
    public function test_get_items_not_sticky_query_string_boolean() {
        $id1 = $this->post_id;
        $id2 = $this->factory->post->create( array( 'post_status' => 'publish' ) );

        update_option( 'sticky_posts', array( $id2 ) );

        $request->set_param( 'sticky', 'false' );

        $response = $this->server->dispatch( $request );
        $this->assertCount( 1, $response->get_data() );

        $posts = $response->get_data();
        $post = $posts[0];
        $this->assertEquals( $id1, $post['id'] );
    }
```

lib/endpoints/class-wp-rest-posts-controller.php
```diff
@@ -1776,6 +1798,14 @@ public function get_collection_params() {
 				'default'           => array(),
 			);
 		}
+
+		if ( 'post' === $this->post_type ) {
+			$params['sticky'] = array(
+				'description' => __( 'Limit result set to items that are sticky.' ),
+				'type'        => 'boolean',
```
BE-Webdesign: Add in the `rest_validate_request_arg` and `rest_sanitize_request_arg` for validate and sanitize callbacks and this patch will be da hotness!

BE-Webdesign (changes requested):

tests/test-rest-posts-controller.php
kadamwhite: After discussion w/ @BE-Webdesign & @rmccue testing this case independently is probably overkill once the right sanitize callback is put in place

lib/endpoints/class-wp-rest-posts-controller.php
joehoyle: This will automajically be done based off the schema already.

joehoyle: Actually this is not the schema, sorry my bad!

joehoyle:
@rmccue @BE-Webdesign @kadamwhite ok I pushed an update to fix the `false` in the URL thang.

kadamwhite (approved):

kadamwhite:
+1 from me, `sticky=false` now works as expected

BE-Webdesign (approved):

PR #2718, opened 2016-09-22
Add Codecov configuration

### LINKED ISSUE #2717: Disable Codecov comments
This was unintentionally re-enabled when Codecov removed the configuration on their site, and it now requires having a file in the repo for the configuration instead.

We need to disable these again.

rmccue:
Fixes #2717.

### DIFF
--- a/.codecov.yml
+++ b/.codecov.yml
@@ -1,0 +1,1 @@
+comment: false


kadamwhite (approved):
💬 🚫

Issue #2701, opened 2016-09-14
Add support for password-protected posts


rmccue:
Previously: #1055.

We need to re-add support for password-protected posts in a nice, sane way. I've created [a core ticket](https://core.trac.wordpress.org/ticket/38056) to do step 1 of this: allow us to control authentication.
## Step 2: Work out how to handle passing the passwords.

We still need to work out how to actually pass the passwords in. There are basically three options:
### Cookies

Something like `Cookie: wp_rest_password=xyz` or `wp_rest_password_42=xyz` maybe. This overloads the meaning of cookies, and may not actually be possible to do from in-browser JavaScript.
### Authorization Header

`Authorization: Basic ...` or potentially a custom scheme like `Authorization: WP-Post ...`. The spec says you can't have multiple authorization headers, so it wouldn't be possible to be authenticated _and_ send a per-post password.
### Query Param

Something like `password=...` in the query. Not necessarily a terrible idea.
### Things we can't do
- Custom header: these aren't guaranteed to affect caching, so if a request includes a password and gets cached, the password is rendered useless.
- Existing cookie system: the cookie name includes a hash (`COOKIEHASH`), so is unguessable. This isn't usable for API clients.

rmccue:
I dug into this further; query parameters are the only viable option. We [discussed briefly on Slack](https://wordpress.slack.com/archives/core-restapi/p1474296851002021).

All that remains is to implement this. :)

kadamwhite:
Resolved by #2720

jetaro1:
thank you :)

PR #2720, opened 2016-09-22
Password Protected Posts Support

joehoyle:
In an attempt to better support password protected posts (see #2701) I've done a few things
now it's possible to better filter the password stuf in WordPress Core:
1. Password protected posts are now readable by every one.
2. All posts how have `content.protected` and `excerpt.protected` fields to indicate if those fields are protected by a password.
3. The `content.rendered` and `excerpt.rendered` are now empty strings if the post requires a password to be viewed.
4. Password protected posts in the rest api means "the post's content and excerpt are protected with a password".
5. If the user is viewing a post or listing posts in the `edit` context, they can see the protected fields.
6. The client can pass `?password=thepassword` to a single post endpoint to read it's protected fields.
7. An incorrect password with the above will result in a Forbidden response.
8. A client can now update and create password protected posts.
9. Post Types other than `post` no longer have support (or the property) for `password`.
10. Password protected post titles no longer have the "Protected: " prefix in REST API responses.

### DIFF
--- a/lib/endpoints/class-wp-rest-posts-controller.php
+++ b/lib/endpoints/class-wp-rest-posts-controller.php
@@ -37,7 +37,10 @@ class WP_REST_Posts_Controller extends WP_REST_Controller {
 				'callback'        => array( $this, 'get_item' ),
 				'permission_callback' => array( $this, 'get_item_permissions_check' ),
 				'args'            => array(
-					'context'          => $this->get_context_param( array( 'default' => 'view' ) ),
+					'context'  => $this->get_context_param( array( 'default' => 'view' ) ),
+					'password' => array(
+						'description' => __( 'The password for the post if it is password protected.' ),
+					),
 				),
 			),
 			array(
@@ -73,6 +76,11 @@ class WP_REST_Posts_Controller extends WP_REST_Controller {
 
 		if ( 'edit' === $request['context'] && ! current_user_can( $post_type->cap->edit_posts ) ) {
 			return new WP_Error( 'rest_forbidden_context', __( 'Sorry, you are not allowed to edit these posts in this post type' ), array( 'status' => rest_authorization_required_code() ) );
+		}
+
+		// Allow access to all password protected posts if the context is edit.
+		if ( 'edit' === $request['context'] ) {
+			add_filter( 'post_password_required', '__return_false' );
 		}
 
 		return true;
@@ -222,6 +230,20 @@ class WP_REST_Posts_Controller extends WP_REST_Controller {
 
 		if ( 'edit' === $request['context'] && $post && ! $this->check_update_permission( $post ) ) {
 			return new WP_Error( 'rest_forbidden_context', __( 'Sorry, you are not allowed to edit this post' ), array( 'status' => rest_authorization_required_code() ) );
+		}
+
+		if ( $post && ! empty( $request['password'] ) ) {
+			if ( ! hash_equals( $post->post_password, $request['password'] ) ) {
+				return new WP_Error( 'rest_post_incorrect_password', __( 'Incorrect post password.' ), array( 'status' => 403 ) );
+			} else {
+				// Allow access to the post going forward.
+				add_filter( 'post_password_required', '__return_false' );
+			}
+		}
+
+		// Allow access to all password protected posts if the context is edit.
+		if ( 'edit' === $request['context'] ) {
+			add_filter( 'post_password_required', '__return_false' );
 		}
 
 		if ( $post ) {
@@ -990,10 +1012,6 @@ class WP_REST_Posts_Controller extends WP_REST_Controller {
 	 * @return boolean Can we read it?
 	 */
 	public function check_read_permission( $post ) {
-		if ( ! empty( $post->post_password ) && ! $this->check_update_permission( $post ) ) {
-			return false;
-		}
-
 		$post_type = get_post_type_object( $post->post_type );
 		if ( ! $this->check_is_post_type_allowed( $post_type ) ) {
 			return false;
@@ -1137,10 +1155,12 @@ class WP_REST_Posts_Controller extends WP_REST_Controller {
 		}
 
 		if ( ! empty( $schema['properties']['title'] ) ) {
+			add_filter( 'protected_title_format', array( $this, 'protected_title_format' ) );
 			$data['title'] = array(
 				'raw'      => $post->post_title,
 				'rendered' => get_the_title( $post->ID ),
 			);
+			remove_filter( 'protected_title_format', array( $this, 'protected_title_format' ) );
 		}
 
 		if ( ! empty( $schema['properties']['content'] ) ) {
@@ -1150,15 +1170,11 @@ class WP_REST_Posts_Controller extends WP_REST_Controller {
 			}
 
 			$data['content'] = array(
-				'raw'      => $post->post_content,
+				'raw'       => $post->post_content,
 				/** This filter is documented in wp-includes/post-template.php */
-				'rendered' => apply_filters( 'the_content', $post->post_content ),
+				'rendered'  => post_password_required( $post ) ? '' : apply_filters( 'the_content', $post->post_content ),
+				'protected' => (bool) $post->post_password,
 			);
-
-			// Don't leave our cookie lying around: https://github.com/WP-API/WP-API/issues/1055.
-			if ( ! empty( $post->post_password ) ) {
-				$_COOKIE[ 'wp-postpass_' . COOKIEHASH ] = '';
-			}
 		}
 
 		if ( ! empty( $schema['properties']['excerpt'] ) ) {
@@ -1414,11 +1430,6 @@ class WP_REST_Posts_Controller extends WP_REST_Controller {
 					'context'     => array( 'view', 'edit' ),
 					'readonly'    => true,
 				),
-				'password'        => array(
-					'description' => __( 'A password to protect access to the post.' ),
-					'type'        => 'string',
-					'context'     => array( 'edit' ),
-				),
 				'slug'            => array(
 					'description' => __( 'An alphanumeric identifier for the object unique to its type.' ),
 					'type'        => 'string',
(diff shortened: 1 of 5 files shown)


kadamwhite:
Will test and review 9/23 AM by 12EDT

rmccue:
All 10 points here seem 👍 to me.

lib/endpoints/class-wp-rest-posts-controller.php
```diff
@@ -75,6 +78,11 @@ public function get_items_permissions_check( $request ) {
 			return new WP_Error( 'rest_forbidden_context', __( 'Sorry, you are not allowed to edit these posts in this post type' ), array( 'status' => rest_authorization_required_code() ) );
 		}
 
+		// Allow access to all password protected posts if the context is edit.
+		if ( 'edit' === $request['context'] ) {
+			add_filter( 'post_password_required', '__return_false' );
+		}
```
rmccue: This could be combined into the above block by splitting it.

lib/endpoints/class-wp-rest-posts-controller.php
```diff
@@ -224,6 +232,20 @@ public function get_item_permissions_check( $request ) {
 			return new WP_Error( 'rest_forbidden_context', __( 'Sorry, you are not allowed to edit this post' ), array( 'status' => rest_authorization_required_code() ) );
 		}
 
+		if ( $post && ! empty( $request['password'] ) ) {
+			if ( ! hash_equals( $post->post_password, $request['password'] ) ) {
+				return new WP_Error( 'rest_post_incorrect_password', __( 'Incorrect post password.' ), array( 'status' => 403 ) );
+			} else {
+				// Allow access to the post going forward.
+				add_filter( 'post_password_required', '__return_false' );
+			}
```
rmccue, line 266: Unnecessary `else`

rmccue (changes requested):
The filters also need to be reset after use, otherwise the API isn't re-entrant.

lib/endpoints/class-wp-rest-posts-controller.php
```diff
@@ -265,10 +287,6 @@ public function create_item_permissions_check( $request ) {
 
 		$post_type = get_post_type_object( $this->post_type );
 
-		if ( ! empty( $request['password'] ) && ! current_user_can( $post_type->cap->publish_posts ) ) {
-			return new WP_Error( 'rest_cannot_publish', __( 'Sorry, you are not allowed to create password protected posts in this post type' ), array( 'status' => rest_authorization_required_code() ) );
```
kadamwhite, line 291: Not 100% sure why this special case is needed, do password-protected posts have different capability requirements than regular posts? I'd not have expected to see a custom error message for the password-protected case here.

lib/endpoints/class-wp-rest-posts-controller.php
```diff
@@ -1137,34 +1111,30 @@ public function prepare_item_for_response( $post, $request ) {
...
-				$this->prepare_password_response( $post->post_password );
-			}
-
 			$data['content'] = array(
-				'raw'      => $post->post_content,
+				'raw'       => $post->post_content,
 				/** This filter is documented in wp-includes/post-template.php */
-				'rendered' => apply_filters( 'the_content', $post->post_content ),
+				'rendered'  => post_password_required( $post ) ? '' : apply_filters( 'the_content', $post->post_content ),
+				'protected' => (bool) $post->post_password,
```
kadamwhite, line 1191: It's a little counterintuitive to me to have the `protected: true` enumerated in every `content` and `excerpt` object, _vs_ just having it declared on the root level. This feels like it's adding a lot of duplication in to every response object, especially given that `protected: false` is by far the most common case. Why not use a top-level `protected` key?

lib/endpoints/class-wp-rest-posts-controller.php
```diff
@@ -1137,34 +1111,30 @@ public function prepare_item_for_response( $post, $request ) {
...
 
 		if ( ! empty( $schema['properties']['excerpt'] ) ) {
+			/** This filter is documented in wp-includes/post-template.php */
+			$excerpt = apply_filters( 'the_excerpt', apply_filters( 'get_the_excerpt', $post->post_excerpt, $post ) );
 			$data['excerpt'] = array(
-				'raw'      => $post->post_excerpt,
-				'rendered' => $this->prepare_excerpt_response( $post->post_excerpt, $post ),
+				'raw'       => $post->post_excerpt,
+				'rendered'  => post_password_required( $post ) ? '' : $excerpt,
+				'protected' => (bool) $post->post_password,
```
kadamwhite, line 1201: I may have missed this, but I also want to understand the reasoning for `.protected` instead of `.password`, which would seem to better represent what's being asked for as well as better mirror the existing core post object

lib/endpoints/class-wp-rest-posts-controller.php
```diff
@@ -1244,6 +1214,20 @@ public function prepare_item_for_response( $post, $request ) {
 	}
 
 	/**
+	 * Overwrite the default protected title format.
+	 *
+	 * By default WordPress will show password protected posts with a title of
+	 * "Protected: %s", as the REST API communicates the protected status of a post
+	 * in a machine readable format, we remove the "Protected: " prefix.
```
kadamwhite, line 1290: Minor quibble: including "Protected: " has nothing to do with machine readability and everything to do with that string being purely presentational. I'd alter the explanation of this decision here.

lib/endpoints/class-wp-rest-posts-controller.php
```diff
@@ -1619,6 +1610,12 @@ public function get_item_schema() {
 				'type'        => 'boolean',
 				'context'     => array( 'view', 'edit' ),
 			);
+
+			$schema['properties']['password'] = array(
+				'description' => __( 'A password to protect access to the content and excerpt.' ),
+				'type'        => 'string',
+				'context'     => array( 'edit' ),
+			);
```
kadamwhite, line 1687: I think I answered my question here... we can't have `.password` at the top level because that's where the string that sets that password must live. But I'm still 👎  on embedding the `.protected` key within the content and excerpt blocks; I'd prefer one of,
1. Have a `.password` property on the top level that is set as a string, and can be read as a string when password-authenticated, but otherwise displays as the boolean `true`; or
2. Have a `.protected` boolean property on the top level of the post object that represents whether the string `.password` field is present (whether a password is required)

kadamwhite, line 1687: I think I answered my question here... we can't have `.password` at the top level because that's where the string that sets that password must live. But I'm still 👎  on embedding the `.protected` key within the content and excerpt blocks; I'd prefer one of,
1. Have a `.password` property on the top level that is set as a string, and can be read as a string when password-authenticated, but otherwise displays as the boolean `true`; or
2. Have a `.protected` boolean property on the top level of the post object that represents whether the string `.password` field is present (whether a password is required)

kadamwhite (changes requested):

kadamwhite:
I confirmed the PR works as advertised, but had questions around the specifics of the implementation

rmccue:
Given that passwords affect more than one field, it makes sense to move `protected` out of the content/excerpt itself, I think. One reason we might not want to is if content and excerpt are independently protected, but I don't think that's possible?

lib/endpoints/class-wp-rest-posts-controller.php
joehoyle: Hmm, this is _no longer_ needed. It was here before to avoid support for password protected posts, which I now removed.

lib/endpoints/class-wp-rest-posts-controller.php
joehoyle: The reason we put it under only these fields is to make it clear that it's only the `content` and the `excerpt` that is protected. The post it's self is not protected, hence it not being a property of the post object it's self. "protected posts" is a bit of a misnomer, "posts with protected content and excerpt". As for duplication, we'll need a `protected: false` on every post either way. However, it's true that (I think anyway) it's not possible to have an excerpt which is public and content which is private, however I'm not too concerned for the duplication. it's not really any more bytes over the wire with gzipping.

lib/endpoints/class-wp-rest-posts-controller.php
joehoyle: Sure can do - I was getting at there's no need to mark a post as "Protected: $x" because we communicate that it's protected in other ways.

lib/endpoints/class-wp-rest-posts-controller.php
joehoyle: I'm 👎  on `password` being a string or a boolean depending on context. As I mentioned above, I don't think `protected` at the top level makes as much sense because it's not saying the whole post is protected, however I don't mind too much on that.

lib/endpoints/class-wp-rest-posts-controller.php
kadamwhite: :100: on the message, was just nitpicking the specifics :D

lib/endpoints/class-wp-rest-posts-controller.php
kadamwhite: I'm curious whether anyone else agrees with me, then :) If not, I won't push.

Would we ever have content protected, but excerpt not?

kadamwhite:
Missed @rmccue's comment, I had the same question about content/excerpt protection independence. Based on Ryan's comment I'd like to consolidate the two unless some specific use-case or possibility for separate protection states is demonstrated.

rmccue:
I've changed my mind on this; I think `protected` inside the content/excerpt makes more sense, as it makes it much more obvious as to why the data is empty. I could see the potential for content and excerpt being protected separately as well, either in a plugin or in the future of core, and I think it's a little more future-proof.

joehoyle:
@rmccue your review still shows requesting changes. With your PR that is now merged into this, is there any more changes required from you?

rmccue (approved):

kadamwhite (approved):
I've been talked down on top-level `.protected` key, which is my one disagreement here; +1 from me on this

PR #2721, opened 2016-09-23
Ability to orderby slug, email and url on users endpoints.

### LINKED ISSUE #2658: Support more orderby params in the users controller
In https://github.com/WP-API/WP-API/commit/69148d506f72e4b2d7fb71d0ed5e18cedecea1be we seemingly arbitrarily added some more orderby params, however there's other we could also add, such as `slug` and `email` that are support by `get_users()`. We should need to be careful to only allowing ordering by private fields for auth'ed users.

joehoyle:
Fixes #2658

### DIFF
--- a/lib/endpoints/class-wp-rest-users-controller.php
+++ b/lib/endpoints/class-wp-rest-users-controller.php
@@ -86,6 +86,10 @@ class WP_REST_Users_Controller extends WP_REST_Controller {
 			return new WP_Error( 'rest_forbidden_context', __( 'Sorry, you cannot view this resource with edit context.' ), array( 'status' => rest_authorization_required_code() ) );
 		}
 
+		if ( in_array( $request['orderby'], array( 'email' ), true ) && ! current_user_can( 'list_users' ) ) {
+			return new WP_Error( 'rest_forbidden_orderby', __( 'Sorry, you cannot order by this parameter.' ), array( 'status' => rest_authorization_required_code() ) );
+		}
+
 		return true;
 	}
 
@@ -112,6 +116,9 @@ class WP_REST_Users_Controller extends WP_REST_Controller {
 			'include'         => 'include',
 			'name'            => 'display_name',
 			'registered_date' => 'registered',
+			'slug'            => 'user_nicename',
+			'email'           => 'user_email',
+			'url'             => 'user_url',
 		);
 		$prepared_args['orderby'] = $orderby_possibles[ $request['orderby'] ];
 		$prepared_args['search'] = $request['search'];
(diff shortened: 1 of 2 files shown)


joehoyle:
@rmccue #reviewmerge

lib/endpoints/class-wp-rest-users-controller.php
```diff
@@ -86,6 +86,10 @@ public function get_items_permissions_check( $request ) {
 			return new WP_Error( 'rest_forbidden_context', __( 'Sorry, you cannot view this resource with edit context.' ), array( 'status' => rest_authorization_required_code() ) );
 		}
 
+		if ( in_array( $request['orderby'], array( 'email' ), true ) && ! current_user_can( 'list_users' ) ) {
```
BE-Webdesign: `in_array( $request['orderby'], array( 'email' ), true )`  What does that help tell us exactly? Couldn't any user who can't list users not order them, not necessarily just by email?  I am bit confused by the conditional.  Is it here because the other fields are publicly exposed, so if someone wanted to they could order by the other fields? If that's the intention, we would need to add in `registered_date` as well because that is not publicly queryable as well.  If the conditional is for some other reason please let me know, I am probably overlooking something obvious.

joehoyle: Ah, yes we should add `registered_date` in that case, we shouldn't be exposing this right now.

BE-Webdesign: Okay, cool.  Everything else looks great to me.

tests/test-rest-users-controller.php
```diff
@@ -259,6 +259,103 @@ public function test_get_items_orderby() {
...
+		$request->set_param( 'orderby', 'email' );
+		$request->set_param( 'order', 'asc' );
+		$request->set_param( 'per_page', 1 );
+		$request->set_param( 'include', array( $low_id, $high_id ) );
+		$response = $this->server->dispatch( $request );
+		$data = $response->get_data();
+		$this->assertEquals( $low_id, $data[0]['id'] );
+	}
+
+	public function test_get_items_orderby_email_unauthenticaed() {
```
kadamwhite: Typo

kadamwhite (changes requested):

rachelbaker (approved):

joehoyle:
I removed the ability to order by `registered_date` for unauthenticated requests, as that's not a public field, fixed up the typo per @kadamwhite.

kadamwhite (approved):

Issue #2694, opened 2016-09-13
Required custom fields not honored when registered in rest_api_init


psalz:
When adding a custom field inside a `rest_api_init` handler, setting `required => true` doesn't have any effect.

Example:

``` php
add_action('rest_api_init', function() {
  register_rest_field('post', 'foo', [
    'get_callback' => 'get_foo',
    'update_callback' => 'set_foo',
    'schema' => [
      'required' => true
    ]
  ]);
};
```

Requesting `POST /posts` does not return the expected `rest_missing_callback_param` error.

This is due to the call to `WP_REST_Controller::get_endpoint_args_for_item_schema` being made via `create_initial_rest_routes` (which in turn runs on `rest_api_init:0`, well before a default priority hook would run). This means the schema entry for `foo` doesn't even exist at that point, and so it isn't included in the arguments for the `POST` handler.

This issue can be circumvented by registering custom fields on `init` or `rest_api_init:-1`, but I find that to be rather unintuitive.

websupporter:
Is handled in https://github.com/WP-API/WP-API/issues/2810

PR #2727, opened 2016-09-23
Store generated schema in class property

### LINKED ISSUE #2558: Increasing per_page scales poorly due to WP_REST_Posts_Controller::get_item_schema calls
Hi,

Whilst profiling our application, I found that as you increase per_page on a posts request there is a subsequent rise in get_item_schema calls. These calls are quite intensive (particularly if there is extra schema being added in a custom post type).

Here are some stats from kcachegrind showing that the inclusive cost of get_item_schema() calls trends upwards without caching, but with caching the inclusive cost is lower, and trends sharply downwards as per_page increases. The effect of caching is also significant increase in the speed of the overall HTTP request.

```
without schema caching:
per_page incl  self  called (overall time to run 5x http requests as reported by curl)
10       24.74 6.67  278    (3.577)
20       31.15 8.65  528    (5.701)
30       33.23 9.51  778    (7.199)
40       34.70 10.61 1028   (8.671)
100      37.85 10.80 2528   (18.925)

with schema caching:
per_page incl self called (overall time to run 5x http requests as reported by curl)
10       2.57 0.22 278    (2.899)
20       1.39 0.18 528    (3.946)
30       0.87 0.13 778    (4.887)
40       0.17 0.10 1028   (6.239)
100      0.28 0.04 2538   (12.072)
```

Here is the code that did the caching from a subclass:

```
  private $schema;
  public function get_item_schema() {
    if ($this->schema) {
      return $this->schema;
    }

    $this->schema = parent::get_item_schema();
    return $this->schema;
  }
```

rmccue:
Fixes #2558.

### DIFF
--- a/lib/endpoints/class-wp-rest-attachments-controller.php
+++ b/lib/endpoints/class-wp-rest-attachments-controller.php
@@ -297,6 +297,9 @@ class WP_REST_Attachments_Controller extends WP_REST_Posts_Controller {
 	 * @return array
 	 */
 	public function get_item_schema() {
+		if ( isset( $this->schema ) ) {
+			return $this->schema;
+		}
 
 		$schema = parent::get_item_schema();
 
@@ -355,7 +358,9 @@ class WP_REST_Attachments_Controller extends WP_REST_Posts_Controller {
 			'context'         => array( 'view', 'edit', 'embed' ),
 			'readonly'        => true,
 		);
-		return $schema;
+
+		$this->schema = $schema;
+		return $this->schema;
 	}
 
 	/**
--- a/lib/endpoints/class-wp-rest-comments-controller.php
+++ b/lib/endpoints/class-wp-rest-comments-controller.php
@@ -4,6 +4,12 @@
  * Access comments
  */
 class WP_REST_Comments_Controller extends WP_REST_Controller {
+	/**
+	 * Response schema.
+	 *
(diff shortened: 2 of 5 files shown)


danielbachhuber:
Do you think it will ever post a problem to have the schema cached? e.g. what happens if a field is registered after the schema is cached?

BE-Webdesign:
> Do you think it will ever post a problem to have the schema cached? e.g. what happens if a field is registered after the schema is cached?

The tests seem to reflect that concern, as the additional fields aren't registering anymore.  The registering could be changed to just work with the object property though.  I mainly think using schema as an object property would be a problem with internal requests, via batch processes.

rmccue:
> Do you think it will ever post a problem to have the schema cached? e.g. what happens if a field is registered after the schema is cached?

This will already be a problem, since the schema won't be exposed to the client, and will be inconsistent internally. I think having internal consistency makes it more obvious if you're hooking in at the wrong point.

danielbachhuber:
> I think having internal consistency makes it more obvious if you're hooking in at the wrong point.

Maybe all of these schemas should be generated and cached at the same time (on some hook) to ensure there's internal consistency from hook to hook.

rmccue:
I'm pretty sure it is consistently generated right now on `rest_api_init`, since the schema is passed into the route registration. See also #2694.

Issue #2582, opened 2016-07-04
Using term filters can cause the API to return objects instead of arrays


27pchrisl:
Hi,

WordPress allows you to filter processes like get_terms. Through this you can unset terms from the array that's then returned for further processing. A side effect of doing this that affects the JSON encoder but not normally further PHP processing is that the JSON encoder must now output arrays filtered in this way as JSON objects instead of arrays, because the internal structure of the array is now key/value due to deleted numeric keys (wrong terminology, but it sort of works).

An example of the problem is here: http://stackoverflow.com/questions/5943149/rebase-array-keys-after-unsetting-elements

Should the API:

a) defensively call array_values on things it expects to be trying to output as an array? (So that the API contract can't be broken), or:
b) should this be documented behaviour and it's the responsibility of the filter developer to call array_values?

I might suggest (a), if only because there may be plugins out in the world that do not call array_values because it's never affected PHP, but being used on sites that then would have potentially random WP-API behaviour switching between Object and Array for filtered and non-filtered term arrays.

This probably isn't restricted to term filters, but it's where I noted the behaviour.

BE-Webdesign:
Hi @27pchrisl,

Can you provide some sample requests to the API that would cause the array to be broken? We should probably go through the code base and look at where this can occur and I would lean towards doing option a that you have suggested.

27pchrisl:
Sure, you can reproduce the issue by creating a post and adding it to some categories, then adding the filter below:

Normally when you hit /wp-json/wp/v2/posts/1 you'll get (as part of the response) "categories":[2,3,4,1]

If I include this filter in the process to delete one of the terms:

```
function restapi_term_filter($terms, $taxonomies, $args) {
  unset($terms[1]);
  return $terms;
}
add_filter('get_the_terms', 'restapi_term_filter', 10, 3);
```

this is now the response: "categories":{"0":2,"2":4,"3":1} because of the internal way PHP is keying the array.

Changing this to:

```
function restapi_term_filter($terms, $taxonomies, $args) {
  unset($terms[1]);
  return array_values($terms);
}
add_filter('get_the_terms', 'restapi_term_filter', 10, 3);
```

gives the expected response: "categories":[2,4,1]

Since there was never any need in the PHP world to return from that filter with with array_values(), there may be themes and plugins that are manipulating using 'get_the_terms' (and other related filters) that will have inconsistent results with the REST API.

Hope that helps!

BE-Webdesign:
[...]

PR #2728, opened 2016-09-24
#2582 call array values on roles

BE-Webdesign:
#2582 was fixed in #2724, but it is also possible to remove user roles from an array in `set_user_role`, `add_user_role`, or `remove_user_role`. It would be pretty hard to do so, and most likely no one does it, but here is a PR just in case. Feel free to close the PR if it is unwanted.


kadamwhite (approved):
IMO it's not an impactful change and it adds consistency and safety in the way the API functions... I'd be +1. Will wait for another team member to weigh in before pulling the trigger though.

I assume that given the situation is hard to cause to begin with, the overhead of adding a test for this is prohibitive.

PR #2732, opened 2016-09-25
#2729 add tests update extra field callbacks

### LINKED ISSUE #2729: Add tests for returning an error from field update callbacks
We should always have tests covering functional changes.

From https://github.com/WP-API/WP-API/pull/2702#issuecomment-247153482

BE-Webdesign:
Fixes #2729.  Here are tests to show that the update callback can now handle returning errors for the response.  The tests just add a `returnError` flag to the update callback and if that value is supplied an error will be returned.


BE-Webdesign:
Build needs a restart SSL handshake failed.

kadamwhite (approved):
Looks good to me, I confirmed that undoing #2702 will cause all of these tests to fail. Thanks @BE-Webdesign !

PR #2735, opened 2016-09-26
Move post_password_required filtering to preparation

rmccue:
This allows resetting the filter after the content/excerpt have been generated, ensuring the endpoints are re-entrant.

Builds on #2720.


lib/endpoints/class-wp-rest-posts-controller.php
```diff
@@ -254,6 +257,36 @@ public function get_item_permissions_check( $request ) {
...
+			return true;
+		}
+
+		// No password, no auth.
+		if ( empty( $request['password'] ) ) {
+			return false;
+		}
+
+		// Double-check the request password.
+		return hash_equals( $post->post_password, $request['password'] );
```
rmccue, line 286: Note that this is checked twice: once in the permissions check, and again here. I don't see any huge issues with that, and it makes it a bit cleaner IMO.

joehoyle:
Looks good!

PR #2738, opened 2016-09-28
Fix regular expression used to determine "single" routes

fabrizim:
The regular expression was matching "single" routes for anything ending in either "]", "m", or "e", instead of routes that end in either "+)" or "me" as intended. This fixes the regular expression to use subpatterns properly instead of the character set.

### DIFF
--- a/wp-api.js
+++ b/wp-api.js
@@ -1080,7 +1080,7 @@
 				) {
 
 					// Single items end with a regex (or the special case 'me').
-					if ( /.*[+)|me]$/.test( index ) ) {
+					if ( /.*(\+\)|me)$/.test( index ) ) {
 						modelRoutes.push( { index: index, route: route } );
 					} else {
 


BE-Webdesign:
Hi @fabrizim,

Thank you for the PR!  This involves the client js repo, which is built into this repo.  If you could make this PR in the [client JS repo](https://github.com/WP-API/client-js) it would be greatly appreciated so that build process will carry in these changes.

fabrizim:
Ah - thanks! I didn't realize there was a separate repo for the client. Will do now.

fabrizim:
@BE-Webdesign - Just FYI - it looks like they fixed that bug back in May (https://github.com/WP-API/client-js/commit/e5bbca6c681341677843504ea7cd875dd7146f41). Do you know when changes are merged into this repo? In the meantime, it looks like I can install the client repo as a separate plugin and use that instead. Thanks.

rmccue:
ping @adamsilverstein @kadamwhite can we update the JS? :)

BE-Webdesign:
> Just FYI - it looks like they fixed that bug back in May

Thank you for the due diligence, we will get that fixed!  Did not realize that this was manually being done.

adamsilverstein:
@rmccue I verified that updated code is in the current release. https://github.com/WP-API/WP-API/blob/develop/wp-api.js#L1104

PR #2741, opened 2016-09-29
Require comment content on Create & type check on Update

### LINKED ISSUE #2284: comments can be send with empty values

### LINKED ISSUE #2758: Standardize the expected POST/PUT properties expected for schema nested objects
In the [Posts Controller we accept POST/PUT values for `title` and `content` to be sent as `title.raw` or `content.raw`](https://github.com/WP-API/WP-API/blob/develop/lib/endpoints/class-wp-rest-posts-controller.php#L782). In this controller, we do NOT have a `santization_callback` set on the top-level `title` or `content` properties in the schema.

The way the Comments Controller is setup prevents being able to pass `content.raw` to set the comment_content [because we DO have a sanitization_callback set on the top level `content` property](https://github.com/WP-API/WP-API/blob/develop/lib/endpoints/class-wp-rest-comments-controller.php#L867) in the schema.  This means attempting to create or update a comment with `content.raw` set results in the following error:

> stripslashes() expects parameter 1 to be string, array given   

Some questions that need answering:
1. Do we want to limit the acceptable parameters for the nested objects?  Examples: 
   - only accept `content` and ignore `content.raw` values
   - only accept `content.raw` and ignore `content` values
2. Do we want to accept both top-level and `.raw` property values for the nested objects?
3. How can we modify the schema (if needed) to apply callbacks or required parameters to nested objects?

rachelbaker:
In `wp_handle_comment_submission()` core requires new comments to have content, we should match that behavior.  Comments can be edited to have empty content (if the user has `edit_comment` permission), but we should still check that at least the comment content is a string.

Included in this PR are the following functionality changes:
- Allow the Comment content to be provided for create or update requests via the `$request['content']` OR `$request['content']['raw']` parameters, matching logic from the Posts Controller. (See #2758)
- Sanitize the comment_content with `wp_filter_kses()` which matches WP Core. **EXCEPTION**: Core sanitizes the comment_content for users with the `unfiltered_html` capability with `wp_filter_kses_post()`. See ticket for discussion #2787.
- Return a WP_Error if the Comment content is empty on create.
- Return a WP_Error if the Comment content is an empty string `''` on create.
- Return a WP_Error if the Comment content is empty on update. **EXCEPTION**: when an update request [_only_ includes that status parameter](https://github.com/WP-API/WP-API/blob/develop/lib/endpoints/class-wp-rest-comments-controller.php#L452), we do not require the Comment content parameter.

Fixes #2758 and #2284.
Addresses PART OF #2049


tests/test-rest-comments-controller.php
```diff
@@ -674,6 +674,25 @@ public function test_create_item_invalid_date() {
...
 	}
 
+	public function test_create_item_invalid_content() {
+		wp_set_current_user( 0 );
+
+		$params = array(
+			'post'         => $this->post_id,
+			'author_name'  => 'Reverend Lovejoy',
+			'author_email' => 'lovejoy@example.com',
+			'author_url'   => 'http://timothylovejoy.jr',
```
rmccue, line 780: 👏

rmccue (approved):

lib/endpoints/class-wp-rest-comments-controller.php
```diff
@@ -432,6 +436,10 @@ public function update_item( $request ) {
 			return new WP_Error( 'rest_comment_invalid_type', __( 'Sorry, you cannot change the comment type.' ), array( 'status' => 404 ) );
 		}
 
+		if ( isset( $request['content'] ) && ! is_string( $request['content'] ) ) {
+			return new WP_Error( 'rest_comment_content_invalid', __( 'Invalid comment content.' ), array( 'status' => 400 ) );
```
joehoyle: We are using `wp_filter_post_kses` for the `sanitize_callback` on this, so I'm not sure how this would fail?

lib/endpoints/class-wp-rest-comments-controller.php
```diff
@@ -318,6 +318,10 @@ public function create_item( $request ) {
 			return new WP_Error( 'rest_comment_exists', __( 'Cannot create existing comment.' ), array( 'status' => 400 ) );
 		}
 
+		if ( empty( $request['content'] ) || ! is_string( $request['content'] ) ) {
+			return new WP_Error( 'rest_comment_content_required', __( 'Missing comment content.' ), array( 'status' => 400 ) );
```
joehoyle: Can we not just make this a `required` param in the schema here: https://github.com/WP-API/WP-API/blob/f73f3e8f7df3bebfc6e6c2476a4d6b2f1f2ec136/lib/endpoints/class-wp-rest-comments-controller.php#L854?

rachelbaker: Well in the schema `content` is an object, like title and content is in the Posts Controller....so how would we adapt the schema to handle the create/update params that are not objects?

lib/endpoints/class-wp-rest-comments-controller.php
rachelbaker: I didn't realize that. Good catch.

rachelbaker: @joehoyle Where are you seeing the `wp_filter_post_kses` callback being set for update requests?

lib/endpoints/class-wp-rest-comments-controller.php
```diff
@@ -728,8 +734,20 @@ protected function prepare_status_response( $comment_approved ) {
 	protected function prepare_item_for_database( $request ) {
 		$prepared_comment = array();
 
-		if ( isset( $request['content'] ) ) {
-			$prepared_comment['comment_content'] = $request['content'];
+		if ( isset( $request['content'] ) && is_string( $request['content'] ) ) {
```
rachelbaker, line 776: See similar logic in the Posts Controller: https://github.com/WP-API/WP-API/blob/develop/lib/endpoints/class-wp-rest-posts-controller.php#L782

lib/endpoints/class-wp-rest-comments-controller.php
rachelbaker: NVM, I found it.

rachelbaker:
Ugghh.. ran into a problem #2758

lib/endpoints/class-wp-rest-comments-controller.php
```diff
@@ -728,9 +743,20 @@ protected function prepare_status_response( $comment_approved ) {
...
+		 */
+		if ( isset( $request['content']['raw'] ) ) {
+			$prepared_comment['comment_content'] = $request['content']['raw'];
+		} elseif ( isset( $request['content'] ) ) {
 			$prepared_comment['comment_content'] = $request['content'];
 		}
+		if ( current_user_can( 'unfiltered_html' ) && isset( $prepared_comment['comment_content'] ) ) {
+			$prepared_comment['comment_content'] = wp_filter_post_kses( $prepared_comment['comment_content'] );
+		} elseif ( isset( $prepared_comment['comment_content'] ) ) {
+			$prepared_comment['comment_content'] = wp_filter_kses( $prepared_comment['comment_content'] );
```
rmccue: Think this should be `wp_kses`, and the above should be `wp_kses_post`

rachelbaker: Doh!

rachelbaker: @rmccue Turns out the problem stems from authenticating as an `admin` level user and then updating a comment for a subscriber.  I don't have easy access to a PHP 5.2 or 5.3 machine to debug this fully.  Any thoughts?

See https://travis-ci.org/WP-API/WP-API/builds/165834410 which finally passes

rachelbaker:
@WP-API/amigos Sorry for the messy commit history here, but I have this ready for review.  I updated the [PR description above](https://github.com/WP-API/WP-API/pull/2741#issue-179942623) to include an accurate summary of the changes here.

kadamwhite (approved):
I wouldn't use my review as the authoritative go-ahead but the patch looks good to my eyes!

rachelbaker:
> I wouldn't use my review as the authoritative go-ahead but the patch looks good to my eyes!

@kadamwhite I just needed a final set of eyes since @rmccue already gave the authoritative go-ahead.  Merged, to keep us moving.

PR #2744, opened 2016-09-30
Don't allow reading / creating of posts with no parent

joehoyle:
This also stops access to comments on private posts when accessing them
directly.


rmccue (approved):

rachelbaker:
Props to @kadamwhite for responsibly disclosing this issue.

Issue #2737, opened 2016-09-27
Support `auto-draft` status in the API


kadamwhite:
Splitting this out from #2629 where it was proposed:

In WP-Admin, `auto-draft` is used to account for posts that are not yet “fully” drafted, and therefore disables _e.g._ revisions. In #2629 @adamsilverstein proposed that the API support the `auto-draft` status, to permit a client to set a post into that state to permit it to be updated without generating revisions (and to support all valid statuses).

ellatrix:
I didn't _propose_ that. I would like there to be a way to update a post without creating revisions. The means don't really matter to me. @adamsilverstein suggested the status "trick", but that's not an option currently.

kadamwhite:
@iseulde ah, right you are -- my apologies, was skimming the issue in a hurry and mis-attributed!

_Edit: updated issue description to capture Adam's origination of the idea_

rmccue:
See also #2253.

kadamwhite:
I am +1 on supporting putting a post into `auto-draft`, and have yet to hear any specific arguments against

celloexpressions:
`auto-draft` is becoming increasingly important for live-previewing things in the customizer. I imagine that there would be similar benefits to supporting it here, but can't comment on specifics beyond that.

westonruter:
I am also +1 on supporting `auto-draft`. As @celloexpressions noted, the customizer is now leveraging `auto-draft` posts more and more to create placeholders for posts created in customizer sessions. So considering a customizer application that used a transaction or “changeset” endpoint to manipulate the `customized` state, the client would need to create an `auto-draft` post/page first and then reference that `auto-draft` post ID in an update to the `customized` state.

kadamwhite:
As discussed in [dev chat on 2016-09-28](https://wordpress.slack.com/archives/core/p1475096746003195) supporting `auto-draft` is not a blocker for merge, but there's general support for supporting this in the API and I will attempt to address it as time permits (or anybody else can pick it up, not going to hog the issue!)

I am changing the issue title from "should the API do this" to "Implement support for this"

kadamwhite:
The behavior of auto-drafts is inconsistent and does not do the revisions opt-out thing for which this ticket was originally opened. After discussion with @joehoyle and @rmccue, this ticket (and the code proposed in PR #2745) are being moved to 2.1 so that more time can be spent establishing what the role of auto-drafts should be in an API-enabled WordPress context.

PR #2745, opened 2016-09-30
Support setting posts to "auto-draft" status

kadamwhite:
For #2737

The list of statuses supported by the API omits auto-draft, both in the list of statuses a post will accept and in the collection returned from `/wp/v2/statuses`. This PR begins to add support; the prior case should be working with the below, but I'll need to investigate more to understand where I need to make a change to get auto-draft to show up in the queryable statuses list. (Would only show when authenticated)

### DIFF
--- a/lib/endpoints/class-wp-rest-post-statuses-controller.php
+++ b/lib/endpoints/class-wp-rest-post-statuses-controller.php
@@ -63,6 +63,7 @@ class WP_REST_Post_Statuses_Controller extends WP_REST_Controller {
 	public function get_items( $request ) {
 		$data = array();
 		$statuses = get_post_stati( array( 'internal' => false ), 'object' );
+		$statuses['auto-draft'] = get_post_status_object( 'auto-draft' );
 		$statuses['trash'] = get_post_status_object( 'trash' );
 		foreach ( $statuses as $slug => $obj ) {
 			$ret = $this->check_read_permission( $obj );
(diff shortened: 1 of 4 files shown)


westonruter:
> I'll need to investigate more to understand where I need to make a change to get auto-draft to show up in the queryable statuses list. (Would only show when authenticated)

I think I would advise against this. An `auto-draft` post should not be exposed in any collection responses. It should be hidden and its existence can only be known if a client creates such a post with that status and then has the post ID that it can reference in requests. So it should only appear in singular `item` requests, not in collection `items` requests.

kadamwhite:
@westonruter 

> An auto-draft post should not be exposed in any collection responses.

Can you explain this point more? I know auto-drafts are hidden in the post list in WP-Admin, but I believe that if the API can be used to set a resource into a state, it should be able to query for that resource by filtering on that state; this is predictable and consistent with all other states. It is possible to query `/wp/v2/posts?status=auto-draft` today, there's just not any way to _create_ an auto-draft post using the API.

rmccue:
I agree with @kadamwhite; if you can set the post into a state, you should be able to access it back out. I'm not sure whether actually exposing the state of `auto-draft` is necessarily what we want though, so we should revisit whether that's the best exposure method post-4.7.

westonruter:
@kadamwhite As long as the `auto-draft` post status has to be explicitly provided, then I think that's fine.

kadamwhite:
@westonruter yep, that is correct, a post created through the API defaults to `draft` (the regular kind) unless you explicitly specify otherwise.

kadamwhite:
This is now ready for review, I've cleaned up the patch & fixed a bad test assertion.

Full changes:
- Whitelist `auto-draft` status in posts controller so that posts may be set to `auto-draft`
- Include `auto-draft` _[sic, WP returns name as lower-cased]_ in the list of valid statuses returned from the `/wp/v2/statuses` endpoint; it will only appear if the authenticated user has the capability to edit posts.
- Add or update unit tests for the above

kadamwhite:
After testing revealed inconsistencies in how revisions are actually being set when posts are in different statuses, this PR is being closed and the parent issue has been moved to 2.1 or beyond

PR #2748, opened 2016-10-01
#2145 tax query in filter param

### LINKED ISSUE #2145: Support taxonomy queries

BE-Webdesign:
Fixes #2145, this was put into a new PR to make it cleaner.  Original PR is in #2273.  This enables `tax_query` in the `filter` param and validates that the taxonomies that are queried are `show_in_rest => true`


lib/endpoints/class-wp-rest-posts-controller.php
```diff
@@ -1874,4 +1876,73 @@ public function validate_user_can_query_private_statuses( $value, $request, $par
...
+					// If an error is found keep returning it up the chain.
+					if ( is_wp_error( $valid ) ) {
+						return $valid;
+					}
+				}
+
+				if ( 'taxonomy' === $key ) {
+					// If $taxonomies is empty by default set it. On reoccurring calls the function overhead is avoided.
+					if ( empty( $taxonomies ) ) {
+						$taxonomies = wp_list_filter( get_object_taxonomies( $this->post_type, 'objects' ), array( 'show_in_rest' => true ) );
```
kadamwhite, line 1933: Could/should this check & assignment go at the top of this block, before the `foreach`? I doubt there's much perf impact here but since the list of taxonomies that apply to this request's post type won't be changing during this loop, the list of `$taxonomies` could be declared outside the foreach loop.

kadamwhite: Also "subsequent" is probably a better descriptor in that comment than re-occurring

kadamwhite:
This does what it says on the tin; tested it out fairly fully with wp-api/node-wpapi#243 and it's behaving as I would expect. `tax_query` definitely introduces some potential performance overhead, but I'm not sure whether it's our responsibility to limit that artificially in any way.

lib/endpoints/class-wp-rest-posts-controller.php
BE-Webdesign: If you are using tax_query in combination with the top level params we have, keeping this call here inside the loop you avoid ever checking against taxonomies if you don't need to.  For instance if you were going to do something like `?categories=1,4&filter[tax_query][relation]=AND` you would not be checking against taxonomies again that were originally done in the `categories=1,4` part of that.  If the code doesn't need to run, why run it?

Subsequent sounds better to me too, I can change that.  Or maybe "following calls" would be good to?

kadamwhite: either/or on the wording!

It took me a second to understand what you meant in terms of the `?categories` impact, but that makes sense to me. +1

kadamwhite (approved):

rachelbaker (approved):

lib/endpoints/class-wp-rest-posts-controller.php
```diff
@@ -1874,4 +1876,73 @@ public function validate_user_can_query_private_statuses( $value, $request, $par
...
+				return $is_valid;
+			}
+		}
+		return $is_valid;
+	}
+
+	/**
+	 * Validate tax query.
+	 *
+	 * @param  mixed           $value      Value of the parameter.
```
rachelbaker, line 1906: Wrong `param` name here.  Should be `$values`.

tests/test-rest-posts-controller.php
```diff
@@ -588,6 +588,67 @@ public function test_get_items_valid_date() {
 		$this->assertEquals( $post2, $data[0]['id'] );
 	}
 
+	public function test_get_items_valid_tax_query_in_filter() {
+		$term1 = $this->factory->term->create( array( 'taxonomy' => 'category', 'name' => 'basie' ) );
+		$post1 = $this->factory->post->create();
```
rachelbaker, line 593: Not a blocker, but no need to assign the `$post1` and `$post3` variables since you don't use them.

tests/test-rest-posts-controller.php
```diff
@@ -588,6 +588,67 @@ public function test_get_items_valid_date() {
...
+		) );
+		$response = $this->server->dispatch( $request );
+		$data = $response->get_data();
+		$this->assertCount( 1, $data );
+		$this->assertEquals( $post2, $data[0]['id'] );
+		$this->assertEquals( $term1, $data[0]['categories'][0] );
+	}
+
+	public function test_get_items_invalid_tax_query_in_filter() {
+		$post1 = $this->factory->post->create();
```
rachelbaker, line 617: Not a blocker, but no need to assign the `$post1` variable, it isn't used.

kadamwhite:
The flexibility of completely allowing `tax_query` was deemed too big a cat to let out of the bag in [slack discussion today](https://wordpress.slack.com/archives/core-restapi/p1475769430005510), so closing this. The ability to set an AND relation for taxonomy IDs included via `{taxonomyname}[]=id` query parameters is an open question to be addressed another day.

PR #2751, opened 2016-10-01
Remove Scrutinzer

joehoyle:
We don't use this anymore, and it's an additional build step that we
might as well save on.


tobeycodes:
Is there a reason you don't use it anymore? Never used it before but looks useful

rmccue:
We're already testing in Travis in the `phpvalidate` test.

rmccue (approved):

rmccue:
I disabled the Scrutinizer webhook, so shouldn't get the fail on future PRs.

PR #2753, opened 2016-10-03
Use get_comment_type() when comparing updating comment status

joehoyle:
Currently we don't allow updating a comment type, however in the update
code we are checking the raw value vs get_comment_type() which means a
comment of type "comment" can not be updated, as that's actually an
empty value in the database.

### DIFF
--- a/lib/endpoints/class-wp-rest-comments-controller.php
+++ b/lib/endpoints/class-wp-rest-comments-controller.php
@@ -765,7 +765,8 @@ class WP_REST_Comments_Controller extends WP_REST_Controller {
 		}
 
 		if ( isset( $request['type'] ) ) {
-			$prepared_comment['comment_type'] = $request['type'];
+			// Comment type "comment" needs to be created as an empty string.
+			$prepared_comment['comment_type'] = 'comment' === $request['type'] ? '' : $request['type'];
 		}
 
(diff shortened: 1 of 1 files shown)


rachelbaker:
Related core ticket that would be the IDEAL solution: https://core.trac.wordpress.org/ticket/38200

rachelbaker:
@joehoyle the default value for the `type` parameter is `''` in the schema, that should be updated to be `comment` to match. https://github.com/WP-API/WP-API/blob/develop/lib/endpoints/class-wp-rest-comments-controller.php#L925

rachelbaker:
@joehoyle Another note: if we use `get_comment_type()` to solve the lack of a comment_type for 'comment' then only the default comment types (trackback, pingback, comment) can be added or updated via the REST API.  Which means we should not support custom comment types at all.

rachelbaker:
Ugghh... ran into #2758

joehoyle:
@rachelbaker I also had to modify `prepare_item_for_database` to set type to an empty string in the event of `comment`. Turns out `wp_insert_comment( [ type => 'comment' ] )` breaks things a lot!

joehoyle:
@WP-API/amigos #reviewmerge

rachelbaker (approved):

PR #2756, opened 2016-10-03
Add `?{taxonomy}_exclude=` query parameter to posts

kadamwhite:
This mirrors our existing support for `?{taxonomy}=` filtering in the posts controller (which allows querying for only records with are associated with any of the provided term IDs for the specified taxonomy) by adding an equivalent `_exclude` variant to list IDs of terms for which associated posts should NOT be returned.

For categories _e.g._, `/wp/v2/posts?categories[]=7&categories[]=9` would return a post in categories 7 _or_ 9; this PR adds support for the inverse query `?categories_exclude[]=7&categories_exclude[]=9`, meaning "only return posts that are associated with neither category 7 or category 9."

`{term base}_exclude` is supported for any taxonomy registered with the API, and is set within the same loop inside the controller as the existing term include parameters.

While using the existing categories filter with a negative number was considered ([see slack discussion](https://wordpress.slack.com/archives/core-restapi/p1475525429004981)), `{base}_exclude` maintains more consistency with other exclusion filters present in the API such as `parent_exclude` (and is much easier to implement, to boot).


kadamwhite:
@WP-API/amigos #reviewmerge

joehoyle (approved):

Issue #2696, opened 2016-09-13
API should send and accept `null` for undefined values


kadamwhite:
See #2657 for the decision on `null` handling: `null` should be sent by the API whenever a value is unknown or not yet set.

A prime example of this is `post_date_gmt`, as described in #2303, which will be `0000-00-00 00:00:00` (an invalid date) for posts newly created via WP-Admin or via this API: if this value is sent to a client literally the API is providing an invalid value. `null` should be sent instead.

Correspondingly, `null` should be _accepted_ by the API for `post_date_gmt` (and other fields that display this issue), so that a client can always turn the response for a resource around and send it back without further transformation: returning the API's output JSON as unchanged input should never trigger an error.

adamsilverstein:
I am starting to work on this issue by writing some unit tests to verify current behavior. 

This ticket should probably be broadened to "The API should accept returned values as valid for updates of the same object." For dates this means that if a null value is returned a null value should be accepted. Currently this appears to be the case. 

For dates, I think we are set for null values... Since https://github.com/WP-API/WP-API/commit/4dc05f35 the API returns null for the '0000-00-00 00:00:00' that WordPress returns for draft posts. I also tested sending in a null value (which you can only do from PHP as far as I can tell). I'm working on unit tests(s) to demonstrate this.

Another edge case I am digging into is slugs. When you create a draft post with no title, the API returns a slug of  "" (empty string). Sending back the same empty string for the slug when trying to update the draft fails because a blank string is an invalid slug value. Here similarly the API should probably return a null value for the invalid/blank slug value (as well as accepting that on input). 

I will continue to work on this issue this week.

adamsilverstein:
After digging into this and writing a few tests, the only issue I found was slugs. When initially creating a post via the api, the slug is returned as an empty string `""`. Taking the post data returned from the API and sending it back was throwing an error because of the empty slug. I added an additional validation check for the slug value and this resolved the issue. Without this change the `test_update_draft_post_with_returned_draft_post_data` test fails.

adamsilverstein:
After updating to to the latest develop branch, I verified that everything _is working as expected_ in terms of unset, empty or null values here - I added a few tests in https://github.com/WP-API/WP-API/pull/2763 to validate. We can address any specific cases that were missed if they arise.

PR #2763, opened 2016-10-06
Add tests to ensure post controller accepts the same values it returns when creating a post

adamsilverstein:
- Adds some tests to ensure API accepts null date data and empty slug data for draft posts.
- Adds a test to ensure API accepts back all values it returns when creating a new post as valid.
- Adds an is_object test to slug validation in `prepare_item_for_database`, preventing empty slug objects values from being added to the post


adamsilverstein:
Fixes https://github.com/WP-API/WP-API/issues/2696

lib/endpoints/class-wp-rest-posts-controller.php
```diff
@@ -797,7 +797,7 @@ protected function prepare_item_for_database( $request ) {
 			}
 		}
 		// Post slug.
-		if ( ! empty( $schema['properties']['slug'] ) && isset( $request['slug'] ) ) {
+		if ( ! empty( $schema['properties']['slug'] ) && isset( $request['slug'] ) && ! empty( $request['slug'] ) && ! is_object( $request['slug'] ) ) {
```
adamsilverstein: A blank slug sent to the api winds up here as an empty object which was passing the empty test, I didn’t figure out why. Without this change, the `test_update_draft_post_with_returned_draft_post_data` test below fails.

joehoyle: Hmm @adamsilverstein did you figure out where this empty object is coming from?

adamsilverstein: Not yet, I'll dig into it @joehoyle. I would expect a blank string here.

adamsilverstein: @rmccue @joehoyle I tracked this down to our use of the `sanitize_title` to sanitize the slugs. `sanitize_title` takes an optional second parameter as a 'fallback'. in class-wp-rest-controller:L807 in `sanitize_params` where we use `call_user_func`, we are passing the value, then also passing the request object and the argument key.

In this case `sanitize_title` sees an empty string as failing and returns the fallback, the request object, and the post update fails.

Do you know why the request and key are added to the sanitization callback? can those be removed? Going to dig into when they were added.

https://core.trac.wordpress.org/browser/trunk/src/wp-includes/rest-api/class-wp-rest-request.php#L807

adamsilverstein: ps. I removed the is_object test and now my `test_update_draft_post_with_empty_slug` test fails

adamsilverstein:
@joehoyle I think we need to change our sanitize callback to fix this issue.

In https://github.com/WP-API/WP-API/pull/2763/commits/a3c95b5884c3d22344233da8758a61e17248c1b2 I added a new `rest_sanitize_sanitization callback for the slug that ignores the passed $request and $key, calling`sanitize_title`on the value alone, and properly returning a blank string when a blank string is sanitized. I was mixed about whether this should be named as is or`rest_sanitize_title`

plugin.php
```diff
@@ -487,3 +487,16 @@ function rest_is_boolean( $maybe_bool ) {
 		return false;
 	}
 }
+
+/**
```
joehoyle: Hrmmm I feel like we already added this somewhere, @rmccue is that true?

rmccue: Indeed; it's in the base controller as `WP_REST_Controller->sanitize_slug`

rmccue: Which, in fact, is already used above, so not sure why this PR is changing it?

adamsilverstein: @joehoyle @rmccue I think you were one step ahead of me, I just updated develop and see the new function. updating the PR, all I have to add now is the new tests, everything checks out with this fix in place.

adamsilverstein:
After updating develop, everything checks out for this issue. I'm leaving the test in place for consideration because they validate that everything works as expected.

rmccue:
@adamsilverstein The tests are failing here; does this need `develop` merged in again to resolve?

adamsilverstein:
@rmccue not sure its my branch? I can refresh patch if needed, I'm seeing this in Travis failure: `PHP Fatal error:  Cannot use object of type stdClass as array in /home/travis/build/WP-API/WP-API/lib/fields/class-wp-rest-meta-fields.php on line 104`

PR #2765, opened 2016-10-06
Add meta fields

rmccue:
Ported from https://github.com/WP-API/wp-api-meta-endpoints/pull/15, see discussion there. This is just the integration of that back into the API core.


joehoyle:
it’s been noted that we should merge the meta and settings PRs for testing in the plugin, there’s a couple of outstanding edgecases which will be incorporated shortly.

joehoyle (approved):

PR #2770, opened 2016-10-06
Exclude unsupported query parameters in posts controller

### LINKED ISSUE #2767: `parent` is incorrectly used when querying posts if not in schema
Splitting from #2201.

rmccue:
Fixes #2767.


lib/endpoints/class-wp-rest-posts-controller.php
```diff
@@ -94,22 +94,32 @@ public function get_items( $request ) {
...
+		$args['offset']         = $request['offset'];
+		$args['order']          = $request['order'];
+		$args['orderby']        = $request['orderby'];
+		$args['paged']          = $request['page'];
+		$args['post__in']       = $request['include'];
+		$args['post__not_in']   = $request['exclude'];
+		$args['posts_per_page'] = $request['per_page'];
+		$args['name']           = $request['slug'];
+		$args['post_status']    = $request['status'];
+		$args['s']              = $request['search'];
```
danielbachhuber: We should have conditionals around these too, so a user can deregister the parameter if they don't want to support the query (or if the resource isn't meant to implement a given field like `slug`)

danielbachhuber (changes requested):

lib/endpoints/class-wp-rest-posts-controller.php
```diff
@@ -94,42 +94,74 @@ public function get_items( $request ) {
...
+		}
+		if ( isset( $registered['menu_order'] ) ) {
+			$args['menu_order'] = $request['menu_order'];
+		}
+		if ( isset( $registered['parent'] ) ) {
+			$args['post_parent__in'] = $request['parent'];
+		}
+		if ( isset( $registered['parent_exclude'] ) ) {
+			$args['post_parent__not_in']  = $request['parent_exclude'];
+		}
```
kadamwhite: What about possibly changing this section to something like 

``` php
$parameter_mappings = array(
  'offset' => 'offset',
  'order' => 'order',
  'orderby' => 'orderby',
  'page' => 'page',
  'include' => 'post__in',
  'exclude' => 'post__not_in',
  'slug' => 'name',
  'status' => 'post_status',
  'search' => 's',
  'author' => 'author__in',
  'author_exclude' => 'author__not_in',
  'menu_order' => 'menu_order',
  'parent' => 'post_parent__in',
  'parent_exclude' => 'post_parent__not_in',
)
foreach ( $parameter_mappings as $query_param => $wp_query_param ) {
  if ( isset( $registered[ $query_param ] ) ) {
    $args[ $wp_query_param ] = $request[ $query_param ];
  }
}
```

to make it a little less repetitive?

rachelbaker: that would also be so much easier to read.

lib/endpoints/class-wp-rest-posts-controller.php
rachelbaker: Addressed in 62a75b9

rachelbaker:
Just a note that after this PR gets merged we will need to implement in the other endpoints as well.

lib/endpoints/class-wp-rest-posts-controller.php
```diff
@@ -94,42 +94,74 @@ public function get_items( $request ) {
...
-		$args['orderby']              = $request['orderby'];
-		$args['paged']                = $request['page'];
-		$args['post__in']             = $request['include'];
-		$args['post__not_in']         = $request['exclude'];
-		$args['name']                 = $request['slug'];
-		$args['post_parent__in']      = $request['parent'];
-		$args['post_parent__not_in']  = $request['parent_exclude'];
-		$args['post_status']          = $request['status'];
-		$args['s']                    = $request['search'];
+		$registered = $this->get_collection_params();
```
rachelbaker, line 98: @rmccue Seems like this check against the registered params should happen in `get_allowed_query_vars()`, there is some duplicate logic there.

danielbachhuber (approved):

rachelbaker:
This looks to be a pretty high-priority ticket if we want to get this done prior to a merge.  The approach set in this PR will need to be duplicated in all of our other controllers.

kadamwhite:
Assigning myself to take a stab at refreshing this, will be coordinating with @rachelbaker to get more feedback

rachelbaker (changes requested):
1. I agree with @kadamwhite's suggestion regarding using a loop to reduce the long series of  isset()  conditionals. 
2. Because we have to do this collection parameter processing across all of our Controllers, it would be _ideal_ to think about how we want to define the mapped parameters in the Controllers.  With a protected property or method?  This would be cleaner than shoving this logic into the `get_items()` logic across the board.
3. Would it make more sense for the check against the registered collection params to happen in `prepare_items_query()` or `get_allowed_query_vars()`?
4. (Very related to #3) The query parameter handling logic in the Posts Controller is already difficult to follow, is there anything we can simplify?

kadamwhite:
Added a commit to address (1) of @rachelbaker's feedback. Thoughts on the remainder below:
- `get_allowed_query_vars` doesn't feel like the right place for this because the set of allowed query vars is independent of the vars that are passed for any given request, and the goal here is to ensure a given request does not accept or consider parameters that are not allowed.
- `prepare_items_query` feels like the "right" place to do this: except that by that point we've set WP-internal properties on `$args`, while the registered parameters list is built off of the API-oriented parameters (`include` instead of `post__in`, e.g.). It almost feels like what we want is a `prepare_items_request` or something that pre-processes the `request` object to strip values that aren't whitelisted. I don't think that solves the need for much of this code, though, just a couple of the statements in conditionals below; our handling of parameters is non-trivial!
- Regarding how to broadly define this type of functionality across controllers... I'm still thinking about it.

kadamwhite:
Open question:
- Should "context" be treated in this way too, or should we presume it is present? Seems to be more "global" than these others.

The only params not currently being checked against registered are the "context", which I have an open question on above; and taxonomies, for which a fix would be this diff:

``` diff
diff --git lib/endpoints/class-wp-rest-posts-controller.php lib/endpoints/class-wp-rest-posts-controller.php
index 86ef4da..45acf9b 100755
--- lib/endpoints/class-wp-rest-posts-controller.php
+++ lib/endpoints/class-wp-rest-posts-controller.php
@@ -191,19 +191,19 @@ class WP_REST_Posts_Controller extends WP_REST_Controller {

        $taxonomies = wp_list_filter( get_object_taxonomies( $this->post_type, 'objects' ), array( 'show_in_rest' => true ) );
        foreach ( $taxonomies as $taxonomy ) {
-           $base = ! empty( $taxonomy->rest_base ) ? $taxonomy->rest_base : $taxonomy->name;
-           $tax_exclude = $base . '_exclude';
+           $tax_base = ! empty( $taxonomy->rest_base ) ? $taxonomy->rest_base : $taxonomy->name;
+           $tax_exclude = $tax_base . '_exclude';

-           if ( ! empty( $request[ $base ] ) ) {
+           if ( isset( $registered[ $tax_base ] ) && ! empty( $request[ $tax_base ] ) ) {
                $query_args['tax_query'][] = array(
                    'taxonomy'         => $taxonomy->name,
                    'field'            => 'term_id',
-                   'terms'            => $request[ $base ],
+                   'terms'            => $request[ $tax_base ],
                    'include_children' => false,
                );
            }

-           if ( ! empty( $request[ $tax_exclude ] ) ) {
+           if ( isset( $registered[ $tax_exclude ] ) && ! empty( $request[ $tax_exclude ] ) ) {
                $query_args['tax_query'][] = array(
                    'taxonomy'         => $taxonomy->name,
                    'field'            => 'term_id',
```

but this depends on #2802.

rachelbaker:
> Should "context" be treated in this way too, or should we presume it is present? Seems to be more "global" than these others.

@kadamwhite Great question.  Sounds like **ideally** we want to prevent `context` from being de-registered (just a "nice to have" suggestion for a future enhancement), but in the meantime we should fallback to `view` if it is not present.

rachelbaker (approved):

Issue #2771, opened 2016-10-07
How To Delete Some Info From API


bestony:
I Want to Only show title,author,id exclude content to save bandwidth for some bad network

kosso:
See here to learn how to modify the responses with a plugin script. 

http://v2.wp-api.org/extending/modifying/

Or you could create your own namespaced endpoints and deliver what you need. 

http://v2.wp-api.org/extending/adding/

bestony:
Well,I don't want to make a ext . i just want to hide the content for save bandwidth

kosso:
I agree that it would be nice to filter the fields we need, but see the later comments on #446 and #572 and it seems that it's the only way to go. For now.

bestony:
well, i can make new namespace , but i also want it can add to do list

kosso:
I found this useful : http://boris.kuzmanov.ninja/wp-rest-api-modifying-json-response/ 

But I would recommend a custom endpoint. As they mention on the [Modifying Responses](http://v2.wp-api.org/extending/modifying/#important-note-about-changing-responses) page, it can be unwise to remove fields from the standard wp/v2 responses, as other clients might expect the fields to be present.

kosso:
How about something like this? Using a custom `_compact` parameter: 

```
function allow_compact_post( $data, $post, $request ) {
    // return compact data when _compact GET parameter exists 
    if(isset($_GET['_compact'])){
        return [
            'id'        => $data->data['id'],
            'title'     => $data->data['title']['rendered'],
            'link'      => $data->data['link'],
        ];
    }
    return $data;
}
add_filter( 'rest_prepare_post', 'allow_compact_post', 10, 3 );
```

then hit `/wp-json/wp/v2/posts?_compact`

kadamwhite:
@kosso @bestony I'm closing this as superseded by https://core.trac.wordpress.org/ticket/38131 which proposes a `?fields=id,title,link` query parameter that can be used to restrict results; that interface is used by the WordPress.com API and I've proposed it for inclusion here in #2773 but the change would likely need to occur in within the infrastructure already committed to core.

kadamwhite:
Thanks for raising this, by the way; it's been weighing on my mind that we don't have an interface for it and this was a good push to try to do something about it :)

kosso:
👍

bestony:
:+1:

Issue #2509, opened 2016-05-23
There's any way to filter the GET /posts request?


joanlopez:
Hi mates,

I'm building a personal Wordpress. It will be accessible from a typical url: blog.example.org.

In my case, i'm using the WP-API because I wanna show a random post from my blog on another site: example.org.

So, my question is:

There is any way to get only the ids of the available (published) posts?

Of course, I could retrieve all the posts with a simply GET /posts call, but this option implies a lot of unnecessary data because it returns all the data from all the posts while I only need one post, and apart from the performance, it could be critical for mobile devices without a wifi connection.

Also another nasty solution is to have a hardcoded array on my frontend, pick one randomly and make the GET /posts/id request only for the chosen one. I'm not sure which option is worst, because this implies that I have to update my frontend on each post publication..

If there's not implemented solution yet. Are you open to my contribution for this task? Or I should make my own plugin?

Many thanks!

danielbachhuber:
> There is any way to get only the ids of the available (published) posts?

There isn't, no.

At some point in the future, we may include a flag to include or exclude specific fields: https://github.com/WP-API/WP-API/issues/572

> If there's not implemented solution yet. Are you open to my contribution for this task? Or I should make my own plugin?

You'd be best writing your own plugin for this, at this point.

Issue #2350, opened 2016-03-08
jmespath support to limit responses


elyobo:
Feature request -

The documentation sensibly suggests that users [avoid modifying standard responses](http://v2.wp-api.org/extending/modifying/#important-note-about-changing-responses) to avoid problems when using other clients that expect those fields to be present.  The AWS CLI supports this with the [--query argument](https://docs.aws.amazon.com/cli/latest/userguide/controlling-output.html#controlling-output-filter) which makes use of [JMESPath](http://jmespath.org/) ([PHP library](https://github.com/jmespath/jmespath.php)) to allow clients to better filter the data that they want without modifying the underlying data format (the standard format would remain for other clients).

Implementing support for this in the REST plugin would reduce the need for users to remove fields from the data.  If there is an appropriate filter that fires just before the rest API sends the response then this could easily be packaged into a plugin as well.

rmccue:
See #572. We're unlikely to add this to the main plugin right now, as this is plugin territory for now.

You can use either the `rest_post_dispatch` or `rest_pre_serve_request` filters for this, depending on what you want to do.

elyobo:
Thanks for the pointers.

elyobo:
Proof of concept at https://github.com/elyobo/wp-rest-jmespath

elyobo:
If you have any feedback let me know, otherwise please add it to the list of compatible plugins.

rmccue:
Awesome! If you pop us a pull request on the [docs repo](https://github.com/WP-API/docs-v2) with the description you'd like, happy to add that for you :)

elyobo:
Thanks @rmccue PR added at https://github.com/WP-API/docs-v2/pull/126 now that the plugin is published on wp.org.

Issue #1530, opened 2015-09-06
Specify which post data fields to include or exclude


spectodesign:
How can I specify which post data fields to include or exclude. E.g. for a call to list a collection of posts I don't need the content field, for a detail view of a post I do.

rmccue:
This isn't possible in the core REST API; see #572.

PR #2773, opened 2016-10-07
Add "?fields" query parameter to return only specific fields in response

kadamwhite:
This PR builds upon discussion in #2771, #572, #2350, #2509, #1530, #1371, #446, #827, and the trac ticket [38131](https://core.trac.wordpress.org/ticket/38131), proposing a solution implemented within this plugin's wp-rest-controller class.

The patch implements a `?fields=` filter parameter that takes a comma-separated list of field names, e.g. "id" and "title" via `?fields=id,title`. It mirrors the fields parameter implemented within the [WordPress.com API](https://developer.wordpress.com/docs/api/1.1/get/sites/%24site/posts/):

| Parameter | Type | Description |
| --- | --- | --- |
| fields | (string) | Optional. Returns specified fields only. Comma-separated list. Example: fields=ID,title |

This interface was previously implemented in [this plugin](https://wordpress.org/plugins/wp-rest-api-filter-fields/), and could still be implemented trivially as a plugin; however, referencing the [2016 Ericsson Mobility Report](https://www.ericsson.com/res/docs/2016/mobility-report/ericsson-mobility-report-feb-2016-interim.pdf) and the democratic focus of WordPress, transferring as little data as is necessary for a given interface is a responsibility I submit that we should take on as application developers, and that _should_ therefore be a first-party capability of the API, as it is for .com's.

Introducing this field has no risk of breaking the contract between the API and clients you do not control, as the `fields` list would only alter the response when provided so each client can tailor its own responses to the subset of data relevant to that interface.

Open questions:
- Is there a less repetitive way this could be implemented across controllers, or is this duplication within each controller necessary?
- How would I register `fields` as a globally-applicable parameter for exposure in the schema, whitelisting, etc?
- How could/How should we restrict this to GET requests only?
- What degree of per-controller testing is needed?

### DIFF
--- a/lib/endpoints/class-wp-rest-attachments-controller.php
+++ b/lib/endpoints/class-wp-rest-attachments-controller.php
@@ -276,8 +276,9 @@ class WP_REST_Attachments_Controller extends WP_REST_Posts_Controller {
 		}
 
 		$context = ! empty( $request['context'] ) ? $request['context'] : 'view';
-
 		$data = $this->filter_response_by_context( $data, $context );
+		$fields = ! empty( $request['fields'] ) ? explode( ',', $request['fields'] ) : array();
+		$data = $this->filter_response_by_fields( $data, $fields );
 
 		// Wrap the data in a response object
 		$response = rest_ensure_response( $data );
--- a/lib/endpoints/class-wp-rest-comments-controller.php
+++ b/lib/endpoints/class-wp-rest-comments-controller.php
@@ -588,6 +588,8 @@ class WP_REST_Comments_Controller extends WP_REST_Controller {
 		$context = ! empty( $request['context'] ) ? $request['context'] : 'view';
 		$data = $this->add_additional_fields_to_object( $data, $request );
 		$data = $this->filter_response_by_context( $data, $context );
+		$fields = ! empty( $request['fields'] ) ? explode( ',', $request['fields'] ) : array();
+		$data = $this->filter_response_by_fields( $data, $fields );
 
 		// Wrap the data in a response object
 		$response = rest_ensure_response( $data );
--- a/lib/endpoints/class-wp-rest-controller.php
+++ b/lib/endpoints/class-wp-rest-controller.php
@@ -210,6 +210,28 @@ abstract class WP_REST_Controller {
 	}
 
 	/**
+	 * Filter a response to include only a subset of response fields
+	 *
+	 * @param array $data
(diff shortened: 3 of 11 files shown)


kadamwhite:
This could and likely should be implemented on the infrastructure level; I could use direction around where to do so, if that is the case.

The 80/20 rule was referenced in a prior ticket on this as justification for this being plugin territory and since I closed the original issue, I have come to disagree: 100% of API-consuming applications are more respectful of their users' bandwidth, time and money if they can request only the data they need.

mnelson4:
We did something similar in the [Event Espresso REST API too](https://github.com/eventespresso/event-espresso-core/blob/master/docs/C--REST-API/ee4-rest-api-GET-including-specific-fields-and-related-entities-in-results.md) and our mobile app developers have used the feature quite a bit (they're quite concerned about avoiding pulling too much data). So +1 to adding this to core.

> Is there a less repetitive way this could be implemented across controllers, or is this duplication within each controller necessary?

Maybe you could perform the filtering inside `WP_REST_Server::dispatch()`, or somewhere it could filter any response. Because needing to call `$this->filter_response_by_fields()` on every endpoint is a bit tedious, and easy to forget if you're adding new endpoints.

kadamwhite:
I believe @rmccue is investigating how this could be implemented on the infrastructure level, but an open question on what we have here already:

If my `/wp/v2/posts` response is 39.1kb to start (as it is with my test data), `?fields=title,id` will drop the response to 11.5kb. That's less than a third of the response size: not bad! However, my implementation _does not affect the `_links` property_, which is actually the majority of the response body; removing `_links` would on the one hand make the responses _very small_, which is good for bandwidth; on the other it removes the traverseability of the response. Since this is purely opt-in behavior that may be desirable, but I want to discuss it specifically:

Is removing links something that should be accomplished with this query parameter, _e.g._ if you pass `?fields=id,title,_links` they will appear and if you pass `fields=id,title` they will not; or should that be something that must be done mindfully with a separate parameter?

kadamwhite:
Based on feedback from @danielbachhuber and @rmccue that this is still an infrastructure level change, and that doing it with a filter at that level would avoid a lot of this duplication, I'm closing this in favor of a patch on https://core.trac.wordpress.org/ticket/38131 that Ryan helped point me towards creating. I've attempted to capture these discussion points in that ticket and will be pushing to keep that conversation going so we can find a good resolution for this.

PR #2313, opened 2016-02-24
Ensure `GET /wp/v2/categories` with out of bounds offset doesn't return results

danielbachhuber:
Fixes #2309

PR #2779, opened 2016-10-09
Remove temporary fix for upstream bug 35614.

boonebgorges:
PR #2313 included a fix for a core bug that caused `get_terms()` to return
results when the `offset` parameter was greater than the total number
of terms available. This bug was fixed in WP 4.5 and the workaround in
WP-API is no longer needed.


rachelbaker (approved):

rachelbaker:
Thank you @boonebgorges

Issue #2740, opened 2016-09-28
Use WP_Term_Query for querying terms


danielbachhuber:
It's new and shiny as of 4.6.

We should be able to remove `WP_REST_Terms_Controller->get_terms_for_post()` in the process.

boonebgorges:
I've moved this to Trac https://core.trac.wordpress.org/ticket/38504#ticket, so the current ticket can probably be closed.

PR #2780, opened 2016-10-09
Use `WP_Term_Query` to fetch a post's terms.

boonebgorges:
See #2740.

The most parsimonious approach would've been to replace the `get_terms()` call with `new WP_Term_Query`, with the 'object_id' argument set to the value of`$request['post']`. But this would be a change in behavior for the term controller, since it'd mean bypassing the 'get_terms' filter. If we're sticking with the 'get_terms' filter, it seems consistent to use `wp_get_object_terms()` for object-specific queries, which is what I've done in this PR. Feedback welcome on this decision.

### DIFF
--- a/tests/test-rest-tags-controller.php
+++ b/tests/test-rest-tags-controller.php
@@ -665,6 +665,31 @@ class WP_Test_REST_Tags_Controller extends WP_Test_REST_Controller_Testcase {
 		$wp_rest_additional_fields = array();
 	}
 
+	public function test_object_term_queries_are_cached() {
+		global $wpdb;
+
+		$tags = $this->factory->tag->create_many( 2 );
+		$p = $this->factory->post->create();
+		wp_set_object_terms( $p, $tags[0], 'post_tag' );
+
+		$request = new WP_REST_Request( 'GET', '/wp/v2/tags' );
+		$request->set_param( 'post', $p );
+		$response = $this->server->dispatch( $request );
+		$found_1 = wp_list_pluck( $response->data, 'id' );
+
+		unset( $request, $response );
+
+		$num_queries = $wpdb->num_queries;
+
+		$request = new WP_REST_Request( 'GET', '/wp/v2/tags' );
+		$request->set_param( 'post', $p );
+		$response = $this->server->dispatch( $request );
+		$found_2 = wp_list_pluck( $response->data, 'id' );
+
+		$this->assertEqualSets( $found_1, $found_2 );
+		$this->assertSame( $num_queries, $wpdb->num_queries );
+	}
+
 	public function additional_field_get_callback( $object, $request ) {
 		return 123;
 	}
--- a/lib/endpoints/class-wp-rest-terms-controller.php
+++ b/lib/endpoints/class-wp-rest-terms-controller.php
@@ -138,36 +138,28 @@ class WP_REST_Terms_Controller extends WP_REST_Controller {
 		 */
 		$prepared_args = apply_filters( "rest_{$this->taxonomy}_query", $prepared_args, $request );
 
-		// Can we use the cached call?
-		$use_cache = ! empty( $prepared_args['post'] )
-			&& empty( $prepared_args['include'] )
-			&& empty( $prepared_args['exclude'] )
-			&& empty( $prepared_args['hide_empty'] )
-			&& empty( $prepared_args['search'] )
-			&& empty( $prepared_args['slug'] );
-
-		if ( ! empty( $prepared_args['post'] )  ) {
-			$query_result = $this->get_terms_for_post( $prepared_args );
-			$total_terms = $this->total_terms;
+		if ( ! empty( $request['post'] ) ) {
+			$query_result = wp_get_object_terms( $request['post'], $this->taxonomy, $prepared_args );
 		} else {
 			$query_result = get_terms( $this->taxonomy, $prepared_args );
-
-			$count_args = $prepared_args;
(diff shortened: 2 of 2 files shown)


boonebgorges:
Another note: this PR includes #2779, since the offending code addressed in that ticket is removed altogether here. Apologies for the crossed streams.

danielbachhuber:
> The most parsimonious approach would've been to replace the `get_terms()` call with new `WP_Term_Query`, with the `'object_id'` argument set to the value of `$request['post']`.

I think we should do whatever you think is best, and put our best foot forward. We've made breaking changes in the past, so I don't see that as a problem.

joehoyle:
@boonebgorges does this mean we no longer get the object caching on a post's terms?

boonebgorges:
> I think we should do whatever you think is best, and put our best foot forward. We've made breaking changes in the past, so I don't see that as a problem.

Thinking more about it: As a plugin author, I'd expect that my plugin filtering `'get_object_terms'` would affect all instances of object-term queries, API included. So the approach in the PR seems like the right one to me, even though it results in more code.

> does this mean we no longer get the object caching on a post's terms?

No. `wp_get_object_terms()` queries are cached starting in WP 4.7, since they now run through `WP_Term_Query`. (The caching mechanism is a bit different - it's the 'terms' cache group rather than `{$taxonomy}_relationships`.) See https://core.trac.wordpress.org/changeset/38667

I see that the PR is no longer applying properly because of #2779. I'll fix and resubmit.

boonebgorges:
I've updated the PR:
- Fixed merge conflicts
- Fixed variable name typo
- Added test that demonstrates that the cache is being hit for object-term queries

The last item required a core bugfix, so the test will only pass against WP trunk https://core.trac.wordpress.org/changeset/38784

joehoyle (approved):

tests/test-rest-tags-controller.php
```diff
@@ -665,6 +665,31 @@ public function test_additional_field_update_errors() {
...
+
+		$num_queries = $wpdb->num_queries;
+
+		$request = new WP_REST_Request( 'GET', '/wp/v2/tags' );
+		$request->set_param( 'post', $p );
+		$response = $this->server->dispatch( $request );
+		$found_2 = wp_list_pluck( $response->data, 'id' );
+
+		$this->assertEqualSets( $found_1, $found_2 );
+		$this->assertSame( $num_queries, $wpdb->num_queries );
```
rmccue, line 690: This is currently failing in the tests.

rmccue (changes requested):

tests/test-rest-tags-controller.php
kadamwhite: @rmccue this is the test Boone indicated would only pass against trunk; are we using trunk in our test suite?

This patch looks good to me, other than I wonder if we can gate this test somehow for now.

joehoyle:
Hmm this is causing more issues, as the WP_Term_Query changes in 4.7 are making backwards compatibility for 4.6 break in several places. It's not just the new unit test that fails under 4.6, we have others ones too, which the change to using WP_Term_Query. I feel like we should be supported the latest release of WordPress, and right now that means supporting 4.6, so I'm -1 on merging this right now.

boonebgorges:
Tests were passing locally for me with WP nightly. Not sure why it's failing on Travis. I'll look into it more carefully.

> Hmm this is causing more issues, as the WP_Term_Query changes in 4.7 are making backwards compatibility for 4.6 break in several places. It's not just the new unit test that fails under 4.6, we have others ones too, which the change to using WP_Term_Query

@joehoyle When you say that "the change to using `WP_Term_Query`" is causing BC breaks, do you mean that (a) the change in the current PR breaks compatibility with sites running older versions of WP, or do you mean that (b) the change to `WP_Term_Query` in core is causing other BC breaks in the API?  If (b), please provide details. If (a), it's fine with me if the PR is not merged, if your rule is to support the current stable WP - but it should be part of any core merge, where BC with older versions of WP will not be an issue.

joehoyle:
@boonebgorges right, it's (b), sorry for not being clear. this isn't a backwards compat issue for core, it's because with this PR, we are switching old code to WP_Term_Query even under 4.6, but expecting it to do the same things as under 4.7, which isn't the case, of course.

It makes sense for this to be merged if we merge the API into core, as that code can then of course be dependant on 4.7, however as it stands right now, I'd rather not release another beta of the plugin a couple of days and break things for sites on 4.6 using the plugin.

joehoyle:
FYI tests _do_ pass on nightly.

boonebgorges:
Thanks, @joehoyle. I asked in Slack a couple days ago about your policy regarding fixes that depend on core patches, but didn't get a response :) Your policy sounds fine to me.

Tests don't seem to be passing on all versions of PHP. See eg https://travis-ci.org/WP-API/WP-API/builds/167090500. I don't understand why. There could be a race condition or something that's causing a cache miss (hard to believe that our tests are not bulletproof, I know)

joehoyle:
I think those tests were failing due to running on Nightly too early, I re-ran them and they all pass now.

boonebgorges:
I think that's called an "external race condition"

kadamwhite:
Marking this "blocked on core" since it depends on 4.7; maybe an over-extrapolation of the ticket but feels right :)

boonebgorges:
I've migrated to Trac. https://core.trac.wordpress.org/ticket/38504#ticket

PR #2789, opened 2016-10-10
Correct copy pasta @see url in Posts Controller query args filter

rachelbaker:
The query args in the Posts Controller can be found at https://developer.wordpress.org/reference/classes/wp_query/  NOT https://developer.wordpress.org/reference/classes/wp_user_query/


rachelbaker:
@WP-API/amigos #reviewmerge please

joehoyle (approved):

PR #2794, opened 2016-10-11
Add WordPress Dependency

joehoyle:


rmccue:
Thanks for the PR! With the REST API, we follow the [WordPress coding standards](https://make.wordpress.org/core/handbook/coding-standards/php/); can you please update the code here to match those?

pento:
I have concerns about the performance of the code in this PR. I think it needs to be significantly refactored before it can be considered for merge.

samwilson:
1,613 files? Is this all of Wordpress or something?

Wouldn't it be better to add it as a Composer dependency?

aaronjorbin:
Do you think it might make more sense to include Drupal instead?  Have all the options been explored?

samwilson:
Hm yes, I see what you mean @aaronjorbin! Could all of MediaWiki please be included? It's not huge, and might, umm, help? :-P

rmccue:
I'm locking this topic, there are too many Australians here.

rmccue:
I apologise, must be something in the water here in NZ. Reopened.

samwilson:
Ahg you're just rubbing it in that you've got nicer water than Perth! :-)

pdewouters:
Isn't there already a PR to include the Internet? it would make this one redundant.

kadamwhite:
Implementing a highly usable, back-compatible server-side application for WYSIAlmostWYG content management for content already addressable by API endpoints is deemed out of scope for this project.

PR #2795, opened 2016-10-11
Remove wp_die in comment creation

websupporter:
Solves https://github.com/WP-API/WP-API/issues/1660 and https://github.com/WP-API/WP-API/issues/1784

With https://core.trac.wordpress.org/ticket/36901 in core, we can now use `wp_allow_comment()` without running into `wp_die()` \o/

### DIFF
--- a/tests/test-rest-comments-controller.php
+++ b/tests/test-rest-comments-controller.php
@@ -1048,6 +1048,7 @@ class WP_Test_REST_Comments_Controller extends WP_Test_REST_Controller_Testcase 
 	}
 
 	public function test_create_item_duplicate() {
+		global $wp_version;
 		if ( version_compare( $wp_version, '4.7-alpha', '<' ) ) {
 			return $this->markTestSkipped( 'WordPress version not supported.' );
 		}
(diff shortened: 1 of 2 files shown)


tests/test-rest-comments-controller.php
```diff
@@ -1048,7 +1048,6 @@ public function test_create_comment_private_post_invalide_permission() {
 	}
 
 	public function test_create_item_duplicate() {
-		$this->markTestSkipped( 'Needs to be revisited after wp_die handling is added' );
```
rachelbaker, line 1051: @websupporter Because the test will only pass for WP trunk, you need a conditional here.

```
if ( version_compare( $wp_version, '4.7-alpha', '<' ) ) {
            return $this->markTestSkipped( 'WordPress version not supported.' );
}
```

rachelbaker:
@websupporter This should pass soon within the next ~12 hours, after the nightly is rebuilt

websupporter:
@rachelbaker. Thanks :) Do you see other issues in the PR?

rachelbaker (approved):

PR #2800, opened 2016-10-11
Error duplicate value update #2792

### LINKED ISSUE #2792: Updating a meta value which is the same as the current value returns error
I'm not sure if this is desired but it surprised me. If you try to update a meta value to the value currently in the database, the response is a 500 error. This is due to a check in update_metadata() which returns false if the values are the same. It might make a bit more sense to not return an error as you end up with the expected meta_value. [This checks and avoids the error.](https://github.com/tharsheblows/WP-API/commit/b6d986fcce6a2c16ac83beee72ea1ff7ee16b243)

tharsheblows:
This fixes #2792 by using the same check as in update_metadata, ie the check which is causing the error


rachelbaker (commented):
@tharsheblows Thank you for the patch.  Looks like there just some minor code style and indentation fixes needed so we can merge it.  See https://travis-ci.org/WP-API/WP-API/jobs/166817693

tharsheblows:
I have to admit I'm not quite sure what I'm doing. I used phpcs with the WordPress coding standards so if it should be something else, please let me know. :)

I forgot the main thing: that == vs === check is because I copied the _exact_ same check from update_metadata() in core because I want it to check exactly the same way. I can change it if you'd like but I think it should stay the same.

kadamwhite:
@tharsheblows Thanks for the explanation, that makes sense. However, in this case I think that core's out of date and we can rule that we should do the right thing; because `count()` will always return an integer there will never be any difference between `==` and `===` so we have the opportunity to maintain the coding standards without possibly altering functionality, and may as well take it.

rachelbaker:
Moved to #2805.

rachelbaker:
@tharsheblows The indentation on the tests file was all off, so I just cherry picked your commits (making sure you would still receive contributor credit) to a new branch and posted them along with the code standards fix in #2805.

PR #2804, opened 2016-10-11
Exclude unsupported params in all controllers (extends #2770)

kadamwhite:
This is a PR against #2770, **not `develop`** , and extends the approach taken there to apply to all other controllers in which get_items assigns query properties from requests.
- [x] Attachments (No get_items defined, extends WP_REST_Posts_Controller)
- [x] Comments
- [x] Post Statuses (No change, get_items assigns no values from $request)
- [ ] Post Types (Needed? Only "context" is used)
- [ ] Revisions (Needed? Only "parent" is used, and it's part of the URL string)
- [x] Taxonomies
- [x] Terms
- [x] Users

### DIFF
--- a/lib/endpoints/class-wp-rest-comments-controller.php
+++ b/lib/endpoints/class-wp-rest-comments-controller.php
@@ -115,42 +115,68 @@ class WP_REST_Comments_Controller extends WP_REST_Controller {
 	 * @return WP_Error|WP_REST_Response
 	 */
 	public function get_items( $request ) {
-		$prepared_args = array(
-			'author_email'    => isset( $request['author_email'] ) ? $request['author_email'] : '',
-			'comment__in'     => $request['include'],
-			'comment__not_in' => $request['exclude'],
-			'karma'           => isset( $request['karma'] ) ? $request['karma'] : '',
-			'number'          => $request['per_page'],
-			'post__in'        => $request['post'],
-			'parent__in'      => $request['parent'],
-			'parent__not_in'  => $request['parent_exclude'],
-			'search'          => $request['search'],
-			'offset'          => $request['offset'],
-			'orderby'         => $this->normalize_query_param( $request['orderby'] ),
-			'order'           => $request['order'],
-			'status'          => $request['status'],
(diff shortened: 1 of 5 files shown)


kadamwhite:
@rachelbaker This now includes almost all properties across all controllers. The duplication is evident as discussed in #2770 but I think I'm more in favor of duplication now and a mindful abstraction later, than prematurely introducing an insufficient abstraction now.

kadamwhite:
Open questions:
1. Should `context` be checked in this way, or are we able to treat it as "global"?
2. If a parameter is part of the route URL, do we need to check it in this manner?

These questions determine whether post types & revisions need similar treatment

lib/endpoints/class-wp-rest-comments-controller.php
```diff
@@ -115,43 +115,69 @@ public function get_items_permissions_check( $request ) {
...
-			'author__not_in'  => $request['author_exclude'],
+
+		// Retrieve the list of registered collection query parameters.
+		$registered = $this->get_collection_params();
+
+		// This array defines mappings between public API query parameters whose
+		// values are accepted as-passed, and their internal WP_Query parameter
+		// name equivalents (some are the same). Only values which are also
+		// present in $registered will be set.
+		$parameter_mappings = array(
```
danielbachhuber, line 126: What do you think about including this parameter mapping in the collection param definition itself?

kadamwhite: Do you mean in the schema? Certainly not opposed, but I guess I'm not sure how that would look, given what's there now.

joehoyle:
I think it's ok for this to stay where it is in get_items rather than moving to get_collection params, this is a procedural / controller thing, and `get_collection_params` is a bit more pure in what it does.

joehoyle:
> Revisions (Needed? Only "parent" is used, and it's part of the URL string)

I don't think so no.

> Post Types (Needed? Only "context" is used)

Like #2770 I don't think we need to solve context

joehoyle (approved):

PR #2805, opened 2016-10-12
Tharsheblows error duplicate value update

rachelbaker:
Replaces #2800 with corrected code formatting.


rachelbaker:
@WP-API/amigos #reviewmerge please

lib/fields/class-wp-rest-meta-fields.php
```diff
@@ -222,7 +222,19 @@ protected function update_meta_value( $object, $name, $value ) {
...
 		}
 
-		if ( ! update_metadata( $this->get_meta_type(), $object, wp_slash( $name ), wp_slash( $value ) ) ) {
+		$meta_type  = $this->get_meta_type();
+		$meta_key   = wp_slash( $name );
+		$meta_value = wp_slash( $value );
+
+		// Do the exact same check for a duplicate value as in update_metadata() to avoid update_metadata() returning false.
+		$old_value = get_metadata( $meta_type, $object, $meta_key );
+		if ( 1 === count( $old_value ) ) {
```
joehoyle, line 231: Does this need to be nested? Seems like it might be simpler as:

``` php
if ( 1 === count( $old_value ) && $old_value[0] === $meta_value ) ...
```

joehoyle: What if the meta value has more than a single item? Does `update_metadata` not return `false` ever in that case for duplicate entries?

tests/test-rest-post-meta-fields.php
```diff
@@ -230,6 +230,37 @@ public function test_set_value() {
 	}
 
 	/**
+	 * @depends test_get_value
+	 */
+	public function test_set_duplicate_single_value() {
```
joehoyle, line 235: Do we need to test duplicate with multiple, or is that covered in another test / PR?

rachelbaker: @joehoyle `update_metadata()` only does the duplicate check for single meta keys, for multiples it just goes ahead and performs the update.

lib/fields/class-wp-rest-meta-fields.php
rachelbaker: @joehoyle No, `update_metadata()` only returns `false` if you are updating a single meta value and the previous and new values are the same.

rachelbaker: @joehoyle RE: the simpler conditional... technically yes (although a micro-optimization at best) - but the structure used by @tharsheblows here DOES indeed match the conditional structure already in `update_metadata()`

joehoyle:
Sounds good!

joehoyle (approved):

PR #2810, opened 2016-10-12
Change priority for create_initial_rest_routes

websupporter:
Solves https://github.com/WP-API/WP-API/issues/2694

### DIFF
--- a/plugin.php
+++ b/plugin.php
@@ -129,7 +129,7 @@ require_once( dirname( __FILE__ ) . '/core-integration.php' );
 
 add_filter( 'init', '_add_extra_api_post_type_arguments', 11 );
 add_action( 'init', '_add_extra_api_taxonomy_arguments', 11 );
-add_action( 'rest_api_init', 'rest_register_settings', 0 );
+add_action( 'rest_api_init', 'rest_register_settings', 10 );
 add_action( 'rest_api_init', 'create_initial_rest_routes', 99 );
 
 /**
--- a/plugin.php
+++ b/plugin.php
@@ -130,7 +130,7 @@ require_once( dirname( __FILE__ ) . '/core-integration.php' );
 add_filter( 'init', '_add_extra_api_post_type_arguments', 11 );
 add_action( 'init', '_add_extra_api_taxonomy_arguments', 11 );
 add_action( 'rest_api_init', 'rest_register_settings', 0 );
-add_action( 'rest_api_init', 'create_initial_rest_routes', 0 );
+add_action( 'rest_api_init', 'create_initial_rest_routes', 99 );
 
 /**
  * Adds extra post type registration arguments.


plugin.php
```diff
@@ -130,7 +130,7 @@
 add_filter( 'init', '_add_extra_api_post_type_arguments', 11 );
 add_action( 'init', '_add_extra_api_taxonomy_arguments', 11 );
 add_action( 'rest_api_init', 'rest_register_settings', 0 );
```
rachelbaker: @websupporter Would there be a similar problem with the rest_register_settings hooked function taking place at priority 0?

kadamwhite:
Also, potentially naive question, @websupporter why 99 and not 11?

plugin.php
websupporter: Will have a look by tomorrow.

websupporter:
@kadamwhite From https://github.com/WP-API/WP-API/issues/2694 we know, we should not hook in to early. I thought to give some space, maybe someone has a reason to hook in with the priority 20. I think to remember I saw sometimes in core the priority 99 which I liked. :slightly_smiling_face:

plugin.php
websupporter: I do not quite get it, why `0` was used as priority. The `rest_register_setting()` only registers like `posts_per_page` but does not invoke critical REST API components like `create_inital_rest_routes()` does. So to me, it looks save, but if there is no reason for priority `0`, I think we should switch to default? /cc @joehoyle

joehoyle: @websupporter the Settings controller is initialized on `10`, so I set this to `0` to make sure settings were registered before the settings controller was spun up. However, thinking now, I don't think this was necessary, as the settings don't need to be present for the _initialization_ of the settings controller, so I think this could change to the default without issue.

rmccue: @joehoyle I think settings do need to be present, because the initialisation is also when the schema is generated?

websupporter:
@joehoyle @rmccue the settings controller is initialized in `create_initial_rest_routes()`, right?

```
        // Settings. 4.7+ only.
        global $wp_version;
        if ( version_compare( $wp_version, '4.7-alpha', '>=' ) ) {
            $controller = new WP_REST_Settings_Controller;
            $controller->register_routes();
        }
```

This would happen right now with priority 99. Before it was priority 0, so the same with `rest_register_settings()`. I did a quick `OPTIONS wp/v2/settings` and the settings appear in the current schema.

joehoyle:
Ahh hmm, but I don't know if this will be in the `wp-json` index, @websupporter can you confirm it's all there?

websupporter:
In `GET wp-json/` and `wp-json/wp/v2`

```
    "/wp/v2/settings": {
      "namespace": "wp/v2",
      "methods": [
        "GET",
        "POST",
        "PUT",
        "PATCH"
      ],
      "endpoints": [
        {
          "methods": [
            "GET"
          ],
          "args": []
        },
        {
          "methods": [
            "POST",
            "PUT",
            "PATCH"
          ],
          "args": {
            "title": {
              "required": false,
              "description": "Site title."
            },
```

joehoyle:
Ok sorry I missed we were also moving `create_initial_rest_routes` to 99, in that case, I don't see an issue for settings.

joehoyle (approved):

joehoyle:
I am happy for this, however I'd like to get the go-ahead from 1 other, so I'm not sure on the core precedent for 99 as a priority.

rmccue:
99 is used in 4 places in `default-filters.php`, so I'm happy with this.

Issue #2806, opened 2016-10-12
load_plugin_textdomain is missing.


mayukojpn:
We need `load_plugin_textdomain` to recognize text domain for translation editor like Poedit or GlotPress.

websupporter:
Hi @mayukojpn,
thanks so much for your effort and contributing. It is really appreciated. 

But I think, since this plugin is supposed to merge into core we do not need a textdomain.

mayukojpn:
Hi @websupporter 
Jorbin asked us to test internationalization on Slack polyglots channel, but we need text domain to do that.
https://wordpress.slack.com/archives/polyglots/p1476112154002294

websupporter:
Ah sorry. I didn't get that. If you need any help, hit me on Slack :)

websupporter:
But right now, we do not use the textdomain `rest-api`. We just use `__( 'Translate' )`... Is this a problem?

See for example here: https://github.com/WP-API/WP-API/blob/develop/lib/endpoints/class-wp-rest-attachments-controller.php#L45

mayukojpn:
Yes, We also need add textdomain after strings. Otherwise WordPress will use core translation file. I'm also working for that!

miya0001:
I feel that textdomain should not be added to this plugin, because we have to remove them after merging into WordPress core.
We need to try to generate `.po` with WordPress core, so we can translate them. 
Maybe it is not difficult.

@mayukojpn May I do it?

mayukojpn:
Thank you @miya0001 ! Can you talk with Jorbin about that?

miya0001:
Hi,
I developed a plugin that allowed you to translate temporarily rest-api plugin.
If you activate this plugin, you don't need to change textdomain or so.

Thanks!

https://github.com/miya0001/wp-api-i18n

kadamwhite:
Milestoning this as 2.1 because text domains can be added once the go/no-go on 4.7 is received. Thanks for your assistance @miya0001, this is hugely helpful!

PR #2811, opened 2016-10-12
Add textdomain.

mayukojpn:
#2806 We need load_plugin_textdomain to recognize text domain for translation editor like Poedit or GlotPress.

### DIFF
--- a/plugin.php
+++ b/plugin.php
@@ -7,6 +7,8 @@
  * Version: 2.0-beta15
  * Plugin URI: https://github.com/WP-API/WP-API
  * License: GPL2+
+ * Text Domain: rest-api
+ * Domain Path: /languages
  */
 
 /**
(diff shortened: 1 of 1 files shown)


miya0001:
Textdomain is not defined in this plugin like following.
https://github.com/WP-API/WP-API/blob/a53eec37cca0fa3f3bf932f04d0a02d45e1cb9d9/lib/endpoints/class-wp-rest-comments-controller.php#L58

So we have to generate `.po` and merge them into WordPress core's `.po`.

miya0001:
Related #2806

rmccue:
@miya0001 With your [comment](https://github.com/WP-API/WP-API/issues/2806#issuecomment-253177945) on #2806, is this pull request still needed?

miya0001:
@rmccue We don't need this. Thanks! 😄

mayukojpn:
Thank you for quick work @miya0001 san! 🙏

PR #2816, opened 2016-10-12
Word capitalization.

mayukojpn:
Since it is an abbreviation "Multipurpose Internet Mail Extensions" MIME should be all capitalized.

https://en.wikipedia.org/wiki/Media_type


rachelbaker (approved):

rachelbaker:
Thank you for the patch @mayukojpn

PR #2817, opened 2016-10-13
remove ununsed variable $post_type_obj

miya0001:
It removes following, because `$post_type_obj` isn't used.

```
$post_type_obj = get_post_type_object( $this->post_type );
```


rmccue (approved):

rmccue:
Thank you!

PR #2820, opened 2016-10-13
remove ununsed variable $use_cache

miya0001:
Remove `$use_cache`, because it isn't used in anywhere.


rachelbaker (approved):

rachelbaker:
thank you for the patch @miya0001

PR #2828, opened 2016-10-13
Pass WP_REST_Request to get_allowed_query_vars and containing filters…

### LINKED ISSUE #2821: Pass request to rest_query_vars filter
I want to be able to allow additional query parameters only when a specific request parameter exists to perform a meta query and ultimately return some posts.

Currently I've created a completely new endpoint that creates a new instance of `WP_REST_Posts_Controller` and ultimately returns the same output as the posts endpoint. 

I propose that we pass the `WP_REST_Request` instance through the constructor of `WP_REST_Posts_Controller::get_allowed_query_vars` and then pass it as a parameter in the `rest_query_vars` filter.

I'm quite sure other people might want to allow different query parameters for different requests but keep them restricted on others.

What do you think about this?

https://github.com/WP-API/WP-API/blob/develop/lib/endpoints/class-wp-rest-posts-controller.php#L665

https://github.com/WP-API/WP-API/blob/develop/lib/endpoints/class-wp-rest-posts-controller.php#L757

tobeycodes:
Resolves #2821

### DIFF
--- a/lib/endpoints/class-wp-rest-posts-controller.php
+++ b/lib/endpoints/class-wp-rest-posts-controller.php
@@ -662,7 +662,7 @@ class WP_REST_Posts_Controller extends WP_REST_Controller {
 	 */
 	protected function prepare_items_query( $prepared_args = array(), $request = null ) {
 
-		$valid_vars = array_flip( $this->get_allowed_query_vars() );
+		$valid_vars = array_flip( $this->get_allowed_query_vars( $request ) );
 		$query_args = array();
 		foreach ( $valid_vars as $var => $index ) {
 			if ( isset( $prepared_args[ $var ] ) ) {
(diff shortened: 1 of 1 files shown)


lib/endpoints/class-wp-rest-posts-controller.php
```diff
@@ -703,8 +704,9 @@ protected function get_allowed_query_vars() {
 		 * Allows adjusting of the default query vars that are made public.
 		 *
 		 * @param array  Array of allowed WP_Query query vars.
+		 * @param WP_REST_Request $request Request object.
 		 */
-		$valid_vars = apply_filters( 'query_vars', $wp->public_query_vars );
+		$valid_vars = apply_filters( 'query_vars', $wp->public_query_vars, $request );
```
rmccue: This is a [hook in core](https://developer.wordpress.org/reference/hooks/query_vars/) which we're duplicating here for technical reasons, so I'm not sure we _can_ pass the request here. Is `rest_query_vars` not powerful enough for this?

rmccue (changes requested):

schrapel:
It is. I just want to cover all grounds. I'll update

lib/endpoints/class-wp-rest-posts-controller.php
tobeycodes: It is fine. Wanted to cover all grounds. I've updated the PR

kadamwhite:
This is waiting on @rmccue's final review, but with the changes I'm +1 on this. Thanks for the patch @schrapel !

rmccue (approved):
Thanks a bunch! :)

PR #2832, opened 2016-10-14
Use Yoda Condition checks, you must

Soean:
Use Yoda Condition checks, you must
`$post->post_type === 'attachment'` in `delete_item()` in `class-wp-rest-posts-controller.php`


kadamwhite (approved):
Approved, this pull request is

PR #2834, opened 2016-10-14
Improve WP_REST_Revisions_Controller comments

Soean:
Improves the WP_REST_Revisions_Controller comments

### DIFF
--- a/lib/endpoints/class-wp-rest-revisions-controller.php
+++ b/lib/endpoints/class-wp-rest-revisions-controller.php
@@ -17,6 +17,8 @@ class WP_REST_Revisions_Controller extends WP_REST_Controller {
 
 	/**
 	 * Register routes for revisions based on post types supporting revisions
+	 * 
+	 * @access public
 	 */
 	public function register_routes() {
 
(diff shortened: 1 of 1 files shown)


lib/endpoints/class-wp-rest-revisions-controller.php
```diff
@@ -386,7 +386,8 @@ public function get_collection_params() {
 	/**
 	 * Check the post excerpt and prepare it for single post output.
 	 *
-	 * @param string       $excerpt
+	 * @param string      $excerpt The post excerpt.
+	 * @param int|WP_Post $post    Post revision object.
```
kadamwhite: I don't believe the `int` is relevant here -- `get_the_excerpt()` function takes an `int|WP_Post`, but then uses `get_post` to convert it to a WP_Post before passing to the `get_the_excerpt` filter. The $post parameter documented for the filter itself is `WP_Post $post Post object.`, and since we don't do any conversion in this method I believe we'd want to avoid taking in a post ID integer.

kadamwhite (changes requested):

lib/endpoints/class-wp-rest-revisions-controller.php
joehoyle: Agreed.

Soean:
@kadamwhite thanks for you feedback. You are right, the `int` is irrelevant. Here is my refresh.

kadamwhite (approved):

PR #2837, opened 2016-10-14
Do not run revisions excerpts through get_the_excerpt

kadamwhite:
Running revisions excerpts through `'get_the_excerpt'` leads to `get_the_content` being called in such a way that `$post` does not get set. This causes

```
<b>Notice</b>: Trying to get property of non-object in <b>/srv/www/wordpress-develop/src/wp-includes/post-template.php</b> on line <b>298</b>
```

to be printed out before the JSON, which prevents the JSON being properly interpreted by requesting clients, e.g. `$.get`.

@Shelob9 can you pull this down and test?


rachelbaker:
@kadamwhite I can confirm this PR fixes the issue for me as well

rachelbaker (approved):

PR #2839, opened 2016-10-15
Exclude unsupported params more broadly (was #2804)

kadamwhite:
Oops: I forgot that I'd opened #2804 as a pull request against #2770, _not against `develop`_; so while @joehoyle merged that PR, #2770 was already merged at the time so these commits landed in that branch but never made it in to `develop`

This PR is all the same approved and merged changes, just PR'd against the right branch.


BE-Webdesign (approved):
This PR was already approved but was set against the wrong branch.

PR #2840, opened 2016-10-15
Use wp_kses in place of wp_filter_post_kses to sanitize post title

### LINKED ISSUE #2788: Change the `post_title` database sanitization function to`wp_filter_kses()`
We should match WordPress Core's sanitization function when adding/updating the title of a Post.  WordPress Core uses [`wp_filter_kses()`](https://core.trac.wordpress.org/browser/trunk/src/wp-includes/kses.php#L1625) but we use [`wp_filter_post_kses()`](https://github.com/WP-API/WP-API/blob/develop/lib/endpoints/class-wp-rest-posts-controller.php#L804).

kadamwhite:
May fix #2788

This PR switches from using `wp_filter_post_kses()` for `post_title` sanitization to calling `wp_kses` directly: #2788 describes that the core behavior is to call `wp_filter_kses()`, but @westonruter notes in that thread that the slash handling in `wp_filter_kses` is lossy so using the underlying implementation of `wp_filter_kses` without the de-slashing and re-slashing should provide adequate sanitization without compromising the integrity of the content.

Review requested especially from @rachelbaker or @westonruter

### DIFF
--- a/lib/endpoints/class-wp-rest-posts-controller.php
+++ b/lib/endpoints/class-wp-rest-posts-controller.php
@@ -825,9 +825,9 @@ class WP_REST_Posts_Controller extends WP_REST_Controller {
 		if ( ! empty( $schema['properties']['title'] ) && isset( $request['title'] ) ) {
 			if ( is_string( $request['title'] ) ) {
 				// wp_filter_kses slash handling is lossy: use the underlying methods directly
-				$prepared_post->post_title = wp_kses( $request['title'], current_filter() );
+				$prepared_post->post_title = wp_kses( $request['title'], 'title_save_pre' );
 			} elseif ( ! empty( $request['title']['raw'] ) ) {
-				$prepared_post->post_title = wp_kses( $request['title']['raw'], current_filter() );
+				$prepared_post->post_title = wp_kses( $request['title']['raw'], 'title_save_pre' );
 			}
 		}
 
--- a/lib/endpoints/class-wp-rest-posts-controller.php
+++ b/lib/endpoints/class-wp-rest-posts-controller.php
@@ -824,9 +824,10 @@ class WP_REST_Posts_Controller extends WP_REST_Controller {
 		// Post title.
 		if ( ! empty( $schema['properties']['title'] ) && isset( $request['title'] ) ) {
 			if ( is_string( $request['title'] ) ) {
-				$prepared_post->post_title = wp_filter_post_kses( $request['title'] );
+				// wp_filter_kses slash handling is lossy: use the underlying methods directly
+				$prepared_post->post_title = wp_kses( $request['title'], current_filter() );
 			} elseif ( ! empty( $request['title']['raw'] ) ) {
-				$prepared_post->post_title = wp_filter_post_kses( $request['title']['raw'] );
+				$prepared_post->post_title = wp_kses( $request['title']['raw'], current_filter() );
 			}
 		}
 
--- a/tests/test-rest-posts-controller.php
+++ b/tests/test-rest-posts-controller.php
@@ -1275,6 +1275,45 @@ class WP_Test_REST_Posts_Controller extends WP_Test_REST_Post_Type_Controller_Te
 		$this->assertEquals( "Rob O'Rourke's Diary", $new_data['title']['raw'] );
 	}
 
+	public function test_create_post_with_acceptable_html_in_title() {
+		wp_set_current_user( $this->editor_id );
+
+		$request = new WP_REST_Request( 'POST', '/wp/v2/posts' );
+		$params = $this->set_post_data( array(
+			'title' => 'An <em>Emphatic</em> Headline',
+		) );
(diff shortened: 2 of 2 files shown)


lib/endpoints/class-wp-rest-posts-controller.php
```diff
@@ -824,9 +824,10 @@ protected function prepare_item_for_database( $request ) {
 		// Post title.
 		if ( ! empty( $schema['properties']['title'] ) && isset( $request['title'] ) ) {
 			if ( is_string( $request['title'] ) ) {
-				$prepared_post->post_title = wp_filter_post_kses( $request['title'] );
+				// wp_filter_kses slash handling is lossy: use the underlying methods directly
+				$prepared_post->post_title = wp_kses( $request['title'], current_filter() );
```
westonruter: I think that `current_filter()` should be replaced with `'post'`. I'm not sure if there are filters being applied at this point, but even if they are, for sure they're not going to be the same as what are being applied during a call to `sanitize_post()`.

lib/endpoints/class-wp-rest-posts-controller.php
```diff
@@ -824,9 +824,10 @@ protected function prepare_item_for_database( $request ) {
...
 			} elseif ( ! empty( $request['title']['raw'] ) ) {
-				$prepared_post->post_title = wp_filter_post_kses( $request['title']['raw'] );
+				$prepared_post->post_title = wp_kses( $request['title']['raw'], current_filter() );
```
westonruter: ″

westonruter (changes requested):

lib/endpoints/class-wp-rest-posts-controller.php
rachelbaker: I thought `current_filter()` here should be `title_save_pre` because post titles get a stricter set of HTML tags allowed than `wp_kses_post()`

westonruter: You're probably right. So the string literal `'title_save_pre'` should be used instead of `current_filter()`.

kadamwhite:
@rachelbaker @westonruter thank you for clarifying, that makes sense. I've updated the patch

rachelbaker:
@kadamwhite we use `wp_filter_post_kses()` in several other places within the endpoints.
So if we are not trusting the slash handling of `wp_filter_post_kses()` then we  should do that consistently, and switch to `wp_kses_post()` or `wp_kses()` with the  {filter_name}.

Recording this here, but can be done in another PR.

rachelbaker:
@kadamwhite Only thing blocking this would be a unit test.

kadamwhite:
Unit test would be e.g. create a post w/ a title including a disallowed HTML tag such as `<div>`, and ensure it is stripped

tests/test-rest-posts-controller.php
```diff
@@ -1275,6 +1275,45 @@ public function test_create_post_with_quotes_in_title() {
...
+		wp_set_current_user( $this->editor_id );
+
+		$request = new WP_REST_Request( 'POST', '/wp/v2/posts' );
+		$params = $this->set_post_data( array(
+			'title' => '\o/',
+		) );
+		$request->set_body_params( $params );
+		$response = $this->server->dispatch( $request );
+		$new_data = $response->get_data();
+		$this->assertEquals( '\o/', $new_data['title']['raw'] );
```
kadamwhite, line 1314: @westonruter @rachelbaker This test fails, but I would have expected it to pass based on our discussion previously. Any ideas?

rachelbaker: @kadamwhite My guess would be that`wp_kses_no_null()` (called in `wp_kses()`)  is stripping out the preceding slash.

kadamwhite: If we can't escape (har har) the unslashing, is the use of `wp_kses` overkill? if you can suggest a direction I will implement.

rachelbaker: @kadamwhite crazy idea, what happens if you run this test with `wp_filter_kses` instead?

kadamwhite: Same thing

rachelbaker: @kadamwhite yeah, I tested it too and was not able to get a passing test.  😞 
I also WAS able to confirm that within the WP editor I can save a post with a title of `Do what I mean \ Not what I say`, and that is also exactly how the data is stored in the db.  So we are doing something WRONG somewhere here.

rmccue: This is potentially an issue because of the unescaped slash? Try `\\o/` instead?

kadamwhite: @rmccue Same results with `'\\o/'`, with both wp_filter_kses and wp_kses

kadamwhite:
Closing, as I believe this has been superseded by 4.7 trunk

PR #2841, opened 2016-10-16
Use strict comparisons for in_array.

peterwilsoncc:
Adds the strict argument to `in_array()` comparisons where safe to do so.

As WordPress can be somewhat inconsistent with the way it returns IDs, I've left them to be juggled in the unit tests.


kadamwhite:
@rmccue @joehoyle @rachelbaker Defer to you on this one; WP coding standards don't seem to provide an opinion and only about 10% of core's `in_array()` calls specify strict comparison.

joehoyle (approved):

joehoyle:
I'm happy for this to go in, this "feature" of in array has always seemed like a weird default to me.

kadamwhite:
Sounds good to me then!

PR #2842, opened 2016-10-16
Use `elseif` consistently.

peterwilsoncc:
Pure style: replaces a few `else if`s with `elseif` in the PHP to get some consistency.


kadamwhite (approved):
This conforms to [WP coding standards](https://make.wordpress.org/core/handbook/best-practices/coding-standards/php/#use-elseif-not-else-if) and tests pass, mergin'.

PR #2850, opened 2016-10-17
Only allow custom meta of types string, number and boolean

joehoyle:
Currently we handle registering meta even if your register the type as
anything other than `boolean` `number` and `string`, however we don't
currently have great ways to handle non-scaler meta. For now, let's
whitelist these types and throw a `doing_it_wrong` if meta is registered
with an invalid type.

### DIFF
--- a/lib/fields/class-wp-rest-meta-fields.php
+++ b/lib/fields/class-wp-rest-meta-fields.php
@@ -286,6 +286,12 @@ abstract class WP_REST_Meta_Fields {
 					continue;
 				}
 
+				// Whitelist the supported types for types, as we don't want invalid types
+				// to be updated with arbitrary values that we can't do decent sanitizing for.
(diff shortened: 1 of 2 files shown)


kadamwhite:
How do we handle namespacing of objects in meta? e.g., if I'm an seo plugin and I have a "score" meta value, but there's also a keyword-bingo plugin that also has a "score," and I only have a numeric meta field to put them in -- not an object -- how do I ensure there won't be a conflict?

lib/fields/class-wp-rest-meta-fields.php
```diff
@@ -294,6 +300,10 @@ protected function get_registered_fields() {
 						'type' => $args['type'],
 					);
 				}
+			} else {
+				if ( ! in_array( $rest_args['schema']['type'], array( 'number', 'string', 'boolean' ), true ) ) {
```
kadamwhite, line 304: elseif?

Issue #2838, opened 2016-10-14
API should error when attempting to update a registered meta field holding an invalid value


kadamwhite:
Based on discussion following #2765:

Registering a meta field defines the data type that field should contain, but there are ways outside of the API's purview that a post could come to hold an invalid value in that meta field, such as a serialized object, which does not match the registered type. If the API tries to update the value of a meta object and the API discovers that an invalid type of data is being stored in that field, it should return an error rather than replacing the value of that field.

PR #2851, opened 2016-10-17
Send an error if the client tries to delete an option with an non-scalar

joehoyle:
This is to cover the case when there is an invalid option value in the database that does not match the one in register_setting(). 

Currently we are only doing this check for setting values to `null` to work around the Client-sends-the-same-response issue, but we _do_ all setting the value to some other value.

### DIFF
--- a/lib/endpoints/class-wp-rest-settings-controller.php
+++ b/lib/endpoints/class-wp-rest-settings-controller.php
@@ -123,7 +123,7 @@ class WP_REST_Settings_Controller extends WP_REST_Controller {
 				if ( ! is_scalar( get_option( $args['option_name'], false ) ) ) {
 					return new WP_Error(
 						'rest_invalid_stored_value',
-						sprintf( __( 'The %s property has an invalid stored value, and was not able to be updated to null.' ), $name ),
+						sprintf( __( 'The %s property has an invalid stored value, and cannot be updated to null.' ), $name ),
 						array( 'status' => 500 )
 					);
 				}
--- a/lib/endpoints/class-wp-rest-settings-controller.php
+++ b/lib/endpoints/class-wp-rest-settings-controller.php
@@ -70,6 +70,11 @@ class WP_REST_Settings_Controller extends WP_REST_Controller {
 	 * @return mixed
 	 */
 	protected function prepare_value( $value, $schema ) {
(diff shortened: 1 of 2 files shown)


lib/endpoints/class-wp-rest-settings-controller.php
```diff
@@ -98,7 +103,19 @@ public function update_item( $request ) {
 			}
 			// A null value means reset the option, which is essentially deleting it
 			// from the database and then relying on the default value.
+			//
+			// However, bacause a saved "invalid" value in the database will
```
rachelbaker: @joehoyle Typo here.  bacause -> because

kadamwhite:
@joehoyle Looks like this addresses #2838, is that the correct issue for this PR?

lib/endpoints/class-wp-rest-settings-controller.php
```diff
@@ -70,6 +70,11 @@ public function get_item( $request ) {
 	 * @return mixed
 	 */
 	protected function prepare_value( $value, $schema ) {
+		// if the value is not a scaler, it's not possible to cast it to anything.
```
rmccue: "scaler" -> "scalar"

lib/endpoints/class-wp-rest-settings-controller.php
```diff
@@ -98,7 +103,19 @@ public function update_item( $request ) {
...
+			//
+			// However, bacause a saved "invalid" value in the database will
+			// cause a null in the response to the client, we only allow deleting of
+			// options that have a scalar value. Otherwise, sending the response back to
+			// the server will cause a delete of all invalid options.
 			if ( is_null( $request[ $name ] ) ) {
+				if ( ! is_scalar( get_option( $args['option_name'], false ) ) ) {
+					return new WP_Error(
+						'rest_invalid_stored_value',
+						sprintf( __( 'The %s property has an invalid stored value, and was not able to be updated to null.' ), $name ),
```
rmccue: "was not able to be updated to null" -> "cannot be updated"

rmccue (changes requested):

rachelbaker:
@rmccue #reviewmerge I addressed your feedback items, please review the changes when you can.

rachelbaker:
Committed https://core.trac.wordpress.org/changeset/38982

PR #2852, opened 2016-10-17
Remove "validate_callback" on arrays where it's not used.

joehoyle:
Currently we are specify teh `validate_callback` for the number-list
style params that are passed to `wp_parse_id_list()`, however, the
validate callback never does anything here, as our validate callback
doesn't handle array. When we _do_ support arrays, we won't be able to
use it here anyway, as these are "special" params that are not really
arrays, but in stread comma seperated strings that are converted to
arrays.


kadamwhite (approved):
Seems reasonable

Issue #2783, opened 2016-10-10
Modify the schema to apply callbacks or required parameters to nested objects


rachelbaker:
Allow schema properties like `sanitization_callback`, `type`, `required`, etc. to be set and processed on nested objects.

Currently `rest_sanitize_request_arg()` and `rest_validate_request_arg()` only handle properties set on the top level of nested objects, which means if we try to set properties on `content.raw` they are ignored.

rachelbaker:
Split off from #2758

rachelbaker:
@rmccue You mentioned that you might take a swing at this for Meta, still interested?

PR #2853, opened 2016-10-17
Disable the schema sanitization on content / excerpt / title polymorphics

joehoyle:
These values are polymorphic, so we can't do any smart sanitization
based off the schema (this sanitization is already implemented in the
`prepare_item_for_database()` function). We'll be supporting deep object
and array validation from the schema at some point, where we won't want
the schema validation for these polymorphic fields.

It only happens that we _dont_ validate / sanitize fields of type object
right now that these slip through the net.

### DIFF
--- a/lib/endpoints/class-wp-rest-comments-controller.php
+++ b/lib/endpoints/class-wp-rest-comments-controller.php
@@ -953,6 +953,9 @@ class WP_REST_Comments_Controller extends WP_REST_Controller {
 					'description'     => __( 'The content for the object.' ),
 					'type'            => 'object',
 					'context'         => array( 'view', 'edit', 'embed' ),
+					'arg_options' => array(
+						'sanitize_callback' => null,
+					),
 					'properties'      => array(
 						'raw'         => array(
 							'description'     => __( 'Content for the object, as it exists in the database.' ),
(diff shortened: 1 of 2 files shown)


lib/endpoints/class-wp-rest-comments-controller.php
```diff
@@ -953,6 +953,9 @@ public function get_item_schema() {
 					'description'     => __( 'The content for the object.' ),
 					'type'            => 'object',
 					'context'         => array( 'view', 'edit', 'embed' ),
+					'arg_options' => array(
+						'sanitize_callback' => null,
+					),
```
kadamwhite, line 958: I'd like to see a `// polymorphic value: sanitization implemented in controller` or some such on these, so that the rationale for this otherwise inscrutable change doesn't get lost

kadamwhite (changes requested):

rachelbaker:
Related #2783

lib/endpoints/class-wp-rest-comments-controller.php
```diff
@@ -953,6 +953,9 @@ public function get_item_schema() {
 					'description'     => __( 'The content for the object.' ),
 					'type'            => 'object',
 					'context'         => array( 'view', 'edit', 'embed' ),
+					'arg_options' => array(
```
rmccue, line 956: Needs to be aligned.

lib/endpoints/class-wp-rest-comments-controller.php
rmccue: Agreed that this should have a comment, maybe something simpler like "Note: sanitization implemented in callback"?

rmccue (changes requested):

joehoyle:
New patch in https://core.trac.wordpress.org/ticket/38529 with added comments.

PR #2854, opened 2016-10-17
Remove unused parameter

Soean:
Remove unused parameter `$parameter` in `validate_user_can_query_private_statuses()`.

### DIFF
--- a/lib/endpoints/class-wp-rest-posts-controller.php
+++ b/lib/endpoints/class-wp-rest-posts-controller.php
@@ -1929,10 +1929,9 @@ class WP_REST_Posts_Controller extends WP_REST_Controller {
 	 *
 	 * @param  mixed $value
 	 * @param  WP_REST_Request $request
-	 * @param  string $parameter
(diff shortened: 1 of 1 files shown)


lib/endpoints/class-wp-rest-posts-controller.php
```diff
@@ -1929,10 +1929,9 @@ public function get_collection_params() {
 	 *
 	 * @param  mixed $value
 	 * @param  WP_REST_Request $request
-	 * @param  string $parameter
 	 * @return WP_Error|boolean
 	 */
-	public function validate_user_can_query_private_statuses( $value, $request, $parameter ) {
+	public function validate_user_can_query_private_statuses( $value, $request ) {
```
kadamwhite, line 1934: This function is consumed by the attachments controller as well, and matches the signature of other validation methods. As the parameter will be passed in to this method even though it is not used, I believe it is appropriate to leave the parameter in place. (@joehoyle or @rmccue, please disagree if I am off-base here).

PR #2855, opened 2016-10-17
Improves WP_REST_Posts_Controller documentation.

Soean:


rachelbaker:
@Soean thanks you for the patch.  WP Core's documentation standard only uses the `@access` property for functions.  See https://make.wordpress.org/core/handbook/best-practices/inline-documentation-standards/php/#1-functions-class-methods

Soean:
@rachelbaker thanks for the feedback. Core uses `@access` also for properties. See https://make.wordpress.org/core/handbook/best-practices/inline-documentation-standards/php/#2-1-class-members
example: https://core.trac.wordpress.org/browser/tags/4.6/src/wp-includes/rest-api/class-wp-rest-response.php

rachelbaker:
@Soean That is correct.  So keep the `@access` property on the properties, but remove from the methods and this will be 👍

Soean:
@rachelbaker Here is my refresh

kadamwhite:
@Soean +1 on the change; if you can fix the merge conflict (merge latest `develop` into this branch) this is ready to go!

lib/endpoints/class-wp-rest-posts-controller.php
```diff
@@ -677,8 +688,8 @@ public function delete_item( $request ) {
 	 * Determine the allowed query_vars for a get_items() response and
 	 * prepare for WP_Query.
 	 *
-	 * @param array           $prepared_args
-	 * @param WP_REST_Request $request
+	 * @param array           $prepared_args Prepared WP_Query Arguments.
```
rachelbaker, line 691: `Arguments` should be lower case here.

rachelbaker (approved):

Issue #2785, opened 2016-10-10
On Comment POST/PUT return WP_Error for strings that exceed the DB column size


rachelbaker:
Mimic the logic in `wp_handle_comment_submission()` to use `wp_get_comment_fields_max_lengths()` and return the appropriate error for 
'comment_author', 'comment_author_email', 'comment_author_url', or 'comment_content' strings that exceed the maximum length of their column in the database.

salcode:
Congratulations on merging the REST API into WordPress core!

https://core.trac.wordpress.org/changeset/38832

I assume this means development will stop here on GitHub and I should now open a WordPress trac ticket for this issue and submit the patch there.

Unless I hear otherwise, I'll make this my next step.

Thank you for all your work on this great project.

rachelbaker:
Related Trac ticket: https://core.trac.wordpress.org/ticket/38477

PR #2858, opened 2016-10-18
Add check for comments that exceed DB col size

salcode:
Add checks for these fields exceeding the DB col size and tests for each
field:
- comment_author
- comment_author_email
- comment_author_url
- comment_content

See #2785

### DIFF
--- a/lib/endpoints/class-wp-rest-comments-controller.php
+++ b/lib/endpoints/class-wp-rest-comments-controller.php
@@ -409,24 +409,6 @@ class WP_REST_Comments_Controller extends WP_REST_Controller {
 			return $prepared_comment['comment_approved'];
 		}
 
-		$max_lengths = wp_get_comment_fields_max_lengths();
-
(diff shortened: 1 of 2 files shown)


salcode:
`CONTRIBUTING.md` asks that I leave a #reviewmerge comment, which I'm not familiar with.  Hopefully, this comment with the hash tag is sufficient.  If not, I'd appreciate any guidance (or a relevant link) on how to add this properly.

Thanks.

rachelbaker:
@salcode What do you think about moving this into the `prepare_item_for_database()` method so we are covered on create & update actions?

rachelbaker:
This PR would also benefit from a few unit tests.

schlessera:
@rachelbaker, the original PR already includes the corresponding unit tests.

salcode:
@rachelbaker thanks for your suggestion.  I've moved the DB column size check to `prepare_item_for_database()` so as to cover both create and update.

Currently, there are unit tests for **create**.  Am I correct you're suggesting we add tests to cover **update**?

PR #2861, opened 2016-10-19
Comments: Check discussion option setting for name & email during create validation

rachelbaker:
Adds a conditional check to the `create_item()` method in the Comments Controller for the `require_name_email` option.  When this option value is set to `1` comments should not be created without values for the name or email address of the comment author. 

Replaces #2049

### DIFF
--- a/lib/endpoints/class-wp-rest-comments-controller.php
+++ b/lib/endpoints/class-wp-rest-comments-controller.php
@@ -189,8 +189,6 @@ class WP_REST_Comments_Controller extends WP_REST_Controller {
 		if ( get_option( 'require_name_email' ) && ! isset( $user ) ) {
 			if ( empty( $prepared_comment['comment_author_email'] ) || empty( $prepared_comment['comment_author'] ) ) {
 				return new WP_Error( 'rest_require_valid_comment', __( 'Required fields (name, email) missing.' ), array( 'status' => 400 ) );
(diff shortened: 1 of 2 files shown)


rachelbaker:
@WP-API/amigos #reviewmerge

lib/endpoints/class-wp-rest-comments-controller.php
```diff
@@ -384,6 +384,17 @@ public function create_item( $request ) {
 			$prepared_comment['comment_author_url'] = $user->user_url;
 		}
 
+		// Check author name and email if required.
+		if ( get_option( 'require_name_email' ) ) {
+			if ( ! isset( $prepared_comment['comment_author'] ) && ! isset( $prepared_comment['comment_author_email'] ) ) {
+				return new WP_Error( 'rest_comment_author_data_required', __( 'Creating a comment requires valid author name and email values.' ), array( 'status' => 400 ) );
+			} elseif ( ! isset( $prepared_comment['comment_author'] ) ) {
```
rmccue: These can be changed from `elseif`s to `if`s

kadamwhite (approved):
👍 Looks good to me, just waiting for the tests to pass

lib/endpoints/class-wp-rest-comments-controller.php
rachelbaker: Fixed!

Issue #2860, opened 2016-10-18
Settings comparison with WordPress.com


joehoyle:
## Comparison of Settings in WordPress.com

The response format for the WordPress.com API is somewhat different and follows the pattern:

``` json
{
    "ID": 123,
    "name": "My Site"
    ...,
    "settings": {
        "admin_url" : "https://....",
        "require_name_email": false,
        ...
    }
}
```

The WP REST API Response has:

``` json
{
    "url" : "https://...",
    "default_category": 1,
    ...
}
```

It's not clear _why_ the WordPress.com API takes this structure with a `settings` sub key, however it would be good to find out why!

Broadly, I'd say the settings names and values in the WordPress.com API have had little change, and therefor can be somewhat cryptic with names such as `wga`, `holidaysnow`, `comment_whitelist` - as there's no published documentation for the Settings API, it seems rather difficult for a developer to use (however I think this is being worked on).

Updating settings is also different on the WordPress.com api, instead of it returning the updated resource, it returns an object with which settings for were updated:

``` json
{ "updated": [] }
```

To update settings, you don't send the same object shape received in the GET request, you instead send a key => value object. This matches the style in the WP REST API, however specific option names may differ.

``` json
{ "default_ping_status" : "open" }
```
### What we can adopt

I think the best thing we can learn from the WordPress.com Settings API is the fields they have chosen in include in the response. There's a mixture of core and jetpack settings, so we should of course ignore the jetpack ones, but make all the core ones available via WP REST API.

This will provide like-for-like _functionality_ (though not an exact map) for WordPress core features
to the WordPress.com API.

I recommend we _don't_ follow the same response format with the `settings` sub key, as there doesn't appear to be a good enough reason for this abnormal structure. Likewise, I recommend we _don't_ follow the same response format when updating settings with the `updated` key, as we should stay consistent with all other REST endpoints which is to return the same formatted object as a `GET` request.

tobeycodes:
The settings sub key doesn't make much sense to me either. ID and Name are still considered settings to me, usually they won't change but they can.

nylen:
cc @rralian who wrote the initial version of the WP.com settings endpoint for questions on response format.

nylen:
It makes sense for us to add our Jetpack-specific settings via `register_setting`.

[...]

PR #2864, opened 2016-10-19
Settings: Add filters to override getting and updating the values of settings

nylen:
This allows for a couple of interesting use cases:
- Making certain settings read-only (we will need to do this on WP.com, if we don't end up removing the settings in question altogether)
- Exposing properties via the settings endpoint that are stored as something other than simple site options (we will need to do this on WP.com as well)

### DIFF
--- a/tests/test-rest-settings-controller.php
+++ b/tests/test-rest-settings-controller.php
@@ -166,7 +166,7 @@ class WP_Test_REST_Settings_Controller extends WP_Test_REST_Controller_Testcase 
 	}
 
 	public function update_setting_custom_callback( $result, $name, $value, $args ) {
-		if ( $name === 'title' && $value === 'The new title!' ) {
+		if ( 'title' === $name && 'The new title!' === $value ) {
 			// Do not allow changing the title in this case
 			return true;
 		}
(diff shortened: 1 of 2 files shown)


georgestephanis:
related: #2860

tests/test-rest-settings-controller.php
```diff
@@ -166,7 +166,7 @@ public function test_update_item() {
 	}
 
 	public function update_setting_custom_callback( $result, $name, $value, $args ) {
-		if ( $name === 'title' && $value === 'The new title!' ) {
```
rachelbaker: (https://media.giphy.com/media/3o7abBAEIlO4kWucQ8/giphy.gif)

rachelbaker (commented):
@nylen I am thinking that `rest_pre_get_setting` and `rest_pre_update_setting` may better describe the utlity of the filters.  Thoughts?

rmccue:
lgtm, agreed with @rachelbaker. (Restarted the tests, as they were failing due to a momentary issue on WordPress.org)

nylen:
Fine with me.  Updated (and fixed a little bug in the tests).

rachelbaker (approved):

PR #2867, opened 2016-10-19
Merge multiple `unset()` class into a single one with multiple arguments.

schlessera:
There are several places in the code that make two consecutive `unset()` calls.

`unset()` accepts multiple arguments, and can unset all of the passed arguments in one go.

The changes in this PR merge these calls into one single call with multiple arguments, eliminating the overhead of an additional function call.

### DIFF
--- a/lib/endpoints/class-wp-rest-posts-controller.php
+++ b/lib/endpoints/class-wp-rest-posts-controller.php
@@ -259,8 +259,10 @@ class WP_REST_Posts_Controller extends WP_REST_Controller {
 		$request_params = $request->get_query_params();
 		if ( ! empty( $request_params['filter'] ) ) {
 			// Normalize the pagination params.
-			unset( $request_params['filter']['posts_per_page'] );
-			unset( $request_params['filter']['paged'] );
+			unset(
+				$request_params['filter']['posts_per_page'],
+				$request_params['filter']['paged']
+			);
 		}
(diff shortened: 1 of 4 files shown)


lib/endpoints/class-wp-rest-posts-controller.php
```diff
@@ -259,8 +259,10 @@ public function get_items( $request ) {
 		$request_params = $request->get_query_params();
 		if ( ! empty( $request_params['filter'] ) ) {
 			// Normalize the pagination params.
-			unset( $request_params['filter']['posts_per_page'] );
-			unset( $request_params['filter']['paged'] );
+			unset(
+				$request_params['filter']['posts_per_page'],
+				$request_params['filter']['paged']
+			);
```
rmccue: I'd rather stick to the two function calls here than have to break on to multiple lines anyway. Alternatively, we could keep this on a single line, although it will be a long one.

rmccue (changes requested):

lib/endpoints/class-wp-rest-posts-controller.php
schlessera: I have changed the PR to keep it in one line (with only one `unset()`).

The line is long, yes, but it is actually shorter than the one directly below... :)

rmccue (approved):

PR #2869, opened 2016-10-19
Avoid slow `in_array()` for single-item arrays.

schlessera:
`in_array()` is a rather slow operation, and should not be used to compare against arrays that are known to only contain one element.

This PR replaces such comparisons with a simpler and faster equality check.


rmccue (approved):

schlessera:
Added another replacement for `in_array()`, where the check was a simple key lookup.

BE-Webdesign:
Are there any other places in the code where `in_array()` is used ineffectively before I merge this?

schlessera:
I saw one other occurrence here: https://github.com/WP-API/WP-API/blob/develop/lib/endpoints/class-wp-rest-posts-controller.php#L1032

However, tacking square brackets for an array access at the end of that statement will probably result in incomprehensible code...

PR #2872, opened 2016-10-19
Replace PHP4 covnersion functions with typecasts.

schlessera:
The PHP4-style conversion functions like `strval()` and `floatval()` incur an unnecessary function call overhead.

This PR replaces them with the corresponding typecasts (`(string)` and `(float)`, respectively).


BE-Webdesign (approved):
Looks good. There is an error in the code, that I now see thanks to this PR!

PR #2873, opened 2016-10-19
Fix case for `WP_Http` class.

schlessera:
The `WP_Http` class was erroneously used as `WP_HTTP`.

This PR switches to the correct case.


websupporter (approved):
Looks good to me. Additionally: couldn't find other WP_Http instances in the Rest API

PR #2874, opened 2016-10-19
Fix arguments mismatch for `wp_allow_comment()`.

schlessera:
According to the [Codex](https://codex.wordpress.org/Function_Reference/wp_allow_comment), `wp_allow_comment()` only accepts one argument.

This PR fixes an instance where this function is called with a superfluous second boolean argument.


websupporter:
Ahh! No :) See trunk. This took a while to get in 

https://core.trac.wordpress.org/ticket/36901

Codex is out of date here.

schlessera:
Ah, I thought there would probably be a recent change I was missing, but my search did not show any results. I might need to improve my searching then! ;)

PR #2875, opened 2016-10-19
Fix an issue where a static method is called dynamically.

schlessera:
Static methods should always be called statically, to avoid confusion and potential issues.

This PR fixes such an instance, by using the `self` keyword (as late-static binding with the `static` keyword is only available from PHP 5.3 onwards).


rachelbaker (approved):

rachelbaker:
Thanks again @schlessera

PR #2876, opened 2016-10-19
Remove misplaced semicolon.

schlessera:
Remove a misplaced semicolon that was trailing after an if-clause brace.


rachelbaker:
@schlessera Haha, oh the problems of changing from JS to PHP :)

rachelbaker (approved):

PR #2878, opened 2016-10-19
Add missing PHPDoc tags.

schlessera:
This PR completes some of the PHPDoc tags so that typehinting works more reliable within IDEs.


BE-Webdesign (approved):
Looks good.

PR #2881, opened 2016-10-19
Add latest js client

adamsilverstein:
Changes this release:
- No longer auto-parse dates/timestamps - previously these fields were transformed from timestamps to JavaScript Dates when fetching and back in timestamp strings when saving. Calculations were not always accurate leading to unexpected/incorrect times.
- Adds Date helper methods to models with date fields `getDate` and `setDate` that provide the same date transformations. calling these by default in `parse` and `toJSON` would result in the same behavior as before.


adamsilverstein:
Also props @jasontheadams for his contribution: https://github.com/WP-API/client-js/pull/139

kadamwhite (approved):

PR #2885, opened 2016-10-19
`in_array()` should only ever be used in strict mode.

schlessera:
Change 1 occurrence in plugin and 30 occurrences in the tests of `in_array()` that are missing the third boolean argument to enable strict mode.

### DIFF
--- a/plugin.php
+++ b/plugin.php
@@ -455,7 +455,7 @@ if ( ! function_exists( 'rest_validate_request_arg' ) ) {
 		$args = $attributes['args'][ $param ];
 
 		if ( ! empty( $args['enum'] ) ) {
-			if ( ! in_array( $value, $args['enum'] ) ) {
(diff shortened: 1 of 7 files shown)


kadamwhite:
#2841 contained a previous iteration of this change, my understanding from that PR was that there were circumstances where applying strict mode might change the existing behavior; I'm +1 on this change but we should be sure we aren't relying on the lax comparison anywhere.

schlessera:
Yes, I understand that this can be tricky.

However, if something should break because of these strict comparisons, we should strive to cast/validate/filter the data _before_ doing the comparison, instead of leaving the comparison more lax. When not using strict mode, `in_array()` can have very surprising and unexpected results.

For what it's worth, the unit tests all passed.

kadamwhite (approved):
Sold—and as you say, the tests are green. I'm trusting to our coverage :)

PR #2996, opened 2017-01-05
Add github templates to point would-be submitters to trac/forums

kadamwhite:
This might help direct the flow of incoming issues and PRs over to the main development and support channels


Kenshino (commented):
I know everyone should know but I think the reason why Development and support is no longer taking place here should be mentioned - because of the full merge of 4.7

So all support and bugs should now be directed to official avenues etc.

adamsilverstein (approved):

